PR #40, opened 2025-10-16
feat: SSM client

aws-jasakshi:
- Add SSM client 
- Configure extension API server with certificates and service
- Add Helm chart support for extension API toggle
- Update kustomization for extension API patches

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -11,10 +11,7 @@ jobs:
     name: Run on Ubuntu
     runs-on: ubuntu-latest
     permissions:
-      contents: read       # To checkout code
-      pull-requests: write # To add ci:e2e-approved label
-    # Require approval UNLESS: on main, manual trigger, or PR has ci:e2e-approved label
-    environment: ${{ !(github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:e2e-approved'))) && 'e2e-tests' || '' }}
+      contents: read # To checkout code
 
     steps:
       - name: Clone the code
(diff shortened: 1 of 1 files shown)


internal/ssm/client.go
```diff
@@ -0,0 +1,146 @@
+// Package ssm provides AWS SSM client functionality for workspace access
+package ssm
```
aws-prayags: suggestion - `internal/aws` folder and `ssm_client.go`. Other aws clients can then be added in this folder.

internal/aws/ssm_client.go
```diff
@@ -0,0 +1,134 @@
...
+// FindInstanceByPodUID finds SSM managed instance by pod UID tag
+func (c *SSMClient) FindInstanceByPodUID(ctx context.Context, podUID string) (string, error) {
+	input := &ssm.DescribeInstanceInformationInput{
+		Filters: []types.InstanceInformationStringFilter{
+			{
+				Key:    aws.String("tag:workspace-pod-uid"),
+				Values: []string{podUID},
+			},
+		},
+	}
```
JGuinegagne: sanity-check: any pagination consideration? I think the answer is 'no' but just double checking.

internal/aws/ssm_client.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if result.StreamUrl != nil {
+		sessionInfo.StreamURL = *result.StreamUrl
+	}
+	if result.TokenValue != nil {
+		sessionInfo.TokenValue = *result.TokenValue
+	}
+
+	sessionInfo.WebSocketURL = fmt.Sprintf("wss://ssmmessages.%s.amazonaws.com/v1/data-channel/%s", c.region, *result.SessionId)
```
JGuinegagne: non-blocking: please track support for non-aws partitions in a GH issue if you merge this.

dist/chart/values.yaml
```diff
@@ -90,4 +90,9 @@ application:
   imagesRegistry: "docker.io/library"
   # Watch resource configuratio for traefik
   traefik:
-    enable: false
\ No newline at end of file
+    enable: false
+
+# [EXTENSION_API]: Extension API server configuration
```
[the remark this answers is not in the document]
aws-jasakshi: Previously, I was using this functionality only in the main file, which is not included in this commit. The API server resources were being created by default, but could only be used when the flag was enabled. I have updated this PR to also create the resources conditionally

config/manager/manager.yaml
```diff
@@ -67,6 +67,9 @@ spec:
           - --application-images-registry=docker.io/library
         image: controller:latest
         name: manager
+        env:
+        - name: SSM_DOCUMENT_NAME
+          value: "SageMaker-SpaceSSHSessionDocument"
```
JGuinegagne: `SSM_` isn't generic. Maybe `AWS_SSM_DOCUMENT_NAME` ?

Medium term: would it be possible to set the default value to `WorkspaceSessionDocument`, as in not SageMaker specific?
Then you can override this value in the `makefile` / guided charts for SageMaker-specific use cases.

hack/helm-patches/values.yaml.patch
```diff
@@ -14,3 +14,10 @@ accessResources:
     enable: false
   # Additional Group-Version-Kind to watch
   additionalGvk: []
+
+# [EXTENSION_API]: Extension API server configuration  
+extensionApi:
+  # Enable extension API server for WorkspaceConnection
```
aws-prayags: could this extension API server be re-used later for other use cases ? or is this specific to WorkspaceConnection ? 
if specific, we should rename helm value accordingly

internal/aws/ssm_client.go
aws-jasakshi: https://github.com/jupyter-infra/jupyter-k8s/issues/57

hack/helm-patches/values.yaml.patch
aws-jasakshi: Technically it could. It will depend on use case

hack/helm-patches/manager.yaml.patch
```diff
@@ -6,4 +6,7 @@
             - --application-images-registry={{ .Values.application.imagesRegistry }}
             {{- if .Values.accessResources.traefik.enable }}
             - --watch-traefik
+            {{- end}}
+            {{- if .Values.extensionApi.enable }}
+            - --enable-extension-api=true
```
JGuinegagne: this should be passed as a `boolean` flag (look at `--watch-traefik` above. Please add to the GH issue that tracks the fast follows.

PR #42, opened 2025-10-16
feat: pod scheduling support in workspace pod

earaghbidikashani:
## Summary
This PR adds pod scheduling support to the Workspace, enabling users to specify node affinity rules and taint tolerations for fine-grained control over where their Jupyter workspaces
are scheduled.
### API Changes
• **Workspace CRD**: Added affinity, tolerations fields to WorkspaceSpec
• **WorkspaceTemplate CRD**: Added defaultAffinity, defaultTolerations, defaultNodeSelector fields for template-based defaults
• All scheduling fields are optional with proper omitempty tags
### Controller Implementation
• **Template Resolution**: Extended ResolvedTemplate to include scheduling fields with inheritance from templates
• **Deployment Builder**: Updated pod spec generation to apply scheduling constraints with workspace overrides taking precedence over template defaults
### Direct Workspace Configuration
```
apiVersion: [workspaces.jupyter.org/v1alpha1](http://workspaces.jupyter.org/v1alpha1)
kind: Workspace
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: [kubernetes.io/arch](http://kubernetes.io/arch)
            operator: In
            values: ["amd64"]
  tolerations:
  - key: "dedicated"
    operator: "Equal"
    value: "jupyter"
    effect: "NoSchedule"
```
### Template-Based Defaults
```
apiVersion: [workspaces.jupyter.org/v1alpha1](http://workspaces.jupyter.org/v1alpha1)
kind: WorkspaceTemplate
spec:
  defaultTolerations:
  - key: "spot-instance"
    operator: "Exists"
    effect: "NoSchedule"
```

### DIFF
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -116,9 +116,25 @@ func (db *DeploymentBuilder) buildPodSpec(workspace *workspacesv1alpha1.Workspac
 		})
 	}
 
-	// Set node selector if specified
+	// Set node selector - workspace spec takes precedence over template
 	if len(workspace.Spec.NodeSelector) > 0 {
 		podSpec.NodeSelector = workspace.Spec.NodeSelector
+	} else if resolvedTemplate != nil && len(resolvedTemplate.NodeSelector) > 0 {
+		podSpec.NodeSelector = resolvedTemplate.NodeSelector
+	}
+
+	// Set affinity - workspace spec takes precedence over template
+	if workspace.Spec.Affinity != nil {
+		podSpec.Affinity = workspace.Spec.Affinity
+	} else if resolvedTemplate != nil && resolvedTemplate.Affinity != nil {
+		podSpec.Affinity = resolvedTemplate.Affinity
+	}
+
+	// Set tolerations - workspace spec takes precedence over template
+	if len(workspace.Spec.Tolerations) > 0 {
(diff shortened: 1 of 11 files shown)


internal/controller/deployment_builder.go
```diff
@@ -116,9 +116,25 @@ func (db *DeploymentBuilder) buildPodSpec(workspace *workspacesv1alpha1.Workspac
 		})
 	}
 
-	// Set node selector if specified
+	// Set node selector - workspace spec takes precedence over template
 	if len(workspace.Spec.NodeSelector) > 0 {
 		podSpec.NodeSelector = workspace.Spec.NodeSelector
+	} else if resolvedTemplate != nil && len(resolvedTemplate.NodeSelector) > 0 {
```
joshuatowner: pattern here is to use helper methods to find the final values (see getStorageConfig for example)

joshuatowner (approved):

internal/controller/deployment_builder.go
JGuinegagne: More broadly, we should look in injecting the default values in the `Workspace` specs with a mutating admission webhook. However we can discuss.

PR #50, opened 2025-10-17
feat: admission webhook for workspace access control

henrywa2:
This PR implements a validating webhook that enforces access control for workspaces based on their accessType field and ensures annotation immutability.

## Changes
- Added WorkspaceCustomValidator with access control logic
- Configured CLUSTER_ADMIN_GROUP environment variable via Helm
- Implemented username sanitization for special characters
- Refactored validation to avoid early returns for future extensibility

## Testing
- Webhook validates Public workspaces can be modified by anyone
- OwnerOnly workspaces restricted to owner or cluster admins
- Annotations created-by and last-updated-by are immutable after creation

PR #59, opened 2025-10-21
feat: workspace access validation webhook

henrywa2:
Reapplying https://github.com/jupyter-infra/jupyter-k8s/pull/50 and implementing follow up fixes:

- Fixed tests
- Allow last-updated-by annotation to be mutable
- Created constants file
- Moved logic to a validateEditPermission helper function
- Added DefaultAccessType to WorkspaceTemplate

Current behavior:
- For Public AccessType, anyone with RBAC permissions can modify
- If AccessType doesn't exist, behavior is "Public". 
- Admin group can always modify created-by annotation
- system:masters is always considered admin group
- created-by annotation cannot be edited or removed

Follow up:
- Change annotation name to include api group prefix for clarity
- Change API group to workspace.jupyter.org

### DIFF
--- a/internal/webhook/v1alpha1/workspace_webhook.go
+++ b/internal/webhook/v1alpha1/workspace_webhook.go
@@ -45,6 +45,7 @@ func sanitizeUsername(username string) string {
 }
 
 // getEffectiveOwnershipType returns the effective access type, treating empty as Public
+// TODO: think of better way to convey defaults to user.
 func getEffectiveOwnershipType(ownershipType string) string {
 	if ownershipType == "" {
 		return webhookconst.OwnershipTypePublic
@@ -52,8 +53,24 @@ func getEffectiveOwnershipType(ownershipType string) string {
 	return ownershipType
 }
 
-// validateEditPermission checks if the user has permission to modify/delete an OwnerOnly workspace
-func validateEditPermission(ctx context.Context, workspace *workspacesv1alpha1.Workspace) error {
+// isAdminUser checks if the user groups include any admin groups
+func isAdminUser(groups []string) bool {
+	adminGroups := []string{"system:masters"}
+	if clusterAdminGroup := os.Getenv("CLUSTER_ADMIN_GROUP"); clusterAdminGroup != "" {
+		adminGroups = append(adminGroups, clusterAdminGroup)
+	}
+	for _, group := range groups {
+		for _, adminGroup := range adminGroups {
+			if group == adminGroup {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// validateOwnershipPermission checks if the user has permission to modify/delete an OwnerOnly workspace
+func validateOwnershipPermission(ctx context.Context, workspace *workspacesv1alpha1.Workspace) error {
 	req, err := admission.RequestFromContext(ctx)
 	if err != nil {
 		return fmt.Errorf("unable to extract user information from request context: %w", err)
@@ -67,11 +84,8 @@ func validateEditPermission(ctx context.Context, workspace *workspacesv1alpha1.W
 	}
 
 	// Check if user is cluster admin
-	clusterAdminGroup := os.Getenv("CLUSTER_ADMIN_GROUP")
-	for _, group := range req.UserInfo.Groups {
-		if group == clusterAdminGroup {
-			return nil
-		}
+	if isAdminUser(req.UserInfo.Groups) {
+		return nil
 	}
 
 	return fmt.Errorf("access denied: only workspace owner or cluster admins can modify OwnerOnly workspaces")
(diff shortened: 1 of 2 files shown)


api/v1alpha1/workspace_types.go
```diff
@@ -90,6 +90,12 @@ type WorkspaceSpec struct {
 	// +kubebuilder:validation:Enum=Running;Stopped
 	DesiredStatus string `json:"desiredStatus,omitempty"`
 
+	// OwnershipType specifies who can modify the space.
+	// Public means anyone with RBAC permissions can update/delete the space.
+	// OwnerOnly means only the creator can update/delete the space.
+	// +kubebuilder:validation:Enum=Public;OwnerOnly
```
JGuinegagne: should this be `+optional` and default to `Public`?

That would be my vote but feel free to poll the team in the channel.

henrywa2: One concern I had was that in the current user story (default OwnerOnly), if the field is missing it kind of enforces a restrictive behavior which may not be what customers expect. To me, intuitively if the field is missing it should be public behavior and anyone with RBAC permissions can delete it. But maybe others feel differently

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -115,14 +160,38 @@ func (v *WorkspaceCustomValidator) ValidateCreate(_ context.Context, obj runtime
...
+			}
+		}
+	}
+
+	// Check ownershipType transition rules first
+	oldOwnershipType := getEffectiveOwnershipType(oldWorkspace.Spec.OwnershipType)
+	newOwnershipType := getEffectiveOwnershipType(newWorkspace.Spec.OwnershipType)
+	if oldOwnershipType == webhookconst.OwnershipTypePublic && newOwnershipType == webhookconst.OwnershipTypeOwnerOnly {
+		return nil, fmt.Errorf("cannot change ownershipType from Public to OwnerOnly")
+	}
```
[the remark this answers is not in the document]
henrywa2: Short answer is that this was a requirement before migration to OSS. Do you think we should revisit this?

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -28,12 +30,53 @@ import (
...
+	// Use Go's JSON marshaling to properly escape the string
+	escaped, _ := json.Marshal(username)
+	// Remove the surrounding quotes that json.Marshal adds
+	return string(escaped[1 : len(escaped)-1])
+}
+
+// getEffectiveOwnershipType returns the effective access type, treating empty as Public
+func getEffectiveOwnershipType(ownershipType string) string {
+	if ownershipType == "" {
+		return webhookconst.OwnershipTypeOwnerOnly
```
joshuatowner: nit: follow up item for the best place to store and communicate defaults. imo we should try to mutate the space so the ownership is always clear

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -115,14 +160,38 @@ func (v *WorkspaceCustomValidator) ValidateCreate(_ context.Context, obj runtime
...
+	}
+	newWorkspace, ok := newObj.(*workspacesv1alpha1.Workspace)
 	if !ok {
 		return nil, fmt.Errorf("expected a Workspace object for the newObj but got %T", newObj)
 	}
-	workspacelog.Info("Validation for Workspace upon update", "name", workspace.GetName())
 
-	// TODO(user): fill in your validation logic upon object update.
+	// Validate that ownership annotations are immutable
+	if oldWorkspace.Annotations != nil && newWorkspace.Annotations != nil {
```
joshuatowner: does this allow any kind of bypass by deleting / creating all annotations?

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -28,12 +30,53 @@ import (
...
+// getEffectiveOwnershipType returns the effective access type, treating empty as Public
+func getEffectiveOwnershipType(ownershipType string) string {
+	if ownershipType == "" {
+		return webhookconst.OwnershipTypeOwnerOnly
+	}
+	return ownershipType
+}
+
+// validateEditPermission checks if the user has permission to modify/delete an OwnerOnly workspace
+func validateEditPermission(ctx context.Context, workspace *workspacesv1alpha1.Workspace) error {
```
joshuatowner: nit: update name on this to indicate it checks ownership

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -28,12 +30,53 @@ import (
...
 	workspacesv1alpha1 "github.com/jupyter-ai-contrib/jupyter-k8s/api/v1alpha1"
 	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/controller"
+	webhookconst "github.com/jupyter-ai-contrib/jupyter-k8s/internal/webhook"
 )
 
 // nolint:unused
 // log is for logging in this package.
 var workspacelog = logf.Log.WithName("workspace-resource")
 
+func sanitizeUsername(username string) string {
```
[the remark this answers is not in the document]
henrywa2: When I added \n and \t to the username in EKS access entries, it showed up in a new line when describing the workspace. Not sure if that's an issue but wanted to avoid it if possible

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -28,12 +30,53 @@ import (
...
+
+	// Check if user is the owner
+	if workspace.Annotations != nil {
+		if createdBy := workspace.Annotations[controller.AnnotationCreatedBy]; createdBy == sanitizeUsername(req.UserInfo.Username) {
+			return nil
+		}
+	}
+
+	// Check if user is cluster admin
+	clusterAdminGroup := os.Getenv("CLUSTER_ADMIN_GROUP")
```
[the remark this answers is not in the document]
henrywa2: follow up item is to always allow `system:masters` as an Admin group. But it should be optional. I'll make sure this works even if env variable isn't passed.

internal/webhook/v1alpha1/workspace_webhook.go
henrywa2: I don't think this would catch the case of deleting all annotations. Will fix

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -115,14 +160,38 @@ func (v *WorkspaceCustomValidator) ValidateCreate(_ context.Context, obj runtime
...
+	newWorkspace, ok := newObj.(*workspacesv1alpha1.Workspace)
 	if !ok {
 		return nil, fmt.Errorf("expected a Workspace object for the newObj but got %T", newObj)
 	}
-	workspacelog.Info("Validation for Workspace upon update", "name", workspace.GetName())
 
-	// TODO(user): fill in your validation logic upon object update.
+	// Validate that ownership annotations are immutable
+	if oldWorkspace.Annotations != nil && newWorkspace.Annotations != nil {
+		if oldCreatedBy := oldWorkspace.Annotations[controller.AnnotationCreatedBy]; oldCreatedBy != "" {
```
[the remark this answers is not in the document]
henrywa2: We will allow admins to update it

internal/webhook/v1alpha1/workspace_webhook_test.go
```diff
@@ -229,6 +229,60 @@ var _ = Describe("Workspace Webhook", func() {
...
+				"other-annotation":             "other-value",
+			}
+			newWorkspace := workspace.DeepCopy()
+			newWorkspace.Annotations = nil
+
+			warnings, err := validator.ValidateUpdate(ctx, oldWorkspace, newWorkspace)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("created-by annotation cannot be removed"))
+			Expect(warnings).To(BeEmpty())
+		})
```
JGuinegagne: sanity-check: do you think we need to unit test setting the `created-by` annotation to ""?

non-blocking: can be future PRs

JGuinegagne (approved):

aws-jasakshi (approved):

PR #60, opened 2025-10-22
feat: add discovery endpoint for extension api server

aws-jasakshi:
Changes
- Add extension API server with discovery endpoint
- Support CREATE operations for workspace connections  
- Add boolean flag fix for `--enable-extension-api`
- Add basic unit test for discovery endpoint

### DIFF
--- a/internal/extensionapi/server.go
+++ b/internal/extensionapi/server.go
@@ -1,3 +1,4 @@
+// Package extensionapi provides extension API server functionality.
 package extensionapi
 
 import (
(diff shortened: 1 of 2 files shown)


JGuinegagne (approved):
LGTM as scaffolding.

We may want a bit more careful handling of the server process ([example](https://github.com/jupyter-infra/jupyter-k8s/blob/da0d16363f740ac0dcc393a728acfd35ee2862ee/internal/authmiddleware/server.go#L33)), but we can iterate.

internal/extensionapi/server.go
```diff
@@ -0,0 +1,68 @@
...
+		setupLog.Info("Extension API server starting on :8444")
+		if err := http.ListenAndServeTLS(":8444", "/tmp/extension-server/serving-certs/tls.crt", "/tmp/extension-server/serving-certs/tls.key", nil); err != nil {
+			setupLog.Error(err, "Extension API server failed")
+		}
+	}()
+
+	return nil
+}
+
+func handleDiscovery(w http.ResponseWriter, _ *http.Request) {
```
joshuatowner: lets put each route in a new file

PR #73, opened 2025-10-23
fix: OwnershipType bugs and fast follows

henrywa2:
Fixes:

- Enabling webhook for delete
- Moving system:masters group to constants file
- Fixed bug where users were allowed to change ownershipType from OwnerOnly->Public when they weren't supposed to
- Fixed mutating webhook to not change created-by annotation on update call, previously it was re-creating the annotation if the Admin deleted it
- Added unit tests
- Fixed bug where user could update a workspace from public->owneronly despite not being the creator, and then they wouldn't be able to update it again.

### DIFF
--- a/internal/authmiddleware/config.go
+++ b/internal/authmiddleware/config.go
@@ -25,6 +25,7 @@ const (
 	EnvJwtExpiration     = "JWT_EXPIRATION"
 	EnvJwtRefreshWindow  = "JWT_REFRESH_WINDOW"
 	EnvJwtRefreshHorizon = "JWT_REFRESH_HORIZON"
+	EnvEnableBearerAuth  = "ENABLE_BEARER_URL_AUTH"
 
 	// Cookie configuration
 	EnvCookieName     = "COOKIE_NAME"
@@ -64,6 +65,7 @@ const (
 	DefaultJwtExpiration     = 1 * time.Hour
 	DefaultJwtRefreshWindow  = 15 * time.Minute // 25% of the default expiration
 	DefaultJwtRefreshHorizon = 12 * time.Hour
+	DefaultEnableBearerAuth  = false
 
 	// Cookie defaults
 	DefaultCookieName     = "workspace_auth"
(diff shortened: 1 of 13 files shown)


internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -206,9 +215,19 @@ func (v *WorkspaceCustomValidator) ValidateUpdate(ctx context.Context, oldObj, n
 		}
 	}
 
+	originalOwnershipType := getEffectiveOwnershipType(oldWorkspace.Spec.OwnershipType)
 	newOwnershipType := getEffectiveOwnershipType(newWorkspace.Spec.OwnershipType)
+	workspacelog.Info("Ownership validation check", "originalType", originalOwnershipType, "newType", newOwnershipType, "workspace", newWorkspace.GetName())
```
JGuinegagne: could you check the logs? I'm curious in particular if you already get the `Workspace.Name` from context (and the namespace). If not, add both name and namespace please. If they are there, remove the duplicate and rely on the context please.

internal/webhook/v1alpha1/workspace_webhook_test.go
```diff
@@ -213,24 +224,73 @@ var _ = Describe("Workspace Webhook", func() {
...
+			newWorkspace.Spec.OwnershipType = webhookconst.OwnershipTypeOwnerOnly
+			newWorkspace.Annotations = map[string]string{
+				controller.AnnotationCreatedBy: "original-user",
+			}
+
+			warnings, err := validator.ValidateUpdate(nonOwnerCtx, oldWorkspace, newWorkspace)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("access denied"))
+			Expect(warnings).To(BeEmpty())
+		})
```
JGuinegagne: Could you add a UT to ensure that an admin can do it please.

internal/webhook/constants.go
```diff
@@ -6,3 +6,8 @@ const (
 	OwnershipTypeOwnerOnly = "OwnerOnly"
 	OwnershipTypePublic    = "Public"
 )
+
+// Admin group constants
+const (
+	DefaultAdminGroup = "system:masters"
```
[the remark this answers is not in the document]
henrywa2: will do

joshuatowner (approved):

PR #75, opened 2025-10-23
feat: Add extension API server for workspace connections

aws-jasakshi:
## Changes
- Extension API server with discovery and connection endpoints
- VS Code URL generation with SSM session integration
- Pod lookup using workspace labels
- RBAC role for connection permissions

## Testing
tested VS Code URL generation with real workspace pod


internal/extensionapi/utils.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func getPodUIDFromSpaceName(spaceName string) (string, error) {
```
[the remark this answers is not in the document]
aws-prayags: also nit: this method is more of a workspace util rather than extension api util

internal/extensionapi/config.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// Config holds configuration for the extension API server
+type Config struct {
+	ClusterARN string
+}
+
+// NewConfig creates a new configuration from environment variables
+func NewConfig() (*Config, error) {
+	clusterARN := os.Getenv(aws.ClusterARNEnv)
```
aws-prayags: this field is not something always needed for extensionapi - so this class may not be best place

eksClusterArn is similar to ssm document - both are needed specifically to generate url for a specific access strategy - we can simplify for now by reading directly from env variables and removing this class.

config/rbac/connection_role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
aws-prayags: 1/ do we also plan to add a RoleBinding here ?
2/ I see Role is being created in `default` namespace - is that intentional ?

internal/extensionapi/route_connections.go
```diff
@@ -0,0 +1,144 @@
...
+				setupLog.Error(err, "Failed to generate connection URL", "ide", ide)
+				http.Error(w, "Failed to generate connection URL", http.StatusInternalServerError)
+				return
+			}
+		}
+
+		// Create response
+		response := map[string]interface{}{
+			"apiVersion": "connection.workspaces.jupyter.org/v1alpha1",
+			"kind":       "WorkspaceConnection",
```
aws-prayags: move values to constants in extensionapi module

internal/extensionapi/route_connections.go
```diff
@@ -0,0 +1,144 @@
...
+			"metadata":   req["metadata"],
+			"spec":       spec,
+			"status": map[string]interface{}{
+				"type": responseType,
+				"url":  responseURL,
```
aws-prayags: suggestion along same lines as for request
`workspaceConnectionType`
`workspaceConnectionUrl`

PR #79, opened 2025-10-23
feat: workspace access permission

JGuinegagne:
This PR adds fine-grained access control for Workspace web-access.

Also in this PR:
- refactor the extensionapi server to 1/ use the manager for process monitoring, 2/ allow configuration, 3/ change API group to `connection.workspace.jupyter.org` (singular)

### User experience
- if a user sets the `Workspace.Spec.OwnershipType=OwnerOnly` (or omit), then only that particular owner may access this app, provided that they have `create workspaces/connections` RBAC permission
- if a user sets the `Workspace.Spec.OwnershipType=Public`, then any users with RBAC permission can access

### Implementation
1. added a new route to extensionapi: `POST /apis/connection.workspace.jupyter.org/v1alpha1/namespaces/ns/connectionaccessreview`, RBAC action `{verb: create, resource: ConnectionAccessReview}`
   1.1. accept `user` and `groups` (future work: `extras`) parameter, and `workspaceName`
   1.2. this route first performs a [SubjectAccessReview](https://dev-k8sref-io.web.app/docs/authorization/subjectaccessreview-v1/#subjectaccessreviewspec)
   1.3. if pass, performs the equivalent of an admission webhook for connection: public workspace allowed, ownerOnly workspace require username to match owner annotation
   1.4. if workspace does not exist, disallow access
   1.5. return `{status: {allowed, notFound, reason}}`
2. authmiddleware `/auth` route calls `Create:ConnectionAccessReview` using the auth header (username, group)
3. on success, issues a cookie as before
4. refactor and extended `extensionapi`
5. generated `ConnectionAccessReview`
   5.1 using `kubebuilder` as if it was a CRD (it isn't, we don't want record in etcd and `create` must respond a body which k8s standard API do not allow)
   5.2. excluded the generated CRD yaml from kustomize/helm 
6. added process monitor and config to extensionapi
7. switched extensionapi port to 7443 (from 8444), to convey "it's a TLS connection" 

### Future work
- pass `extra` to `Create:ConnectionAccessReview`
- call verify again on cookie refresh
- switch to logic depend on `Workspace.Spec.AccessType` attribute
- peer-to-peer sharing (post JupyterCon) 
- unit test coverage for `extensionapi`
- e2e test coverage for 1/ extensionapi, 2/ public/private access
- e2e test coverage for authmiddleware

### Testing
- extensive unit tests coverage for `authmiddleware` handling of k8s RestClient
- tested extension API against kind cluster
- tested extension API against EKS cluster
- tested end to end against EKS cluster w/ guided deployment
   - public workspace: verified anyone w/ RBAC may access
   - private workspace: verified only owner may access

### Example
```bash
kubectl create --raw /apis/connection.workspace.jupyter.org/v1alpha1/namespaces/default/connectionaccessreview -f - <<EOF
{
    "apiVersion": "connection.workspace.jupyter.org/v1alpha1",
    "kind": "ConnectionAccessReview",
    "metadata": {
      "namespace": "default"
    },
    "spec": {
      "workspaceName": "sample-workspace",
      "user": "kubernetes-admin",
      "groups": ["kubeadm:cluster-admins", "system:authenticated"]
    }
}
EOF
{"kind":"ConnectionAccessReview","apiVersion":"connection.workspace.jupyter.org/v1alpha1","metadata":{"namespace":"default"},"spec":{"workspaceName":"sample-workspace","groups":["kubeadm:cluster-admins","system:authenticated"],"user":"kubernetes-admin"},"status":{"allowed":false,"notFound":true,"reason":"Workspace not found"}}
```

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: jupyter.org/jupyter-k8s
+  newTag: v0.0.1
--- a/config/samples_routing/kustomization.yaml
+++ b/config/samples_routing/kustomization.yaml
@@ -4,4 +4,4 @@ kind: Kustomization
 resources:
   - workspace_access_strategy.yaml
   - workspace_with_access_strategy.yaml
-  - workspace_with_access_strategy2.yaml
+  - workspace_with_access_strategy_private.yaml
--- a/config/samples_routing/workspace_with_access_strategy.yaml
+++ b/config/samples_routing/workspace_with_access_strategy.yaml
@@ -9,6 +9,7 @@ spec:
   # Using the built-in 'uv' image shortcut
   image: uv
   desiredStatus: Running
+  ownershipType: Public
   accessStrategy:
     name: sample-access-strategy
     namespace: default
--- a/config/samples_routing/workspace_with_access_strategy_private.yaml
+++ b/config/samples_routing/workspace_with_access_strategy_private.yaml
@@ -1,7 +1,7 @@
 apiVersion: workspaces.jupyter.org/v1alpha1
 kind: Workspace
 metadata:
-  name: alt-ws-with-access
+  name: private-ws-with-access
   namespace: default
 spec:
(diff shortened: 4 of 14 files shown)


joshuatowner (approved):
will follow up with detailed feedback

Issue #36, opened 2025-10-14
PR20 fast follows


JGuinegagne:
Follow ups on #20 
1. add generic helm parameter for Group-Kind-Version (GVK[]) to watch ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2426499546))
2. track deletion of accessResources in status ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2427610473))
3. evaluate probe in `AccessStrategy` to verify route registration ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2423348154))
4. evaluate redirects instead of path stripping ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2428859096))
5. add source verification for middleware trusted headers ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2428874141))
6. recheck access permissions on cookie refresh ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2430082751))
7. evaluate single name cookie w/ difference based on path ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2423398944))
8. implement `Workspace` access idempotency (ie delete workspace, recreate w/ same name but different access)
9. implement `AccessResourceSelector` in `Workspace.Status` ([ref](https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/20#discussion_r2426562470))

PR #98, opened 2025-10-28
fix: auth-context, cookie refresh

JGuinegagne:
This PR implements fast follow from #79 
1. support full auth context for `ConnectionAccessReview`
2. verify user permissions on cookie refresh (see 6/ in #36 )

### User experience
The only difference in UX is that if the call to `Create:ConnectionAccessReview` returns `allowed=false` on cookie refresh, the `authmiddleware` returns a 403 and the request get denied.

### Implementation
- add `UID` and `Extra` to `ConnectionAccessReview.Spec` and pass them to [SubjectAccessReview](https://dev-k8sref-io.web.app/docs/authorization/subjectaccessreview-v1/)
- read `UID` from header in the `/auth` route of `authmiddleware`, embed in JWT
- in `/verify` route of `authmiddleware`, read user auth context from JWT, call `Create:ConnectionAccessReview`
    - if `allowed=true`, refresh cookie
    - if `allowed=false`, return 403
    - if `Create:ConnectionAccessReview` errors out, set `skipRefresh` in JWT to avoid failing on every subsequent requests
- add `jwtRefreshEnable` flag in helm chart, pass to `authmiddleware` as env var and ultimately to config

### Testing
- added unit tests
- tested directly calling `Create:ConnectionAccessReview` w/ and w/o the `spec.uid` and `spec.extra`
- tested in calling `Create:ConnectionAccessReview` within AWS EKS where extra is required --> confirmed access
- updated guided chart `aws-traefik-dex`, ensure Workspace authZ still worked as expected

### Sample `Create:ConnectionAccessReview` (against kind cluster):
```
kubectl create --raw /apis/connection.workspace.jupyter.org/v1alpha1/namespaces/default/connectionaccessreview -f - <<EOF
{
    "apiVersion": "connection.workspace.jupyter.org/v1alpha1",
    "kind": "ConnectionAccessReview",
    "metadata": {
      "namespace": "default"
    },
    "spec": {
      "workspaceName": "workspace-sample",
      "user": "kubernetes-admin",
      "groups": ["kubeadm:cluster-admins", "system:authenticated"]
    }
}
EOF
{"kind":"ConnectionAccessReview","apiVersion":"connection.workspace.jupyter.org/v1alpha1","metadata":{"namespace":"default"},"spec":{"workspaceName":"workspace-sample","groups":["kubeadm:cluster-admins","system:authenticated"],"user":"kubernetes-admin"},"status":{"allowed":true,"notFound":false,"reason":"Valid RBAC and user is the owner of the private Workspace"}}
```

### DIFF
--- a/internal/authmiddleware/serverroute_verify.go
+++ b/internal/authmiddleware/serverroute_verify.go
@@ -79,7 +79,7 @@ func (s *Server) handleVerify(w http.ResponseWriter, r *http.Request) {
 			} else {
 				// Set refreshed cookie with the same path as the original token
 				s.cookieManager.SetCookie(w, newToken, claims.Path)
-				s.logger.Debug("Token refreshed successfully", "user", claims.User, "path", claims.Path)
+				s.logger.Info("Token refreshed successfully", "user", claims.User, "path", claims.Path)
 			}
 			// UNHAPPY CASE 2: user is no longer allowed, return 403
 		} else if !accessReviewResult.Allowed {
@@ -91,6 +91,7 @@ func (s *Server) handleVerify(w http.ResponseWriter, r *http.Request) {
 				workspaceInfo.Namespace,
 				"reason",
 				accessReviewResult.Reason)
+			s.cookieManager.ClearCookie(w, claims.Path)
 			http.Error(w, "Access denied: you are no longer authorized to access this workspace", http.StatusForbidden)
 			return
 			// HAPPY CASE: user is allowed, refresh their cookie
@@ -103,7 +104,7 @@ func (s *Server) handleVerify(w http.ResponseWriter, r *http.Request) {
 			} else {
 				// Set refreshed cookie with the same path as the original token
 				s.cookieManager.SetCookie(w, newToken, claims.Path)
-				s.logger.Debug("Token refreshed successfully", "user", claims.User, "path", claims.Path)
+				s.logger.Info("Token refreshed successfully", "user", claims.User, "path", claims.Path)
 			}
 		}
 	}
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2025 jupyter-ai-contrib
+Copyright (c) 2025 Amazon Web Services
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ make test-e2e
 
 MIT License
 
-Copyright (c) 2025 jupyter-ai-contrib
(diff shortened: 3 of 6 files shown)


CLAUDE.md
```diff
@@ -23,7 +23,22 @@ The Kubernetes controller implements the operator pattern to manage workspace re
...
+**Code:** `./internal/extensionapi`
+
+In addition to the controller, the jupyter-k8s operator starts an extension API server in the
+same pod as the controller, and managed by the same manager.
+
+The extension API serves APIs under `connection.workspace.jupyter.org` that cannot be represented
+as CRD.
+- `create Connection`
+- `create ConnectionAccessReview`
+`ConnectionAccessReview` performs an RBAC check on a virtual resource `workspaces/connection` in
```
henrywa2: Are we still using workspaces/connection here?

api/connection/v1alpha1/connection_access_review_types.go
```diff
@@ -22,10 +22,11 @@ import (
...
 type ConnectionAccessReviewSpec struct {
-	WorkspaceName string   `json:"workspaceName"`
-	Groups        []string `json:"groups"`
-	UID           string   `json:"uid,omitempty"`
-	User          string   `json:"user"`
+	WorkspaceName string              `json:"workspaceName"`
+	Groups        []string            `json:"groups"`
+	UID           string              `json:"uid,omitempty"`
+	User          string              `json:"user"`
+	Extra         map[string][]string `json:"extra,omitempty"`
```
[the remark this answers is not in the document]
JGuinegagne: This is needed for EKS k8s identity: the permission maps to the extra `sigs.k8s.io/aws-iam-authenticator/principalId`, not to the `username`, `groups` or `uid`.

internal/authmiddleware/serverroute_verify.go
```diff
@@ -67,15 +67,44 @@ func (s *Server) handleVerify(w http.ResponseWriter, r *http.Request) {
...
-		if err != nil {
-			// Log but don't fail the request - just continue with the existing token
-			s.logger.Warn("Failed to refresh token", "error", err)
+		// Verify that the user still has access to the specific Workspace
+		accessReviewResult, workspaceInfo, accessErr := s.VerifyWorkspaceAccessFromJwt(r.Context(), requestPath, claims)
+
+		// UNHAPPY CASE 1: we can't check authZ for some reason, stop attempting to refresh
+		if accessErr != nil {
+			s.logger.Warn("Failed to retrieve the accessReview for cookie refresh", "error", err)
+			newToken, err := s.jwtManager.UpdateSkipRefreshToken(claims)
```
joshuatowner: what's the point of the skip refresh here? won't it try again to refresh on the next try anyways?

i see that the comment here says "stop attempting to refresh", but it seems the method still sets skip token to false?

PR #110, opened 2025-10-29
fix: cleanup hybrid activation for ssm remote access strategy

aws-prayags:
1/ Cleanup hybrid activation also along with managed instances on pod deletion. Since clean up of activation is not critical, prefer to add in to pod deletion handling rather than in init flow
2/ Updated the tags being added to managed instances


internal/aws/ssm_remote_access_strategy.go
```diff
@@ -209,17 +223,24 @@ func (s *SSMRemoteAccessStrategy) createSSMActivation(ctx context.Context, pod *
...
+	// Prepare tags - include SageMaker required tags for policy compliance
 	tags := map[string]string{
-		TagManagedBy:       "jupyter-k8s-operator",
-		TagWorkspaceName:   workspace.Name,
-		TagNamespace:       workspace.Namespace,
-		TagWorkspacePodUID: string(pod.UID),
+		SageMakerManagedByTagKey:  SageMakerManagedByTagValue,
+		SageMakerEKSClusterTagKey: eksClusterARN,
+		TagWorkspaceName:          workspace.Name,
+		TagNamespace:              workspace.Namespace,
```
JGuinegagne: non-blocking: do the `Name` & `Namespace` regex match the AWS tag values regex? If no, add a GH issue tracker to sanitize in please.

PR #111, opened 2025-10-29
fix: Helm installation in all CI workflows

aws-jasakshi:
- Replace get.helm.sh script with direct GitHub releases download
- Apply fix to lint.yml, test-chart.yml, and test-chart-aws-traefik-dex.yml


.github/workflows/lint.yml
```diff
@@ -25,8 +25,7 @@ jobs:
           export PATH=$PATH:$(go env GOPATH)/bin
 
       - name: Install Helm
-        run: |
-          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+        run: sudo snap install helm --classic
```
joshuatowner: not 100% sure on using snap, as it can cause issues due to its containerization. but for now, approving to unblock. would prefer to get this via the package manager, i.e. [this](https://helm.sh/docs/intro/install/#from-apt-debianubuntu)

joshuatowner (approved):

PR #112, opened 2025-10-29
feat: Adding serviceAccount support

henrywa2:
New features:
- serviceAccountName field in workspace spec
- sample serviceAccount with allowlisting for users and groups 
- Admins always have access to use any service account
- Users that match either a username or a group name in the service account have access
- Allowlisted usernames that end with {{SessionName}} check for a valid session name to support default EKS username behavior
- If no service account is passed in, pods will use the namespace "default" service account. Current behavior is that users can use this service account in their pods if they don't pass in a value, but not if they explicitly specify "default" in their workspace spec. This will be updated in a follow up PR to handle default service account.

TODO:
- Implement default service account behavior. We will give Admins the options to label a service account as the default for the namespace, and if no service account is passed into the workspace it will look for a default and use it. If multiple service accounts have the same default label, creation will fail. 
- All users will be allowed to use the default service account


internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -196,6 +191,17 @@ func (v *WorkspaceCustomValidator) ValidateCreate(ctx context.Context, obj runti
 		return nil, err
 	}
 
+	// Admin users bypass user validation
```
joshuatowner: nit: Admin users bypass ~user~ validation

internal/webhook/v1alpha1/service_account_validator.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+	}
+
+	// Check service-account-groups annotation
+	if groupsYaml, exists := sa.Annotations[controller.AnnotationServiceAccountGroups]; exists {
+		var groups []string
+		if err := yaml.Unmarshal([]byte(groupsYaml), &groups); err == nil {
+			for _, group := range groups {
+				for _, userGroup := range userInfo.Groups {
```
JGuinegagne: similarly, should we support `*` and `?` for groups?

JGuinegagne (changes requested):
I think this is too AWS-specific. Would it be possible to implement a simplified, IAM-style wildcards match system with `*` and `?` placeholders only?

internal/webhook/v1alpha1/service_account_validator.go
```diff
@@ -0,0 +1,140 @@
...
+	patternsYaml, exists := sa.Annotations[controller.AnnotationServiceAccountUsersPattern]
+	if !exists {
+		return false
+	}
+	var patterns []string
+	if err := yaml.Unmarshal([]byte(patternsYaml), &patterns); err != nil {
+		return false
+	}
+	for _, pattern := range patterns {
+		if matched, err := filepath.Match(pattern, username); err == nil && matched {
```
joshuatowner: this seems designed for filepaths... is this the right lib to use? any downsides?

joshuatowner:
```
filepath.Match("admin-*", "admin-dev/test")  // false on Unix (/ blocks *)
```
I guess this is okay? we should at least document any edge cases here. or other patterns this supports.

config/samples/serviceaccount_with_allowlisting.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: test-serviceaccount
+  annotations:
+    workspace.jupyter.org/service-account-users: |
+      - arn:aws:sts::561569208013:assumed-role/DSRole/{{SessionName}}
```
joshuatowner: does this need update?

henrywa2: Yes it does but I was going to update it in a follow up

internal/webhook/v1alpha1/service_account_validator.go
henrywa2: I'll look into alternate ways to support this

PR #117, opened 2025-10-30
refactor: use defaulter classes to split up defaulting logic

joshuatowner:
### Notes
- Replaced monolithic ApplyTemplateDefaults method with a modular default applicator approach. 
- Added new tests since this logic was not previously unit tested

### DIFF
--- a/internal/extensionapi/config_test.go
+++ b/internal/extensionapi/config_test.go
@@ -103,5 +103,12 @@ var _ = Describe("ExtensionConfig", func() {
 
 			Expect(config.AllowedOrigin).To(Equal(customOrigin))
 		})
+
+		It("Should allow to override ClusterId", func() {
+			customClusterId := "test-cluster-123"
+			config := NewConfig(WithClusterId(customClusterId))
+
+			Expect(config.ClusterId).To(Equal(customClusterId))
+		})
 	})
 })
--- a/internal/extensionapi/connection_route.go
+++ b/internal/extensionapi/connection_route.go
@@ -13,9 +13,11 @@ import (
 	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/workspace"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
+
+// Function variable for SSM strategy creation (mockable in tests)
+var newSSMRemoteAccessStrategy = aws.NewSSMRemoteAccessStrategy
 
 // noOpPodExec is a no-op implementation of PodExecInterface for connection URL generation
 type noOpPodExec struct{}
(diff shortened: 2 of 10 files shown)


internal/webhook/v1alpha1/metadata_defaulter.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/controller"
+)
+
+// applyMetadataDefaults applies metadata defaults (labels and annotations) from template to workspace
+func applyMetadataDefaults(workspace *workspacev1alpha1.Workspace, template *workspacev1alpha1.WorkspaceTemplate) {
+	// Add template tracking label
+	if workspace.Labels == nil {
+		workspace.Labels = make(map[string]string)
+	}
+	workspace.Labels[controller.LabelWorkspaceTemplate] = template.Name
```
andrii-i: This is a good refactor. Without this PR label is created based on `workspace.Spec.TemplateRef.Name` not actual `template.Name` therefore potentially introducing labels diverging from user's immutable spec if template metadata changes.

andrii-i (approved):
Thank you @joshuatowner. No immediate change requests but please find a couple of comments for information and consideration:  
* Not an immediate change request but PR #129 changes `TemplateRef` from *string to struct (with `name` and optional `namespace` properties) so if it will be merged first `TemplateRef` use will need adjustment everywhere in the PR including testing
* Let's track this together with other places that need e2e coverage

internal/webhook/v1alpha1/core_defaulter.go
```diff
@@ -0,0 +1,39 @@
...
+func applyCoreDefaults(workspace *workspacev1alpha1.Workspace, template *workspacev1alpha1.WorkspaceTemplate) {
+	// Apply image defaults
+	if workspace.Spec.Image == "" && template.Spec.DefaultImage != "" {
+		workspace.Spec.Image = template.Spec.DefaultImage
+	}
+
+	// Apply ownership type defaults
+	if workspace.Spec.OwnershipType == "" && template.Spec.DefaultOwnershipType != "" {
+		workspace.Spec.OwnershipType = template.Spec.DefaultOwnershipType
+	}
```
JGuinegagne: optional: add `accessType` here?

internal/webhook/v1alpha1/template_defaulter.go
```diff
@@ -0,0 +1,79 @@
...
+
+// NewTemplateDefaulter creates a new TemplateDefaulter
+func NewTemplateDefaulter(k8sClient client.Client) *TemplateDefaulter {
+	return &TemplateDefaulter{
+		client: k8sClient,
+	}
+}
+
+// defaultApplicators is the registry of all default applicators
+var defaultApplicators = []DefaultApplicator{
```
JGuinegagne: sanity-check: shouldn't that be a `const` ?

internal/webhook/v1alpha1/template_defaulter.go
```diff
@@ -0,0 +1,79 @@
...
+
+// ApplyTemplateDefaults applies template defaults to workspace
+func (td *TemplateDefaulter) ApplyTemplateDefaults(ctx context.Context, workspace *workspacev1alpha1.Workspace) error {
+	if workspace.Spec.TemplateRef == nil {
+		return nil
+	}
+
+	template, err := td.fetchTemplate(ctx, *workspace.Spec.TemplateRef)
+	if err != nil {
+		return err
```
JGuinegagne: question: if we get here, what happens? Does the `Workspace` get created (because the webhook failed but did not return "no") ?

internal/webhook/v1alpha1/core_defaulter.go
joshuatowner: will add this in follow up if it doesn't come in rebase (I already need to add LCC & app type, I will do full scan)

Issue #99, opened 2025-10-28
Add `WorkspaceTemplate` GPU bounds validation


andrii-i:
> Missing GPU bounds validation  / this function does not check GPU bounds. We also need to add E2E test for it or track the need in the issue.
> 
> Nit: for extra cleanliness separate validation logic for CPU, GPU etc can be broken out into separate helper functions but it's also always a bit of a tradeoff to this so up to you 

 _Originally posted by @andrii-i in [#86](https://github.com/jupyter-infra/jupyter-k8s/pull/86/files/2547b2d5e64ea06bc1f66f450729f5ab7423f805#r2466412462)_

JGuinegagne:
Addressed by #287 (generic resource-bounds validation) and #346 (validate limits against template bounds): `validateResourceListBounds` now covers all resource names generically, and `resource_validator_test.go` has explicit `nvidia.com/gpu` cases.

PR #129, opened 2025-10-30
feat: make templates mutable, remove template enforcement from controller logic, make `templateRef` an object, refactor tests

### LINKED ISSUE #19: Consider user experience around template mutability and workspace reconciliation
this could definitely be annoying for customers... how do other similar patterns handle this? every time someone wants to add a new image, they have to create a whole new template.

_Originally posted by @joshuatowner in https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/10#discussion_r2411972880_

### LINKED ISSUE #87: Workspace template operator validation
# Remove Template Resolver After Webhook Migration

## Current Status
Template resolver is still required for one remaining use case: applying environment variables from templates during deployment creation.
Webhook PR: https://github.com/jupyter-infra/jupyter-k8s/pull/86

## Blocker
The webhook's ApplyTemplateDefaults function handles most template fields but does not apply EnvironmentVariables from WorkspaceTemplate.spec.environmentVariables. The template resolver currently handles this
during deployment creation in the controller.

## Prerequisites for Removal

1. Add EnvironmentVariables field to WorkspaceSpec:
```
   // EnvironmentVariables specifies environment variables for the workspace container
   // +optional
   EnvironmentVariables []corev1.EnvVar json:"environmentVariables,omitempty"
```

2. Update webhook to apply environment variables:

```
   if workspace.Spec.EnvironmentVariables == nil && len(template.Spec.EnvironmentVariables) > 0 {
       workspace.Spec.EnvironmentVariables = make([]corev1.EnvVar, len(template.Spec.EnvironmentVariables))
       copy(workspace.Spec.EnvironmentVariables, template.Spec.EnvironmentVariables)
   }
```

3. Update controller to use workspace spec directly instead of template resolver

### LINKED ISSUE #91: Field specific resource validation in admission webhook
question: does it perform deep checks?

For example, say that we want to enforce deep-compliance check
[FAIL] case
- pre-update resource: `{cpu: 4, memory: 2Gi}`
- update resource request: `{cpu: 5, memory: 2Gi}`
- template: `{maxCpu: 4, maxMemory: 4Gi}`

[PASS] case
- pre-update resource: `{cpu: 4, memory: 2Gi}`
- update resource request: `{cpu: 5, memory: 2Gi}`
- template: `{maxCpu: 5, maxMemory: 1Gi}`

Rationale: the `maxMemory` bound is grandfathered.
We can discuss whether that's the behavior that we want. Whatever we decide, please add UTs to cover these cases.

_Originally posted by @JGuinegagne in https://github.com/jupyter-infra/jupyter-k8s/pull/86#discussion_r2466905714_

### LINKED ISSUE #201: Add separate env variables for disabling workspace and template webhooks
### Problem

`ENABLE_WEBHOOKS` env var check on if set disables both Workspace and WorkspaceTemplate webhooks. This does not allow administrators granular control to disable either Workspace or WorkspaceTemplate webhook, only both.  

Template is created as part of helm install. Enabling webhooks means it install has to wait for image pull and other prcoesses which can create install issues in some deployments / environments.

### Proposed solution

Add another env var / separate env vars for disabling workspace and template webhooks https://github.com/jupyter-infra/jupyter-k8s/pull/129/files/0f5fb9a2f04ee92079696b35ec2a829edf92e39b#r2505796240

### References

As discussed in https://github.com/jupyter-infra/jupyter-k8s/pull/129/files/0f5fb9a2f04ee92079696b35ec2a829edf92e39b#r2505796240

andrii-i:
### Description

Enable template updates without recreation by removing immutability constraint and moving validation to webhooks. This PR allows users and operators to update templates without recreating them improving user experience, enabling easier and safer template evolution, better operational flexibility.

* Make `WorkspaceTemplate` mutable (was immutable via webhook validation)
* Change `templateRef` from `string` to `struct {name, namespace}` for future extensibility
* Remove template enforcement from controller (validation moved to webhooks only) to improve performance and streamline architecture following Kubernetes best practices
* Use static yaml test files instead of defining them inline in `workspacetemplate_test.go`
* Update or remove `Workspace` template label if `templateRef` is updated or removed
* Add comprehensive unit and e2e tests covering functionality changed in the PR
* On workspace update:

```
  - TemplateRef deleted - no validation
  - TemplateRef changed - validation vs template
  - TemplateRef added - validation vs template
  - No templateRef, not updated - no validation
  - TemplateRef unchanged - check other conditions Check if any spec field changed
       * No spec changes - skip validation (metadata-only update)
       * If ONLY DesiredStatus changed to Stopped, allow without validation
       * If other spec fields also changed, those changes must be validated
```
* Add separate env consts controlling if webhooks are initialized (default to true): `ENABLE_WORKSPACE_WEBHOOK` and  `ENABLE_WORKSPACE_TEMPLATE_WEBHOOK` (ENABLE_*, default to true is based on standard k8s env vars), fixes #201 

#### Note on compliance checking: 
Initial implementation included automatic compliance checking whenever templates changed, but this was removed due to scalability concerns (see [commit afcec6b](https://github.com/jupyter-infra/jupyter-k8s/pull/129/commits/afcec6bdbcfe6dbd7b004124702ddfdb625cb4f3)). Without proper pagination and rate limiting, checking 1000+ workspaces could slow down or hang the controller. This will be addressed in a future PR.

#### Workspace validation against template

Workspaces are validated against their referenced template:
1. On creation: Full validation of all fields via `ValidateCreate` webhook
2. On update: Partial validation (only changed fields) via `ValidateUpdate` webhook
3. On template change: Existing workspaces are NOT automatically revalidated (lazy/grandfather pattern)

This means:
- New workspaces must comply with current template
- Updated workspaces must comply with current template for updated values
- Existing unchanged workspaces may not comply if template becomes stricter (by design)

### References

* Fixes #19

### Breaking Changes

BREAKING: `spec.templateRef` field format changed from `string` to `struct`

Before:
```yaml
spec:
  templateRef: "notebook-template"
```

After:
```yaml
templateRef:
  # name is required
  name: "notebook-template"
  # namespace is optional
  namespace: NAMESPACENAME
```

closes #91 
closes #87

### DIFF
--- a/internal/controller/workspacetemplate_controller.go
+++ b/internal/controller/workspacetemplate_controller.go
@@ -173,7 +173,8 @@ func (r *WorkspaceTemplateReconciler) handleDeletion(ctx context.Context, templa
 	if len(workspaces) > 0 {
 		logger.Info("Template is in use, blocking deletion",
 			"templateName", template.Name,
-			"exampleWorkspace", workspace.GetWorkspaceKey(&workspaces[0]))
+			"exampleWorkspace", workspaces[0].Name,
+			"exampleWorkspaceNamespace", workspaces[0].Namespace)
 	} else {
 		logger.Info("No workspaces using template",
 			"templateName", template.Name)
@@ -248,7 +249,8 @@ func (r *WorkspaceTemplateReconciler) findTemplatesForWorkspace(ctx context.Cont
 	templateNamespace := ws.Labels[workspace.LabelWorkspaceTemplateNamespace]
 	if templateName == "" || templateNamespace == "" {
 		logger.V(1).Info("Workspace has incomplete template labels, skipping template reconciliation",
-			"workspace", workspace.GetWorkspaceKey(ws),
+			"workspace", ws.Name,
+			"workspaceNamespace", ws.Namespace,
 			"templateName", templateName,
 			"templateNamespace", templateNamespace,
 			"deletionTimestamp", ws.DeletionTimestamp)
@@ -256,7 +258,8 @@ func (r *WorkspaceTemplateReconciler) findTemplatesForWorkspace(ctx context.Cont
 	}
 
 	logger.Info("Workspace changed, enqueueing template reconciliation",
-		"workspace", workspace.GetWorkspaceKey(ws),
+		"workspace", ws.Name,
+		"workspaceNamespace", ws.Namespace,
 		"template", templateName,
 		"templateNamespace", templateNamespace,
 		"deletionTimestamp", ws.DeletionTimestamp,
--- a/internal/workspace/queries.go
+++ b/internal/workspace/queries.go
@@ -42,12 +42,6 @@ func GetTemplateRefNamespace(ws *workspacev1alpha1.Workspace) string {
 	return ws.Spec.TemplateRef.Namespace
 }
 
-// GetWorkspaceKey returns the namespace/name key for a workspace
-// This follows Kubernetes convention for logging object references (similar to klog.KObj)
-func GetWorkspaceKey(ws *workspacev1alpha1.Workspace) string {
-	return fmt.Sprintf("%s/%s", ws.Namespace, ws.Name)
-}
-
 // ListActiveWorkspacesByTemplate returns all active (non-deleted) workspaces using the specified template.
 // Reads from controller-runtime's informer cache (not direct API calls), providing efficient lookup
 // with eventual consistency guarantees. Filters out workspaces being deleted (DeletionTimestamp set).
@@ -96,7 +90,8 @@ func ListActiveWorkspacesByTemplate(ctx context.Context, k8sClient client.Client
 		// This is somewhat redundant given CEL immutability validation but adds zero cost and adds a layer of verification.
 		if ws.Spec.TemplateRef == nil {
 			logger.Info("Workspace has template label but nil templateRef - data integrity issue",
-				"workspace", GetWorkspaceKey(&ws),
+				"workspace", ws.Name,
+				"workspaceNamespace", ws.Namespace,
 				"label", templateName)
 			continue
 		}
@@ -104,7 +99,8 @@ func ListActiveWorkspacesByTemplate(ctx context.Context, k8sClient client.Client
 		if ws.Spec.TemplateRef.Name != templateName {
 			// This should never happen - log if it occurs
 			logger.Info("Workspace has template label but different templateRef name",
-				"workspace", GetWorkspaceKey(&ws),
+				"workspace", ws.Name,
+				"workspaceNamespace", ws.Namespace,
 				"label", templateName,
 				"spec", ws.Spec.TemplateRef.Name)
 			continue
@@ -115,7 +111,8 @@ func ListActiveWorkspacesByTemplate(ctx context.Context, k8sClient client.Client
 			actualNamespace := GetTemplateRefNamespace(&ws)
 			if actualNamespace != templateNamespace {
 				logger.V(1).Info("Workspace has template label but different namespace",
-					"workspace", GetWorkspaceKey(&ws),
+					"workspace", ws.Name,
+					"workspaceNamespace", ws.Namespace,
 					"labelNamespace", templateNamespace,
 					"specNamespace", actualNamespace)
 				continue


internal/controller/workspace_controller.go
```diff
@@ -112,14 +112,14 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// Ensure template label is set if workspace uses a template
-	if workspace.Spec.TemplateRef != nil && *workspace.Spec.TemplateRef != "" {
+	if workspace.Spec.TemplateRef != nil && workspace.Spec.TemplateRef.Name != "" {
```
JGuinegagne: if the `TemplateRef` was removed by the update (i.e. w/ template > no template), this block wouldn't trigger whereas the label ought to be removed, would it?

My 2c: this doesn't belong to the controller, move to the webhook please. Add the label at the same time as we add the `created-by` / `last-updated-by` annotations in the [defaulting webhook](https://github.com/jupyter-infra/jupyter-k8s/blob/296de4170ccbb7fedf8a5974b06fd86e5efc223b/internal/webhook/v1alpha1/workspace_webhook.go#L147)

internal/controller/workspacetemplate_controller.go
```diff
@@ -72,14 +72,21 @@ func (r *WorkspaceTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	return result, nil
 }
 
 // manageFinalizer implements lazy finalizer management for WorkspaceTemplates
 // Following Kubernetes best practices:
 // - Finalizers are only added when workspaces start using the template
 // - Finalizers are removed when all workspaces stop using the template
 // - This minimizes overhead and complexity compared to adding finalizers to all templates
+//
+//nolint:unparam // ctrl.Result signature maintained for consistency with controller-runtime patterns
```
JGuinegagne: I had missed this in previous PRs, this logic won't scale:
1/ it needs to paginate
2/ it performs too many operations in a single reconciliation loop

The workspace reconciliation loop (or better: webhook) should add finalizers to the template, not the other way around.

api/v1alpha1/workspace_types.go
```diff
@@ -73,6 +73,16 @@ type AccessStrategyRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+// TemplateRef defines a reference to a WorkspaceTemplate
+type TemplateRef struct {
+	// Name of the WorkspaceTemplate
+	Name string `json:"name"`
+
+	// Namespace where the WorkspaceTemplate is located (optional for cluster-scoped)
```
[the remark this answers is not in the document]
joshuatowner: I am thinking this means "default namespace". since at launch we typically consider templates to be global scoped, and cross-namespace templates to be common.

internal/controller/state_machine.go
```diff
@@ -191,11 +191,11 @@ func (sm *StateMachine) reconcileDesiredRunningStatus(
 	logger := logf.FromContext(ctx)
 	logger.Info("Attempting to bring Workspace status to 'Running'")
 
-	// Validate template BEFORE creating any resources
-	resolvedTemplate, shouldContinue, err := sm.handleTemplateValidation(ctx, workspace, snapshotStatus)
-	if !shouldContinue {
-		logger.Info("Template validation failed, stopping reconciliation", "error", err)
-		return ctrl.Result{}, nil
+	// Resolve template for defaults (validation already done by webhook)
+	resolvedTemplate, err := sm.resolveTemplate(ctx, workspace, snapshotStatus)
```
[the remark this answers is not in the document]
joshuatowner: I think all this stuff will need to be removed from controller

internal/webhook/v1alpha1/template_validator.go
```diff
@@ -90,14 +90,48 @@ func (tv *TemplateValidator) ValidateCreateWorkspace(ctx context.Context, worksp
...
+	// Case 1: TemplateRef deleted (template → standalone)
+	// Removing constraints is always safe - no validation needed
+	if templateRefDeleted {
+		workspacelog.Info("TemplateRef deleted, allowing transition to standalone workspace", "workspace", newWorkspace.Name)
 		return nil
 	}
 
+	// Case 2: TemplateRef changed (template A → template B)
+	// Must validate entire spec against NEW template
+	if templateRefChanged {
```
earaghbidikashani: in this case, shouldn't we skip validation if only desired status was set to stopped since we always want to allow customers to stop their space?

api/v1alpha1/workspace_types.go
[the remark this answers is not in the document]
andrii-i: When omitted this defaults to the workspace's namespace at runtime in controller, same as `AccessStrategyRef`. I agree with Jonathan that they and all defaulting strategies and all conventions overall should all be consistent and across the codebase and follow the same pattern. 

Outside of consistency this seemed fine solution to me as it is consistent with Kubernetes Gateway API conventions for cross-namespace references. If we would want to change the defaulting behavior we should change it for both template and AccessStrategyRef.

internal/workspace/utils.go
```diff
@@ -5,17 +5,21 @@ import (
...
 	"fmt"
 
-	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/controller"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Label keys - defined here to avoid import cycles with controller package
+const (
+	LabelWorkspaceName = "workspace.jupyter.org/workspaceName"
```
joshuatowner: i think i added constant file in my CR may need rebase

internal/workspace/queries.go
```diff
@@ -0,0 +1,144 @@
...
+
+	listOptions := []client.ListOption{
+		client.MatchingLabels(labels),
+	}
+
+	// Add pagination options if specified
+	if limit > 0 {
+		listOptions = append(listOptions, client.Limit(limit))
+	}
+	if continueToken != "" {
```
earaghbidikashani: for my understanding, who grants this token?

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -33,10 +36,48 @@ import (
 	webhookconst "github.com/jupyter-ai-contrib/jupyter-k8s/internal/webhook"
 )
 
-// nolint:unused
+const templateFinalizerName = "workspace.jupyter.org/template-protection"
```
earaghbidikashani: non-blocking: use webhook constants file

internal/webhook/v1alpha1/template_webhook.go
```diff
@@ -0,0 +1,267 @@
...
+		templatelog.Info("Template constraints changed, controller will mark workspaces for compliance check", "template", newTemplate.GetName())
+		// Return a warning to inform the user that workspaces will be validated
+		return admission.Warnings{"Template constraints changed. Affected workspaces will be marked for compliance validation by the controller."}, nil
+	}
+
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type WorkspaceTemplate.
+func (v *WorkspaceTemplateCustomValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
earaghbidikashani: is this function necessary?

internal/webhook/v1alpha1/template_validator.go
```diff
@@ -83,52 +83,76 @@ func (tv *TemplateValidator) ValidateCreateWorkspace(ctx context.Context, worksp
...
+	// However, if other spec fields also changed, those changes must be validated
+	if newWorkspace.Spec.DesiredStatus == controller.PhaseStopped &&
+		oldWorkspace.Spec.DesiredStatus != controller.PhaseStopped &&
+		onlyDesiredStatusChanged(&oldWorkspace.Spec, &newWorkspace.Spec) {
+		workspacelog.Info("Allowing workspace stop without template validation (status-only change)", "workspace", newWorkspace.Name)
+		return nil
 	}
 
-	return nil
+	// Spec changed with same template - validate ENTIRE spec against template
```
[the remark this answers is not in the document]
earaghbidikashani: @andrii-i could you add more details regarding template complience and when a workspace will be marked for validation that this PR is addressing, this is specially helpful since this is a bigger PR

internal/webhook/v1alpha1/spec_comparator.go
```diff
@@ -0,0 +1,297 @@
...
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+
+	workspacev1alpha1 "github.com/jupyter-ai-contrib/jupyter-k8s/api/v1alpha1"
+)
+
+// specChanged detects if any spec field changed between old and new workspace
+// This is used to determine if full template validation is needed on update
+func specChanged(oldSpec, newSpec *workspacev1alpha1.WorkspaceSpec) bool {
```
earaghbidikashani: why can't we use a deep equal method to compare 2 objects? why are we comparing every field?

internal/webhook/v1alpha1/resource_validator.go
```diff
@@ -112,6 +112,31 @@ func validateResourceBounds(resources corev1.ResourceRequirements, template *wor
 		}
 	}
 
+	// Validate GPU bounds
+	if bounds.GPU != nil && resources.Requests != nil {
+		gpuResourceName := corev1.ResourceName("nvidia.com/gpu")
```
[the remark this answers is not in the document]
earaghbidikashani: #99 is the issue I believe

internal/controller/constants.go
```diff
@@ -21,11 +21,12 @@ const (
...
-	LabelWorkspaceName           = "workspace.jupyter.org/workspaceName"
-	LabelWorkspaceNamespace      = "workspace.jupyter.org/workspaceNamespace"
-	LabelAccessStrategyName      = "workspace.jupyter.org/accessStrategyName"
-	LabelAccessStrategyNamespace = "workspace.jupyter.org/accessStrategyNamespace"
-	LabelWorkspaceTemplate       = "workspace.jupyter.org/template"
+	LabelWorkspaceName              = "workspace.jupyter.org/workspaceName"
+	LabelWorkspaceNamespace         = "workspace.jupyter.org/workspaceNamespace"
+	LabelAccessStrategyName         = "workspace.jupyter.org/accessStrategyName"
+	LabelAccessStrategyNamespace    = "workspace.jupyter.org/accessStrategyNamespace"
+	LabelWorkspaceTemplate          = "workspace.jupyter.org/template"
```
JGuinegagne: For consistency, could you rename `LabelWorkspaceTemplateName` and `/template-name` please?

While we are at it, shall we change all label names casing (e.g.` /workspaceName` > `workspace-name`).
Fine as well if we think it's better in a future PR

internal/controller/pvc_builder.go
```diff
@@ -178,10 +168,11 @@ func (pb *PVCBuilder) UpdatePVCSpec(ctx context.Context, existingPVC *corev1.Per
 		return fmt.Errorf("cannot update PVC to nil spec")
 	}
 
-	// Update only the fields we control (preserve Kubernetes-managed fields)
+	// Update only the MUTABLE fields (preserve immutable and Kubernetes-managed fields)
+	// Note: StorageClassName is immutable after creation and cannot be updated
 	existingPVC.Spec.AccessModes = desiredPVC.Spec.AccessModes
 	existingPVC.Spec.Resources = desiredPVC.Spec.Resources
-	existingPVC.Spec.StorageClassName = desiredPVC.Spec.StorageClassName
+	// DO NOT update StorageClassName - it's immutable after PVC creation
```
JGuinegagne: just to confirm: this change comes from another branch? otherwise curious why you are making this change (which seems sound btw)

internal/controller/workspacetemplate_controller.go
```diff
@@ -141,7 +165,8 @@ func (r *WorkspaceTemplateReconciler) handleDeletion(ctx context.Context, templa
 	}
 
 	// Check if any workspaces are using this template
-	workspaces, err := r.templateResolver.ListWorkspacesUsingTemplate(ctx, template.Name)
+	// Use empty namespace to match all workspaces (backwards compatible)
+	workspaces, _, err := workspace.ListByTemplate(ctx, r.Client, template.Name, "", "", 0)
```
JGuinegagne: any way to specify a `limit=1`, ie return at most 1 results? That should be more efficient given the user case.

internal/webhook/v1alpha1/template_validator.go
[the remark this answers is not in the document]
JGuinegagne: +1, could you confirm Andrii you're handling this case?

internal/webhook/v1alpha1/template_webhook.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	templatelog.Info("Validation for WorkspaceTemplate upon deletion", "name", template.GetName())
+
+	// Deletion is handled by finalizers in the controller
+	return nil, nil
+}
+
+// constraintsChanged checks if any constraint fields changed between old and new templates
+// Constraint fields are those that affect workspace validation (resource bounds, allowed images, etc.)
+func constraintsChanged(oldTemplate, newTemplate *workspacev1alpha1.WorkspaceTemplate) bool {
```
JGuinegagne: non-blocking but before GA: move to separate files please

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -33,10 +36,48 @@ import (
...
+	if err := k8sClient.Update(ctx, template); err != nil {
+		workspacelog.Error(err, "Failed to add finalizer to template", "template", templateRef)
+		return fmt.Errorf("failed to add finalizer to template %s: %w", templateRef, err)
+	}
+
+	workspacelog.Info("Added finalizer to template", "template", templateRef)
+	return nil
+}
+
+func sanitizeUsername(username string) string {
```
JGuinegagne: question: why do you need this method?

internal/webhook/v1alpha1/workspace_webhook_test.go
```diff
@@ -783,6 +860,263 @@ var _ = Describe("Workspace Webhook", func() {
 			Expect(isAdminUser(groups)).To(BeFalse())
 		})
 	})
+
+	Context("Metadata-Only Updates (GAP-7)", func() {
```
JGuinegagne: ah so this is handled, I'll reread the webhook code :)

internal/workspace/queries.go
```diff
@@ -0,0 +1,144 @@
...
+
+	if err := k8sClient.List(ctx, workspaceList, listOptions...); err != nil {
+		return nil, "", fmt.Errorf("failed to list workspaces by template label: %w", err)
+	}
+
+	// Filter out workspaces being deleted and verify template reference
+	// This follows Kubernetes controller best practice: resources with deletionTimestamp
+	// are considered "gone" for dependency checking purposes
+	activeWorkspaces := []workspacev1alpha1.Workspace{}
+	for _, ws := range workspaceList.Items {
```
JGuinegagne: ah I see, we can't just pass `Limit=1` then...Anyway we could filter on `DeletionTimestamp==0` ?

If no, let's discuss whether we need pagination here.

Also, could you rename `ListNonDeletedWorkspaceByTemplate()` or smth that conveys that logic please?

internal/workspace/queries.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// HasWorkspacesWithTemplate checks if any active workspace uses the specified template
+// More efficient than ListByTemplate when only existence check is needed
+// Returns true if at least one workspace uses the template
+func HasWorkspacesWithTemplate(ctx context.Context, k8sClient client.Client, templateName string) (bool, error) {
+	workspaceList := &workspacev1alpha1.WorkspaceList{}
+	if err := k8sClient.List(ctx, workspaceList,
+		client.MatchingLabels{
+			LabelWorkspaceTemplate: templateName,
```
JGuinegagne: check the template namespace here please

internal/webhook/v1alpha1/template_validator.go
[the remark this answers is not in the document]
andrii-i: @joshuatowner Check out the whole funtion, this is the last block that if reached means that `spec` changed, not metadata. Block below ensures that if only metadata changed, we skip validation: 

```
	// Check if any spec field changed
	if !specChanged(&oldWorkspace.Spec, &newWorkspace.Spec) {
		// No spec changes - skip validation (metadata-only update)
		return nil
	}
```

@earaghbidikashani summarizing code comments in this function, also adding the same to the PR description for better visibility:
```
    TemplateRef deleted - no validation
    TemplateRef changed - validation vs template
    TemplateRef added - validation vs template
    No templateRef, not updated - no validation
    TemplateRef unchanged - check other conditions Check if any spec field changed
        No spec changes - skip validation (metadata-only update)
        If ONLY DesiredStatus changed to Stopped, allow without validation
        if other spec fields also changed, those changes must be validated
```

internal/webhook/v1alpha1/workspace_webhook.go
JGuinegagne: +1

JGuinegagne:
Also, thank you for putting so much thoughts and efforts on building a robust finalization logic.

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -33,10 +36,48 @@ import (
...
+	// Use Go's JSON marshaling to properly escape the string
+	escaped, _ := json.Marshal(username)
```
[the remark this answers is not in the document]
andrii-i: Yep, thank you.

config/rbac/role.yaml
```diff
@@ -79,6 +79,17 @@ rules:
...
+  - workspace.jupyter.org
+  resources:
+  - workspaces
+  - workspacetemplates
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
andrii-i: Yes, workspacetemplate_controller.go:49,52 lead to duplication vs wildcard at workspace_controller.go:81. Removing workspacetemplate_controller.go:49,52 (assuming wildcard / create permission from  workspace_controller.go:81 is correct).

andrii-i:
@JGuinegagne 

>   1.  could we simplify the template webhook by a/ listing at most 1 ws, b/ accept that such a ws might have a TerminationTimestmap, eventually it should be removed and the call would succeed.
>   2.  add namespace handling for templateRef

Done with defaulting to workspace namespace if not set approach.

> 3. if a resource has a finalizer set, do we need to do anything in the delete webhook or will the API server will reject the request at RBAC stage

No delete webhook needed. Finalizers block deletion at API server level before webhooks run. Controller's `handleDeletion` checks workspace usage and removes finalizer when safe.

internal/controller/constants.go
```diff
@@ -22,12 +22,17 @@ const (
...
+	// LabelWorkspaceName is the label key for workspace name
+	LabelWorkspaceName = "workspace.jupyter.org/workspace-name"
+	// LabelWorkspaceNamespace is the label key for workspace namespace
+	LabelWorkspaceNamespace = "workspace.jupyter.org/workspace-namespace"
+	// LabelAccessStrategyName is the label key for access strategy name
+	LabelAccessStrategyName = "workspace.jupyter.org/access-strategy-name"
+	// LabelAccessStrategyNamespace is the label key for access strategy namespace
+	LabelAccessStrategyNamespace = "workspace.jupyter.org/access-strategy-namespace"
+	// LabelWorkspaceTemplate is the label key for workspace template name
+	LabelWorkspaceTemplate = "workspace.jupyter.org/template"
```
JGuinegagne: follow up PR please: `/template-name` for consistency

internal/controller/workspacetemplate_controller.go
```diff
@@ -250,23 +245,28 @@ func (r *WorkspaceTemplateReconciler) findTemplatesForWorkspace(ctx context.Cont
...
+			"workspace", workspace.GetWorkspaceKey(ws),
+			"templateName", templateName,
+			"templateNamespace", templateNamespace,
 			"deletionTimestamp", ws.DeletionTimestamp)
 		return nil
 	}
 
 	logger.Info("Workspace changed, enqueueing template reconciliation",
-		"workspace", fmt.Sprintf("%s/%s", ws.Namespace, ws.Name),
+		"workspace", workspace.GetWorkspaceKey(ws),
```
JGuinegagne: what is this method? Why can't we just do `workspace.metadata.name` ?

internal/controller/workspace_controller.go
```diff
@@ -130,23 +125,62 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if workspace.Labels[workspaceutil.LabelWorkspaceTemplateNamespace] != templateNamespace {
+			workspace.Labels[workspaceutil.LabelWorkspaceTemplateNamespace] = templateNamespace
+			needsUpdate = true
+		}
+
+		if needsUpdate {
+			logger.Info("Adding/updating template labels",
+				"template", templateName,
+				"templateNamespace", templateNamespace)
 			if err := r.Update(ctx, workspace); err != nil {
```
earaghbidikashani: is this update call async?

internal/controller/workspacetemplate_controller_test.go
```diff
@@ -274,6 +279,42 @@ var _ = Describe("WorkspaceTemplate Controller", func() {
...
+				},
+			}
+
+			requests := reconciler.findTemplatesForWorkspace(ctx, workspace)
+			Expect(requests).To(HaveLen(1))
+			Expect(requests[0].Name).To(Equal("shared-template"))
+			Expect(requests[0].Namespace).To(Equal("templates-ns"),
+				"should reconcile template in different namespace")
+		})
+
```
JGuinegagne: fast-follow: could you test all cases of template-related label additions:
- new templateRef, templateRef.ns not empty --> add labels
- new templateRef, templateRef.ns empty --> add labels w/ defaults to ws.ns
- remove templateRef --> remove existing labels
- remove templateRef, not existing labels --> no-op

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -33,10 +36,48 @@ import (
...
 var workspacelog = logf.Log.WithName("workspace-resource")
 
+// ensureTemplateFinalizer ensures the template has a finalizer to prevent deletion while in use
+func ensureTemplateFinalizer(ctx context.Context, k8sClient client.Client, templateRef string) error {
+	if templateRef == "" {
+		return nil
+	}
+
+	// Fetch the template
+	template := &workspacev1alpha1.WorkspaceTemplate{}
```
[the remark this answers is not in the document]
JGuinegagne: fast-follow: document somewhere in code or in README please as this is complex.

earaghbidikashani (dismissed):

JGuinegagne (approved):
LGTM w/ 1 sanity-check and a few fast-follow items.

PR #144, opened 2025-11-01
feat: authentication for extension api server

aws-jasakshi:
Implement standard Kubernetes authentication for the extension API server to secure workspace connection operations.

**Changes:**
- Add AuthConfig with standard k8s.io/apiserver authentication
- Read authentication configuration from kube-system ConfigMap  
- Implement authentication middleware for all API routes

**Security:**
- Uses EKS-compatible header-based authentication
- Validates X-Remote-User and X-Remote-Group headers

**Testing:**
- Unit tests
- Manual testing - happy path

### DIFF
--- a/internal/extensionapi/server_test.go
+++ b/internal/extensionapi/server_test.go
@@ -1036,3 +1036,79 @@ var _ = Describe("ServerWithManager", func() {
 		})
 	})
 })
+
+var _ = Describe("Authentication Middleware", func() {
+	var (
+		server     *ExtensionServer
+		authConfig *AuthConfig
+		logger     logr.Logger
+	)
+
+	BeforeEach(func() {
+		logger = logr.Discard()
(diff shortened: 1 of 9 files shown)


internal/extensionapi/server_test.go
```diff
@@ -1036,3 +1036,79 @@ var _ = Describe("ServerWithManager", func() {
...
+
+			middleware := server.authMiddleware(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+
+			middleware(w, req)
+			Expect(w.Code).To(Equal(http.StatusOK))
+		})
+
+		It("should bypass auth when no auth config", func() {
```
JGuinegagne: isn't this a dangerous behavior? As in, do we want to allow requests to go through if the auth middleware somehow wasn't set?

PR #145, opened 2025-11-01
feat: KMS key and SSH document lazy initialization

aws-jasakshi:
- Add CreateJWTKMSKey method to KMSClient
- Add CreateSSHDocument method to SSMClient 
- Add lazy initialization using sync.Once pattern in connection route
- Update VSCode scheme from /sagemaker to /workspace

### DIFF
--- a/internal/aws/constants.go
+++ b/internal/aws/constants.go
@@ -8,6 +8,9 @@ const (
 	// EKSClusterARNEnv is the environment variable key for EKS cluster ARN
 	EKSClusterARNEnv = "CLUSTER_ID"
 
+	// SSHDocumentContentEnv is the environment variable for SSH document content
+	SSHDocumentContentEnv = "SSH_DOCUMENT_CONTENT"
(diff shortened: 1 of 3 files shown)


internal/aws/initializer.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var (
+	initOnce  sync.Once
+	initError error
+)
+
+// EnsureResourcesInitialized ensures KMS key and SSH document are created (only once)
+func EnsureResourcesInitialized(ctx context.Context) error {
+	initOnce.Do(func() {
```
joshuatowner: issues with this:
- two controllers may create same KMS key. lets catch the error if the alias already exists, use the key from the alias, and delete the created key. 
- if there's an error, this will never recover. we should probably move to just a simple lock here.

JGuinegagne (changes requested):
Reuse aws clients please

PR #164, opened 2025-11-04
refactor: change the aws template to use init-container

edwardps:
**Description**
1. Add the init-container to copy the entrypoint scripts for workspace use
case. The scripts are copied from sidecar image to the shared volume and
later used by the app container launch.

2. Fixed sidecar crash due to limit configuration

**Testing Done**
This is a template change and manually tested by creating spaces.

**Backwards Compatibility Criteria (if any)**
N/A

### DIFF
--- a/internal/extensionapi/connection_route.go
+++ b/internal/extensionapi/connection_route.go
@@ -10,6 +10,7 @@ import (
 
 	connectionv1alpha1 "github.com/jupyter-ai-contrib/jupyter-k8s/api/connection/v1alpha1"
 	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/aws"
+	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/jwt"
 	"github.com/jupyter-ai-contrib/jupyter-k8s/internal/workspace"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


guided-charts/aws-hyperpod/templates/aws-ssm-remote-access-strategy.yaml
```diff
@@ -23,32 +23,41 @@ spec:
           command: ["/bin/sh"]
           args: 
             - "-c"
-            - "cp /usr/local/bin/remote-access-server /ssm-remote-access/ || echo \"Failed to copy: $?\"; sleep infinity"
+            - "cp /usr/local/bin/remote-access-server /opt/amazon/sagemaker/workspace/ || echo \"Failed to copy: $?\"; sleep infinity"
           resources:
             requests:
```
joshuatowner: lets set these to resources = limits. k8s will schedule a pod based on requests only, meaning a node may not have enough resources to support the limits.

edwardps: With requests < limits, it’s a burstable QoS pod which is usually better fit for mixed workloads. Just in case side car needs more resource temporarily. 

> meaning a node may not have enough resources to support the limits

This is as expected if the whole cluster is overloaded as rare case.

aws-prayags (approved):

PR #169, opened 2025-11-04
feat: support Container_Registry in ECR repository

aakashmandavilli96:
Make changes for the ECR repository's to support Container_Registry's
- Changed Traefik
- Auth-Middleware
- Operator
- SsmSideCar 

to follow the same pattern for the launch

### DIFF
--- a/dist/chart/templates/manager/manager.yaml
+++ b/dist/chart/templates/manager/manager.yaml
@@ -44,7 +44,7 @@ spec:
             {{- end}}
           command:
             - /manager
-          image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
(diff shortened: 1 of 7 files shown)


dist/chart/templates/manager/manager.yaml
```diff
@@ -44,7 +44,7 @@ spec:
             {{- end}}
           command:
             - /manager
-          image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
+          image: {{ .Values.controllerManager.container.image.containerRegistry }}/{{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
```
JGuinegagne: This doesn't look right, the `dist/chart` dir is generated.
You need to make changes in `config`.

Also, this looks like a breaking change, why can't the `repository` contain the full path w/ `/` ?

JGuinegagne (changes requested):
This changes doesn't look right.

Make sure to run `make helm-generate` before submitting your changes.

PR #182, opened 2025-11-05
fix: attempting to rename PVC volume name constantly

joshuatowner:


internal/controller/pvc_builder.go
```diff
@@ -144,8 +144,26 @@ func (pb *PVCBuilder) NeedsUpdate(ctx context.Context, existingPVC *corev1.Persi
 		return false, nil
 	}
 
-	// Compare PVC specs using semantic equality
-	return !equality.Semantic.DeepEqual(existingPVC.Spec, desiredPVC.Spec), nil
+	// Compare only the fields we control (ignore Kubernetes-managed fields)
```
earaghbidikashani: why are we adding this? what do you mean by `fields we control`? 
We control storage class, which essentially creates PVC spec and there are many fields that need to be watched, you can find all the fields here: https://pub.dev/documentation/kubernetes/latest/io.k8s.api.core.v1/PersistentVolumeClaimSpec-class.html


What is exactly changing in the pvc spec that is causing this function to return true?

joshuatowner: the PV name is changing, this is set by k8s

earaghbidikashani: pv is not in pvc spec

earaghbidikashani: I see, approving for now but we should investigate if we need to compare storage class instead of pvc

earaghbidikashani (approved):

PR #185, opened 2025-11-05
feat: Adding standard env var

### LINKED ISSUE #160: Support environment variables needed for remote access strategy

earaghbidikashani:
added standard env vars to pod including:
- workspace name
- workspace namespace
- access type
- app type

closes #160


joshuatowner:
sorry i missed this. we need to allow these to be filled in by access strategy merge env. not always done. could you take a look and see how hard this is?

PR #191, opened 2025-11-06
feat: KMS key and SSH document lazy initialization

henrywa2:
Based off https://github.com/jupyter-infra/jupyter-k8s/pull/145

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -285,7 +285,7 @@ func main() {
 		// Create config with a different port to avoid conflict with metrics
 		config := extensionapi.NewConfig(
 			extensionapi.WithServerPort(7443),
-			extensionapi.WithClusterId(os.Getenv("CLUSTER_ARN")),
+			extensionapi.WithClusterId(os.Getenv("CLUSTER_ID")),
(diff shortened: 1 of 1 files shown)


earaghbidikashani (approved):

cmd/main.go
```diff
@@ -285,7 +285,7 @@ func main() {
 		// Create config with a different port to avoid conflict with metrics
 		config := extensionapi.NewConfig(
 			extensionapi.WithServerPort(7443),
-			extensionapi.WithClusterId(os.Getenv("CLUSTER_ID")),
+			extensionapi.WithClusterId(os.Getenv("CLUSTER_ARN")),
```
JGuinegagne: sanity-check: this doesn't look right, we shouldn't use `ARN` in such generic module as `main.go`

JGuinegagne (changes requested):
Verify env var name please

cmd/main.go
henrywa2: I'll change this back to CLUSTER_ID

JGuinegagne (approved):

PR #198, opened 2025-11-07
feat: temporarily add multiple user keys to JWT token

joshuatowner:
Unfortunately a change in one of our clients uses User instead of user. While we fix this change client-side, we want to unblock with a controller change.


joshuatowner:
closing in favor of #199

PR #217, opened 2025-11-11
fix: handle not found errors separately from others

joshuatowner:
- handle notfound specifically, and not all errors
- add tests for template resolver

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -636,6 +636,7 @@ apply-sample-routing:
 		set -e; \
 		. ./.env; \
 		export DOMAIN=$$DOMAIN; \
+		kubectl apply -f config/samples_routing/workspace_access_strategy.yaml --dry-run=client -o yaml | envsubst | kubectl apply -f -; \
 		kubectl apply -k config/samples_routing --dry-run=client -o yaml | envsubst | kubectl apply -f -; \
 	)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -275,6 +275,11 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "WorkspaceTemplate")
 		os.Exit(1)
 	}
+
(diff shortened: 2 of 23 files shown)


internal/workspace/template_resolver.go
```diff
@@ -44,7 +44,7 @@ func (tr *TemplateResolver) ResolveTemplate(ctx context.Context, templateRef *wo
 	err := tr.client.Get(ctx, templateKey, template)
 
 	// If not found and we have a default namespace, try there
-	if err != nil && tr.defaultTemplateNamespace != "" && templateNamespace != tr.defaultTemplateNamespace {
+	if client.IgnoreNotFound(err) == nil && tr.defaultTemplateNamespace != "" && templateNamespace != tr.defaultTemplateNamespace {
```
JGuinegagne: sanity-check: shouldn't you keep an `err != nil &&` first? as in, what would `client.IgnoreNotFound(err)` return if `err == nil` ?

For reference, we're using `errors.isNotFound()` elsewhere in the codebase ([example](https://github.com/jupyter-infra/jupyter-k8s/blob/8de30deb51b2c11564a3bd8d4574ea232a1c1e75/internal/controller/resource_manager.go#L191)), however no objection if this functionally works.

internal/workspace/template_resolver.go
```diff
@@ -44,7 +44,7 @@ func (tr *TemplateResolver) ResolveTemplate(ctx context.Context, templateRef *wo
...
 		templateKey = client.ObjectKey{Name: templateRef.Name, Namespace: tr.defaultTemplateNamespace}
 		if fallbackErr := tr.client.Get(ctx, templateKey, template); fallbackErr == nil {
```
JGuinegagne: shouldn't we log or otherwise keep trace of the `fallbackErr` ?

aws-prayags (approved):

Issue #63, opened 2025-10-23
[aws-hyperpod] Handling failure cases / cleanup for aws-ssm-remote-access strategy


aws-prayags:
1/ We need to handle container crashes ( when pod itself doesn't crash )
2/ Add defensive cleanup in init flow to prevent resource leak

### Not fully covered by current issue - to be addressed in https://github.com/jupyter-infra/jupyter-k8s/issues/274

4/ Should we reuse workspace reconciler for pod related actions ? https://github.com/jupyter-infra/jupyter-k8s/pull/58#discussion_r2454337294

5/ If AWS API call fails in cleanup flow, how do we handle it ? how to prevent leaked resources?

PR #225, opened 2025-11-12
fix: hyperpod remote access - handle container restarts

aws-prayags:
Fixes - https://github.com/jupyter-infra/jupyter-k8s/issues/63

Tested functionality on EKS cluster

### DIFF
--- a/internal/workspace/template_resolver.go
+++ b/internal/workspace/template_resolver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	workspacev1alpha1 "github.com/jupyter-ai-contrib/jupyter-k8s/api/v1alpha1"
@@ -44,10 +45,12 @@ func (tr *TemplateResolver) ResolveTemplate(ctx context.Context, templateRef *wo
 	err := tr.client.Get(ctx, templateKey, template)
 
 	// If not found and we have a default namespace, try there
-	if err != nil && tr.defaultTemplateNamespace != "" && templateNamespace != tr.defaultTemplateNamespace {
+	if apierrors.IsNotFound(err) && tr.defaultTemplateNamespace != "" && templateNamespace != tr.defaultTemplateNamespace {
 		templateKey = client.ObjectKey{Name: templateRef.Name, Namespace: tr.defaultTemplateNamespace}
 		if fallbackErr := tr.client.Get(ctx, templateKey, template); fallbackErr == nil {
 			return template, nil
+		} else {
+			return nil, fmt.Errorf("failed to get template %s from namespace %s or fallback namespace %s: %w", templateRef.Name, templateNamespace, tr.defaultTemplateNamespace, fallbackErr)
 		}
 	}
 
--- a/internal/workspace/template_resolver_test.go
+++ b/internal/workspace/template_resolver_test.go
@@ -1,0 +1,301 @@
+package workspace
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


internal/aws/ssm_remote_access_strategy.go
```diff
@@ -146,64 +245,144 @@ func (s *SSMRemoteAccessStrategy) CleanupSSMManagedNodes(ctx context.Context, po
...
+	if !s.isContainerRunning(pod, WorkspaceContainerName) {
+		logger.V(1).Info("Workspace container not running yet, waiting")
+		return nil
+	}
+
+	// TODO: Make this script idempotent
+	// For now, only called on container restart so should be fine
+	logger.V(1).Info("Starting remote access server in workspace container")
+	serverCmd := []string{"bash", "-c", fmt.Sprintf("sudo %s > /dev/null 2>&1 &", RemoteAccessServerPath)}
+	if _, err := s.podExecUtil.ExecInPod(ctx, pod, WorkspaceContainerName, serverCmd, ""); err != nil {
```
edwardps: Not blocking. 
We can ignore the error when the port is already alive instead of print error.

internal/aws/ssm_remote_access_strategy.go
```diff
@@ -146,64 +245,144 @@ func (s *SSMRemoteAccessStrategy) CleanupSSMManagedNodes(ctx context.Context, po
...
+	}
+	return sidecarCount, workspaceCount
+}
+
+// readRegistrationState reads state from shared volume
+// Returns:
+//   - (*RegistrationState, nil) if state file exists and is valid
+//   - (nil, nil) if state file doesn't exist (first time setup)
+//   - (nil, error) if state file exists but is corrupted
+func (s *SSMRemoteAccessStrategy) readRegistrationState(ctx context.Context, pod *corev1.Pod) (*RegistrationState, error) {
```
edwardps: as we discussed offline, we need to think about better way to address the race condition especially the side car container case. The ssm related status(de-register, resgister) could be tricky when race condition happens. 

Given it's overall improving the UX, this could be p0.5 and not blocking.

internal/aws/ssm_remote_access_strategy.go
```diff
@@ -79,47 +94,131 @@ func NewSSMRemoteAccessStrategy(ssmClient SSMRemoteAccessClientInterface, podExe
...
-func (s *SSMRemoteAccessStrategy) InitSSMAgent(ctx context.Context, pod *corev1.Pod, workspace *workspacev1alpha1.Workspace, accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy) error {
+// SetupContainers initializes SSM agent and remote access server for workspace containers.
+// Handles both first-time setup and recovery from container restarts.
+//
+// Container Restart Detection:
+// - Uses persistent state file (/opt/amazon/sagemaker/workspace/.ssm-registration-state.json)
+// - Compares stored restart counts with current pod status to detect restarts
+// - Selectively re-setups only affected containers (sidecar and/or workspace)
+//
+// Concurrency Protection:
```
joshuatowner: - I believe `Watches` has concurrency protection. why did we decide to add this sleep
- will this sleep cause events to pile up? I think blocking in this controller code seems very dangerous...

joshuatowner (approved):

internal/aws/ssm_remote_access_strategy.go
```diff
@@ -149,64 +248,145 @@ func (s *SSMRemoteAccessStrategy) CleanupSSMManagedNodes(ctx context.Context, po
...
+
+// readRegistrationState reads state from shared volume
+// Returns:
+//   - (*RegistrationState, nil) if state file exists and is valid
+//   - (nil, nil) if state file doesn't exist (first time setup)
+//   - (nil, error) if state file exists but is corrupted
+func (s *SSMRemoteAccessStrategy) readRegistrationState(ctx context.Context, pod *corev1.Pod) (*RegistrationState, error) {
+	logger := logf.FromContext(ctx).WithValues("pod", pod.Name)
+
+	cmd := []string{"cat", SSMRegistrationStateFile}
```
JGuinegagne: This is really brittle.

Medium term: we should define a contract with the image, and hopefully ping an HTTP endpoint or trigger a pre-set `.sh` file that returns the registration state.

PR #229, opened 2025-11-12
chore: Removing KMS create logic

henrywa2:
Closes https://github.com/jupyter-infra/jupyter-k8s/issues/216

Tested initializer logic in EKS

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -39,7 +39,7 @@ webhooks:
       namespace: system
       path: /validate-pods-exec-workspace
       port: 9443
-  failurePolicy: Fail
+  failurePolicy: Ignore
   name: vpods-exec-workspace-v1.kb.io
   rules:
   - apiGroups:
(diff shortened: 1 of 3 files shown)


internal/aws/initializer.go
```diff
@@ -6,44 +6,48 @@ import (
...
-			initError = fmt.Errorf("failed to create SSH document: %w", err)
-			return
-		}
+var globalInitializer *ResourceInitializer
+var initializerOnce sync.Once
+
+// EnsureResourcesInitialized is a global function that uses a singleton initializer
+var EnsureResourcesInitialized = func(ctx context.Context) error {
+	var err error
+	initializerOnce.Do(func() {
```
joshuatowner: add a follow up to rework this logic... imo it is a bit fragile.

if SSM is down temporarily, then we will never retry to create the document. we should instead have a simple lock here and in createSageMakerSpaceSSMDocument catch the resource already exists exceptions. then we can also use same lock for initialization. 

I know it's not new to your CR so not blocking.

Issue #71, opened 2025-10-23
[ssm-remote-access-strategy] track active sessions for each managed instance and enforce an upper limit


aws-prayags:
Context - https://github.com/jupyter-infra/jupyter-k8s/pull/58#discussion_r2455941159

user could create too many sessions to cause throttling or deny of service

PR #231, opened 2025-11-13
fix: limit number of connections to ssm managed instance

aws-prayags:
Fixes https://github.com/jupyter-infra/jupyter-k8s/issues/71

Example output for error
```
Error from server: error when creating "STDIN": failed to start SSM session: instance mi-0b4320ded10283b65 exceeds active sessions limit (11/10)
```

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -39,7 +39,7 @@ webhooks:
       namespace: system
       path: /validate-pods-exec-workspace
       port: 9443
-  failurePolicy: Fail
+  failurePolicy: Ignore
   name: vpods-exec-workspace-v1.kb.io
   rules:
   - apiGroups:
--- a/dist/chart/templates/webhook/webhooks.yaml
+++ b/dist/chart/templates/webhook/webhooks.yaml
@@ -50,7 +50,7 @@ webhooks:
         name: jupyter-k8s-webhook-service
         namespace: {{ .Release.Namespace }}
         path: /validate-pods-exec-workspace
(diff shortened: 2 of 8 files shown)


aws-jasakshi (dismissed):

internal/aws/constants.go
```diff
@@ -37,6 +37,9 @@ const (
 
 	// CustomSSHDocumentName is the name of the SSM document for SSH sessions
 	CustomSSHDocumentName = "SageMaker-SpaceSSHSessionDocument"
+
+	// MaxConcurrentSessionsPerInstance is the maximum number of concurrent SSM sessions allowed per managed instance
+	MaxConcurrentSessionsPerInstance = 10
```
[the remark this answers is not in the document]
joshuatowner: ideally this would also be helm param

internal/aws/ssm_client.go
```diff
@@ -149,6 +155,44 @@ func (c *SSMClient) StartSession(ctx context.Context, instanceID, documentName s
...
+		},
+	}
+
+	result, err := c.client.DescribeSessions(ctx, input)
+	if err != nil {
+		logger.Error(err, "Failed to describe sessions", "instanceId", instanceID)
+		return fmt.Errorf("failed to describe sessions for instance %s: %w", instanceID, err)
+	}
+
+	numActiveSessions := len(result.Sessions)
```
joshuatowner: what about pagination?

aws-prayags: added MaxResults, we don't need to handle pagination now

PR #238, opened 2025-11-13
feat: zero trust in auth route of middleware

### LINKED ISSUE #36: PR20 fast follows

### LINKED ISSUE #227: OSS: implement zero-trust policy in authmiddleware
## Problem
In the OSS flow (`aws-traefik-dex` guided deployment), the `authmiddleware` currently trusts the auth headers as the identity of the caller.

### Threat vectors
1. an attacker directly calls `authmiddleware` with the header
2. an attacker figures out the oauth2 proxy [secret](), and tampers the `_oauth2_proxy` cookie secret with a different identity.

### Existing mitigation
For 1/ we implemented a network policy in #132 but that is partial protection only, an attacker with access to the `jupyter-k8s-router` namespace could setup a pod in that namespace and direct attacks from inside that pod.

2/, is particularly sensitive because right now the oauth2-proxy secret doesn't rotate (will track in a separate issue).
Note: the JWT secret should also rotate (will track in a separate issue).

## Solution
Implement zero-trust paradigm for `authmiddleware`.

### Current auth flow
When the `/auth` route of the `authmiddleware` is called, it reads and trusts the following [headers](https://github.com/jupyter-infra/jupyter-k8s/blob/8de30deb51b2c11564a3bd8d4574ea232a1c1e75/internal/authmiddleware/constants.go#L6) to make authorization decision.

The flow is as follow:
HTTPS-request > traefik > oauth2-proxy > dex > oauth2-proxy (set headers) > traefik > authmiddleware (issue JWT in cookie) > workspace

### Updated flow
oauth2_proxy sets a `_oauth2_proxy` cookie which `oauth2-proxy` might reuse even if it has been tampered with. We need to protect against that by verifying. `oauth2-proxy` will set as `Authorization` header the `dex` token.
Then flow will be:
HTTPS-request > traefik > oauth2-proxy > dex > oauth2-proxy (set headers) > traefik > authmiddleware (verify OIDC token w/ dex) > dex > authmiddleware (issue JWT in cookie) > workspace

### Dex
Add an `authmiddleware` client to `dex` in [dex-config](https://github.com/jupyter-infra/jupyter-k8s/blob/8de30deb51b2c11564a3bd8d4574ea232a1c1e75/guided-charts/aws-traefik-dex/templates/dex/configmap.yaml#L16)

### Oauth2-proxy
Pass the full authorization header from Dex downstream for `authmiddleware` to use. In `oauth2-proxy` deployment:
```yaml
        - --set-authorization-header=true     # Pass ID token in Authorization header
        - --pass-authorization-header=true    # Forward it downstream
```

### Traefik
Update [middlewares](https://github.com/jupyter-infra/jupyter-k8s/blob/main/guided-charts/aws-traefik-dex/templates/traefik/auth-middlewares.yaml) to pass the correct headers.

### Authmiddleware
Read the `Authorization bearer` header in a `verifyDexIDToken` method, with a callback to `dex` service using `github.com/coreos/go-oidc` package

JGuinegagne:
This PR implements the zero-trust paradigm for the authmiddleware `/auth` route. Addresses #227 

Previously, the `authmiddleware` would trust the `X-Auth-Request-Preferred-Username` and `X-Auth-Request-Groups` headers it received from `oauth2-proxy`, and make authorization decisions based on this.

Now, `authmiddleware` requires the `Authorization` header to be set, and expects the token issued from `dex` as its value.

### Implementation
- register `authmiddleware` as client to `dex` (with static client ID and secret)
- pass the `Authorization` header in `oauth2-proxy`
- add `Authorization` header to `authmiddleware` traefik middleware
- pass `dex` configuration to `authmiddleware` via env vars and `config`
- when oauth is enabled, instantiates an `OIDCVerifier` in `authmiddleware`, which on `Start()`
    - queries the OIDC provider public keys (`updateKeySet()`)
    - setup the verifier
- update `/auth` route to:
    - read `Authorization` header, reject if missing
    - use `OIDCVerifier.VerifyToken()` to validate the token (tolerate other audiences), which in most cases _will not_ require to make an HTTPS call (if key is found in keySet)
    - if pass, use the information from the token, not from the headers
    - compares the username, uid and groups headers with what's in the token

### Testing
- tested end to end with the guided deployment `aws-traefik-dex`

### DIFF
--- a/guided-charts/aws-traefik-dex/values.yaml
+++ b/guided-charts/aws-traefik-dex/values.yaml
@@ -92,12 +92,6 @@ dex:
     - 18000
     - 9800
 
-  # Client ID for authmiddleware to authenticate with Dex
-  authmiddlewareClientId: "authmiddleware"
-  # Client secret for authmiddleware to authenticate with Dex (auto-generated if empty)
-  # Generate with 'openssl rand -hex 16'
-  authmiddlewareClientSecret: "852c547d73dfaa51ef3bb290ed29ea5e"
-
 # Github OAuth configuration
 github:
   # GitHub OAuth client ID (required)
--- a/guided-charts/aws-traefik-dex/values.yaml
+++ b/guided-charts/aws-traefik-dex/values.yaml
@@ -91,7 +91,13 @@ dex:
     - 8000
     - 18000
     - 9800
-    
+
+  # Client ID for authmiddleware to authenticate with Dex
+  authmiddlewareClientId: "authmiddleware"
+  # Client secret for authmiddleware to authenticate with Dex (auto-generated if empty)
+  # Generate with 'openssl rand -hex 16'
+  authmiddlewareClientSecret: "852c547d73dfaa51ef3bb290ed29ea5e"
+
 # Github OAuth configuration
 github:
   # GitHub OAuth client ID (required)
@@ -183,3 +189,8 @@ authmiddleware:
   readTimeout: "10s"
   writeTimeout: "10s"
   shutdownTimeout: "30s"
+  # Bearer auth configuration
+  enableOauth: true
+  enableBearerAuth: false
+  # Timeout in seconds for OIDC provider initialization
+  oidcInitTimeoutSecs: 30
--- a/internal/authmiddleware/oidc_verifier.go
+++ b/internal/authmiddleware/oidc_verifier.go
@@ -29,7 +29,6 @@ type OIDCVerifier struct {
 	provider       *oidc.Provider
 	verifier       *oidc.IDTokenVerifier
 	clientID       string
-	clientSecret   string
 	issuerURL      string
 	logger         *slog.Logger
 	timeoutSeconds int // Timeout for OIDC provider initialization
@@ -63,23 +62,21 @@ func NewOIDCVerifier(config *Config, logger *slog.Logger) (*OIDCVerifier, error)
 	)
 
 	// Create the token verifier
-	// authmiddleware should be registered with OIDC provider (e.g. dex) as its own client
-	// so that it can access the public keys.
-	// However, in authmiddleware flow, the OIDC provider will issue token to a different component (e.g. oauth2-proxy)
+	// Note that any caller may obtain the public keys from the OIDC well-known page, a client ID
+	// and client secret are only needed to issue new token (and secret are optional for public
+	// access points).
+	// In authmiddleware flow, the OIDC provider will issue token to a different component (e.g. oauth2-proxy)
 	// and such component will pass the token as Authentication token to authmiddleware,
-	// thus, authmiddleware will not necessarily be in the audience of the token.
-	// In order to prevent the token verification to fail on target audience, set SkipClientIDCheck to true.
-	// This is okay: authmiddleware's only purpose is to verify that the token was issued by the OIDC provider.
+	// thus we use the client ID of the previous component as client.
 	oidcConfig := &oidc.Config{
 		ClientID:          config.OIDCClientID,
-		SkipClientIDCheck: true,
+		SkipClientIDCheck: false,
 	}
 
 	return &OIDCVerifier{
 		provider:       nil, // Will be initialized in Start()
 		verifier:       nil, // Will be initialized in Start()
 		clientID:       config.OIDCClientID,
-		clientSecret:   config.OIDCClientSecret,
 		issuerURL:      config.OIDCIssuerURL,
 		logger:         logger,
 		timeoutSeconds: config.OIDCInitTimeoutSecs,
--- a/internal/authmiddleware/oidc_verifier.go
+++ b/internal/authmiddleware/oidc_verifier.go
@@ -129,24 +129,6 @@ func (v *OIDCVerifier) Start(ctx context.Context) error {
 	return nil
 }
 
-// ExtractBearerToken extracts a bearer token from an Authorization header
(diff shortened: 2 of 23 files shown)


internal/authmiddleware/oidc_verifier.go
```diff
@@ -0,0 +1,220 @@
...
+	// This allows deferring HTTP calls until the application is ready
+	Start(ctx context.Context) error
+}
+
+// OIDCVerifier handles verification of OIDC tokens
+type OIDCVerifier struct {
+	provider       *oidc.Provider
+	verifier       *oidc.IDTokenVerifier
+	clientID       string
+	clientSecret   string
```
joshuatowner: where is this used?

joshuatowner (commented):
overall looks good

guided-charts/aws-traefik-dex/values.yaml
```diff
@@ -91,7 +91,13 @@ dex:
     - 8000
     - 18000
     - 9800
-    
+
+  # Client ID for authmiddleware to authenticate with Dex
+  authmiddlewareClientId: "authmiddleware"
+  # Client secret for authmiddleware to authenticate with Dex (auto-generated if empty)
```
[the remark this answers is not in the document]
JGuinegagne: I removed the value, `authmiddleware` doesn't need to register as a client w/ `dex` since it doesn't need to generate tokens.

joshuatowner (approved):

aws-prayags (approved):

internal/authmiddleware/oidc_verifier.go
```diff
@@ -0,0 +1,217 @@
...
+	if v.verifier == nil {
+		return nil, true, fmt.Errorf("OIDC verifier is not initialized - call Start() first")
+	}
+
+	// Verify the token
+	idToken, err := v.verifier.Verify(ctx, tokenString)
+	if err != nil {
+		// Check if this is a discovery document error
+		errMsg := err.Error()
+		if strings.Contains(errMsg, "failed to get discovery document") ||
```
andrii-i: Should error type assertions be used here instead of string matching to classify errors as string matching is / can be fragile?

internal/authmiddleware/serverroute_auth.go
```diff
@@ -43,10 +39,76 @@ func (s *Server) handleAuth(w http.ResponseWriter, r *http.Request) {
...
+	if headerUID != "" && k8sUID != headerUID {
+		s.logger.Error("UID mismatch between token and headers",
+			"token UID", k8sUID,
+			"header UID", headerUID)
+		http.Error(w, "UID verification failed", http.StatusUnauthorized)
+		return
+	}
+
+	// Verify groups in header if available
+	if len(headerGroups) > 0 {
```
andrii-i: @JGuinegagne you menioned planning to replace header but still can be apploed to critical parts of auth flow: 

should this / critical parts of auth flow have tests?

internal/authmiddleware/oidc_verifier.go
```diff
@@ -0,0 +1,217 @@
...
+		issuerURL:      config.OIDCIssuerURL,
+		logger:         logger,
+		timeoutSeconds: config.OIDCInitTimeoutSecs,
+		oidcConfig:     oidcConfig,
+	}, nil
+}
+
+// Start initializes the OIDC provider and verifier
+// This allows deferring HTTP calls until the application is ready
+func (v *OIDCVerifier) Start(ctx context.Context) error {
```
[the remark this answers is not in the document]
JGuinegagne: wdym?

The only remote call in this method is
```golang
provider, err := oidc.NewProvider(initCtx, v.issuerURL)
```
which fetches the keySet from `dex`. What are the write calls you are referring to?

internal/authmiddleware/serverroute_auth.go
JGuinegagne: WDYM sorry? The `authmiddleware` no longer trusts these headers, in fact it throws a 401 if the header value doesn't match what's in the token. I kept them here just for auditability.

PR #241, opened 2025-11-13
refactor (e2e): eliminate flakiness and improve e2e

andrii-i:
Eliminate flakiness and improve e2e suite:
* Consolidate controller deployment to BeforeSuite to eliminate race conditions
* Add CRD "Established" condition wait to prevent API server timing issues
* Add CONTAINER_TOOL env var support for CI compatibility (docker/finch)
* Set controller env vars via Kubernetes Downward API (declarative, no race conditions)
* Replace `EventuallyWithOffset` with `Eventually` and `time.Sleep` with deterministic waits
* Extract inline YAML to static files and use inline constants
* Add `dumpSetupDiagnostics` for failure debugging and enhanced cleanup

### DIFF
--- a/guided-charts/aws-hyperpod/templates/traefik/ingress.yaml
+++ b/guided-charts/aws-hyperpod/templates/traefik/ingress.yaml
@@ -13,6 +13,7 @@ metadata:
     alb.ingress.kubernetes.io/target-type: ip
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: '443'
+    alb.ingress.kubernetes.io/ssl-policy: 'ELBSecurityPolicy-TLS13-1-2-2021-06'
     {{- if .Values.awsCertificateArn }}
     alb.ingress.kubernetes.io/certificate-arn: {{ .Values.awsCertificateArn | quote }}
     {{- end }}


test/e2e/e2e_test.go
```diff
@@ -40,73 +39,53 @@ const namespace = "jupyter-k8s-system"
...
-		By("setting controller environment variables")
-		envVars := []string{
-			"CONTROLLER_POD_SERVICE_ACCOUNT=jupyter-k8s-controller-manager",
-			"CONTROLLER_POD_NAMESPACE=jupyter-k8s-system",
-		}
-		for _, envVar := range envVars {
-			cmd = exec.Command("kubectl", "set", "env", "deployment/jupyter-k8s-controller-manager", envVar, "-n", namespace)
-			_, err = utils.Run(cmd)
-			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to set environment variable: %s", envVar))
-		}
```
JGuinegagne: sanity-check: remove this? I think we need these introspection variables [here](https://github.com/jupyter-infra/jupyter-k8s/blob/124f41f993d1647665720c2d646aec07fe28fb64/internal/webhook/v1alpha1/pod_exec_webhook.go#L108)

test/e2e/helpers.go
```diff
@@ -0,0 +1,175 @@
...
+				if wsOutput, _ := utils.Run(cmd); len(wsOutput) > 0 && strings.TrimSpace(string(wsOutput)) != "" {
+					_, _ = GinkgoWriter.Write([]byte("   Template '" + tmplName + "' is referenced by workspaces: " +
+						strings.TrimSpace(string(wsOutput)) + " (test leaked resources)\n"))
+				} else {
+					_, _ = GinkgoWriter.Write([]byte("   Template '" + tmplName + "' has NO workspace references - CONTROLLER BUG?\n"))
+				}
+			}
+
+			// Emergency cleanup: Force-remove finalizers to allow suite teardown
+			By("EMERGENCY: Force-removing stuck template finalizers for clean teardown")
```
JGuinegagne: non-blocking: isn't this...hiding a deeper problem w/ template lifecycles?

JGuinegagne (dismissed):

test/e2e/helpers.go
andrii-i: If this emergency cleanup triggers it logs a warning that indicates a controller bug requiring investigation.  Without it, a stuck finalizer would silently timeout during test cleanup, potentially masking the root cause. So this cleanup makes bugs visible, not invisible.

Kubernetes E2E tests have similar finalizer cleanup logic https://github.com/kubernetes/kubernetes/blob/854e67bb51e177b4b9c012928d8271704e9cb80d/test/e2e/apimachinery/namespace.go#L600-L620

earaghbidikashani (approved):

PR #253, opened 2025-11-14
feat: add maximum cache size to KMS cache

### LINKED ISSUE #193: Add max number of keys to KMS data keys cache

joshuatowner:
Resolves #193


aws-prayags (approved):
Non blocking - debug logs on the non frequent cache activities like eviction would be helpful

aakashmandavilli96 (approved):
LGTM

Issue #259, opened 2025-11-14
Fetch AccessStrategy only once per Workspace reconciliation loop


JGuinegagne:
Right now, AccessStrategy is fetched twice:
- once to parametrize the deployment, in [createDeployment](https://github.com/jupyter-infra/jupyter-k8s/blob/78f1e87c0f1051a0f5e74f86e63a145a21544ca8/internal/controller/resource_manager.go#L93) or [updateDeployment](https://github.com/jupyter-infra/jupyter-k8s/blob/78f1e87c0f1051a0f5e74f86e63a145a21544ca8/internal/controller/resource_manager.go#L228)
- once for accessResources in [ReconcileAccessForDesiredRunningStatus](https://github.com/jupyter-infra/jupyter-k8s/blob/78f1e87c0f1051a0f5e74f86e63a145a21544ca8/internal/controller/state_machine_access.go#L19)

Update this logic to fetch it only once, preferably in workspace_controller [Reconcile](https://github.com/jupyter-infra/jupyter-k8s/blob/78f1e87c0f1051a0f5e74f86e63a145a21544ca8/internal/controller/workspace_controller.go#L123) method.

Issue #257, opened 2025-11-14
Ensure access strategy resources match the latest AccessStrategy definition


JGuinegagne:
The reconciliation loop logic for `Workspace` with an `AccessStrategy` that defines access resources is incomplete.

If an access strategy 1/ is defined in `AccessStrategy`, 2/ exists in status, and 3/ is found by k8s client, then the reconciliation is currently a no-op, [here](https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/controller/resource_manager_access.go#L116). This implicitely assumes that the access resources are correct, ie A/ have not been modified elsewhere, and B/ that the `AccessStrategy` has not changed.

This assumption is not valid, we should instead confirm that the spec match the template defined in the `AccessStrategy`.

PR #268, opened 2025-11-15
fix: access strategy-workspace lifecyle

JGuinegagne:
This PR fixes issues between access strategies - workspaces lifecycle.
- addresses #259 
- addresses #257 
- ensures a `Workspace` may not be created if it references an access strategy that does not exists
- ensures an `AccessStrategy` does not get fully deleted as long as any `Workspace` references it (using finalizers)

### Implementation
- update `Workspace` reconcile loop to fetch `AccessStrategy` once
    - if the workspace references an `AccessStrategy`
    - if the workspace desired state is `Running`
- update `Workspace` reconcile loop to better handle access resource edge cases:
    - if access resource A/ is defined in `AccessStrategy`, B/ exists in etcd and C/ is referenced in `Workspace.status.accessResources`, ensure its spec conform to whatever the latest access strategy has
    - if access resources A/ is not defined in `AccessStrategy`, B/ exists in etcd and C/ is referenced in `Workspace.status.accessResources`, ensure the resource is deleted
- trigger `Workspace` reconciliation for all workspaces that reference an `AccessStrategy` that's updated
- update `Workspace` admission webhook to:
    - add finalizer for `AccessStrategy`
    - fail webhook if `AccessStrategy` does not exist
- add `AccessStrategy` reconciliation loop to add/remove finalizers based on whether there exists `Workspace` that reference it
- add method in `workspace/queries` to lookup workspaces by `AccessStrategy` labels

### Testing
- tested end-to-end in kind (creating access strategy, create Workspace, delete Workspace/AccessStrategy, verify behavior)
- same tests in EKS cluster
- tested updating access strategy (e.g. remove all access resources, modify them, verify the related Workspaces access resources are immediately update/deleted/created)
- tested adding or removing `AccessStrategyRef` from a `Workspace.spec`, verified behavior of `AccessResources`

### DIFF
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -87,10 +87,7 @@ const (
 	IdleCheckInterval = 5 * time.Minute
 
 	// WorkspaceFinalizerName is the finalizer name for workspace cleanup protection
-	WorkspaceFinalizerName = "workspace.jupyter.org/cleanup-protection"
-
-	// AccessStrategyFinalizerName is the finalizer name for access strategy protection
-	AccessStrategyFinalizerName = "workspace.jupyter.org/accessstrategy-protection"
+	WorkspaceFinalizerName = "workspace.jupyter.org/workspace-protection"
 
 	// ControllerPodNamespaceEnv is the environment variable for the controller pod namespace
 	ControllerPodNamespaceEnv = "CONTROLLER_POD_NAMESPACE"
--- a/internal/controller/resource_manager_access.go
+++ b/internal/controller/resource_manager_access.go
@@ -184,7 +184,11 @@ func (rm *ResourceManager) ensureAccessResourceExists(
 				return fmt.Errorf("error getting expected spec: %w", err)
 			}
 
-			// If specs are different, update the resource
+			// If specs are different, update the access resource
+			// This captures two different scenarios
+			// case 1: the access resource was modified by another process (in spite of owner reference)
+			// case 2: the AccessStrategy modified the resource template, so that the actual resource
+			// no longer conforms to the resource defined in AccessStrategy
 			if existingFound && expectedFound && !reflect.DeepEqual(existingSpec, expectedSpec) {
 				logger.Info("AccessResource spec doesn't match template, updating",
 					"kind", existingObj.GetKind(),
--- a/internal/controller/workspace_controller.go
+++ b/internal/controller/workspace_controller.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	builderPkg "sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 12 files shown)


internal/controller/constants.go
```diff
@@ -86,6 +86,9 @@ const (
 	// WorkspaceFinalizerName is the finalizer name for workspace cleanup protection
 	WorkspaceFinalizerName = "workspace.jupyter.org/cleanup-protection"
```
aws-prayags: non-blocking - following your naming, should we also update this to `workspace.jupyter.org/workspace-protection`

aws-prayags (approved):

internal/workspace/finalizers.go
```diff
@@ -0,0 +1,153 @@
+package workspace
+
+import (
+	"context"
+	"math/rand"
+	"time"
+
+	"github.com/go-logr/logr"
+	workspacev1alpha1 "github.com/jupyter-ai-contrib/jupyter-k8s/api/v1alpha1"
+	webhookconst "github.com/jupyter-ai-contrib/jupyter-k8s/internal/webhook"
```
joshuatowner: what to we store int his webhook const package, this means dependency chain is like: webhook -> workspace pkg -> webhook const pkg?

we have to be very careful not to introduce dep loops here. i'd prefer to keep the workspace package not dependent on extension, controller, or webhook to prevent issues in future. and maybe store these general constants in the common package.

internal/controller/resource_manager_access.go
```diff
@@ -113,7 +166,46 @@ func (rm *ResourceManager) ensureAccessResourceExists(
...
+			if err != nil {
+				return fmt.Errorf("error getting existing spec: %w", err)
+			}
+
+			expectedSpec, expectedFound, err := unstructured.NestedFieldCopy(expectedObj.Object, "spec")
+			if err != nil {
+				return fmt.Errorf("error getting expected spec: %w", err)
+			}
+
+			// If specs are different, update the resource
```
joshuatowner: we do need to be careful with this one, sometimes spec gets updated due to creation of resource. we had this happen with PVC spec being updated with created PV. OK with it for now since traefik I don't think we have this issue, just calling out.

internal/controller/resource_manager_access.go
```diff
@@ -113,7 +166,46 @@ func (rm *ResourceManager) ensureAccessResourceExists(
...
+			if existingFound && expectedFound && !reflect.DeepEqual(existingSpec, expectedSpec) {
+				logger.Info("AccessResource spec doesn't match template, updating",
+					"kind", existingObj.GetKind(),
+					"name", existingObj.GetName(),
+					"namespace", existingObj.GetNamespace())
+
+				// Preserve metadata from existing object
```
joshuatowner: could you provide info on this one?

internal/controller/workspaceaccessstrategy_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	// Case 3: State is correct (both have workspaces+finalizer, or neither)
+	logger.V(1).Info("Finalizer state is correct, no action needed")
+	return ctrl.Result{}, nil
+}
+
+// handleDeletion handles deletion of AccessStrategy and manages finalizers
+// based on whether any workspaces are still using the AccessStrategy
+func (r *WorkspaceAccessStrategyReconciler) handleDeletion(ctx context.Context, accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy) (ctrl.Result, error) {
```
joshuatowner: nit: can we just run logic based on the finalizer only? why duplicate logic around ListActiveWorkspacesByAccessStrategy ?

internal/workspace/finalizers.go
```diff
@@ -0,0 +1,153 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// SafelyAddFinalizerToAccessStrategy attempts to add a finalizer to an AccessStrategy,
+// handling conflict errors by checking if the finalizer was added in a concurrent operation.
+func SafelyAddFinalizerToAccessStrategy(
```
joshuatowner: why not just re-queue reconcile instead of blocking reconcile thread? treat it as retryable error

joshuatowner (commented):
we will now be saying that updating access strategy means that all spaces couls be interrupted (due to deployment modifications). this is a bit different from templates where updates are not applied to existing spaces (i know different use case as well). not opposed as long as this is very clear and warned to users.

internal/controller/workspace_controller.go
```diff
@@ -100,6 +102,8 @@ func (r *WorkspaceReconciler) SetStateMachine(sm *StateMachine) {
 // move the current state of the cluster closer to the desired state.
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.21.0/pkg/reconcile
+//
+// nolint:gocyclo
```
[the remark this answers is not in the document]
JGuinegagne: yeah, let's track as a separate issue and refactor next week

andrii-i (approved):
Approved with caveat / requirement of adding  label filtering to `accessStrategyEventHandler` as it’s a small change with large impact.

Non-blocking but let's track adding tests. I'd say in terms of priority:
1. E2E test for AccessStrategy deletion blocking
2. Unit tests for `WorkspaceAccessStrategyReconciler` finalizer logic
3. Unit tests for `SafelyAddFinalizerToAccessStrategy`, `SafelyRemoveFinalizerFromAccessStrategy`

internal/controller/workspace_controller.go
```diff
@@ -320,3 +375,58 @@ func (r *WorkspaceReconciler) getWorkspace(ctx context.Context, req ctrl.Request
...
+		return nil
+	}
+
+	logger.Info("Handling AccessStrategy event",
+		"accessStrategy", accessStrategy.Name,
+		"namespace", accessStrategy.Namespace)
+
+	// Find all Workspaces that reference this AccessStrategy
+	var workspaceList workspacev1alpha1.WorkspaceList
+	if err := r.List(ctx, &workspaceList); err != nil {
```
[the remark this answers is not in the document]
JGuinegagne: woaw, miss this one, thanks! Will definitely add. 
It needs to paginate as well.

internal/workspace/finalizers.go
andrii-i: I'd put constants into PACKAGE/constants.go as convention (already used in the codebase) and then evaluate how much duplication there is / if higher level constants file is needed.

internal/controller/resource_manager_access.go
```diff
@@ -57,19 +58,71 @@ func (rm *ResourceManager) EnsureAccessResourcesExist(
...
+			// Keep resources that are still in the AccessStrategy
+			resourcesToKeep = append(resourcesToKeep, resource)
+		}
+	}
+
+	// Delete resources that are no longer in the AccessStrategy
+	resourcesDeleted := false
+	for _, resource := range resourcesToDelete {
+		resourceCopy := resource // Create a copy to avoid pointer issues in the loop
+		removed, err := rm.ensureAccessResourceDeleted(ctx, &resourceCopy)
```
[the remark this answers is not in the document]
JGuinegagne: Long term, we should. Right now, the resources that `AccessStrategy` creates are immediately deleted, so I think this is okay.

internal/workspace/finalizers.go
JGuinegagne: moving to `workspace/constants.go` for now

joshuatowner (approved):

aws-prayags (approved):

Issue #262, opened 2025-11-14
Fix AccessStrategy.spec attributes to make them consistent


JGuinegagne:
Update [AccessStrategy.spec](https://github.com/jupyter-infra/jupyter-k8s/blob/main/api/v1alpha1/workspaceaccessstrategy_types.go)
```yaml
{
  displayName
  accessResourceTemplates: []AccessResourceTemplate
  accessUrlTemplate
  bearerUrlTemplate

  podEventsHandler
  podEventsContext: map[string]string
  createConnectionHandler
  createConnectionContext: map[string]string

  deploymentModifications {
    podModifications {
      addtionalContainers: []corev1.Container
      initContainers: []corev1.Container
      volumes: []covev1.Volumes
      primaryContainerModifications {
        mergeEnv: []AccessEnvTemplate
        volumeMounts: []corev1.VolumeMount
      }
    }
  }
}
```

PR #271, opened 2025-11-16
feat: encryption context and new access strategy fields

### LINKED ISSUE #65: Add support for lifecycle callbacks in WorkspaceAccessStrategy spec
Also, address these comments specifically 
1/ https://github.com/jupyter-infra/jupyter-k8s/pull/58#discussion_r2456921803
2/ https://github.com/jupyter-infra/jupyter-k8s/pull/58#discussion_r2456925959

### LINKED ISSUE #220: Add encryption context for KMS calls

joshuatowner:
- Use new access strategy fields (*handler, *context) instead of controller config 
- add encryption context to requests
- added tests for new logic

current architectural approach:

access strategy -> getSignerForAccessStrategy(strategy) -> signer -> signer.GenerateToken(...)

the getSignerForAccessStrategy handles all impl specific items (like KMS key) to keep the signer generic in the business logic. currently, we create one signer per request, created on the fly as we see access strategies. In future, we can cache this based on the version of the access strategy. 

Partially addresses: #262 

Resolves #65
Resolves #220


test-pods.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: v1
```
JGuinegagne: You probably didn't intend to commit this, or at least under the root of the project :)

internal/extensionapi/connection_route.go
```diff
@@ -39,10 +39,35 @@ func (s *ExtensionServer) generateWebUIBearerTokenURL(r *http.Request, workspace
...
+	ws, err := s.getWorkspace(namespace, workspaceName)
 	if err != nil {
-		return "", "", err
+		return "", "", fmt.Errorf("failed to get workspace: %w", err)
+	}
+
+	accessStrategy, err := s.getAccessStrategy(ws)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get access strategy: %w", err)
+	}
```
JGuinegagne: sanity-check: if I follow the logic, if the workspace has no access strategy, then it will result in a 500 / internal server error. Is that your intention?

internal/extensionapi/connection_route.go
```diff
@@ -39,10 +39,35 @@ func (s *ExtensionServer) generateWebUIBearerTokenURL(r *http.Request, workspace
...
+	// Require AccessStrategy with BearerAuthURLTemplate
+	if accessStrategy == nil {
+		return "", "", fmt.Errorf("no AccessStrategy configured for workspace")
+	}
+	if accessStrategy.Spec.BearerAuthURLTemplate == "" {
+		return "", "", fmt.Errorf("BearerAuthURLTemplate not configured in AccessStrategy")
+	}
+
+	// Create signer based on access strategy
+	signer, err := s.signerFactory.CreateSigner(accessStrategy)
```
JGuinegagne: sanity-check: create signer at each request?

guided-charts/aws-hyperpod/values.yaml
```diff
@@ -48,6 +48,11 @@ clusterWebUI:
         memory: 128Mi
     jwtSigningType: "kms"
     kmsKeyId: ""
+    # KMS encryption context configuration
+    kmsEncryptionContext:
```
aws-prayags: why does this need to be configurable via values ? 
if `kmsKeyId` is provided, we should enable this ?

aws-prayags (approved):

Issue #70, opened 2025-10-23
Improvement for ssm_client.go


aws-prayags:
1/ Failure handling for all methods including deregisterManagedInstance
2/ Add retry with back-off logic
3/ Handle throttling from SSM service
4/ Question "what's the difference between StreamURL and WebSocketURL, why do we need both?" - https://github.com/jupyter-infra/jupyter-k8s/pull/58#discussion_r2452863451
5/ During cleanup, operator should also delete activations
6/ For ssm describe session, loop till you get 10 results - pagination - https://github.com/jupyter-infra/jupyter-k8s/pull/231

aws-prayags:
tracking https://github.com/jupyter-infra/jupyter-k8s/issues/272 as non ga-blocker

PR #275, opened 2025-11-17
fix: ssm client updated retry strategy, retrieve latest managed instance

aws-prayags:
Fixes https://github.com/jupyter-infra/jupyter-k8s/issues/70

### Testing
Tested on EKS cluster by creating a workspace and connecting to it remotely

### DIFF
--- a/internal/aws/ssm_remote_access_strategy.go
+++ b/internal/aws/ssm_remote_access_strategy.go
@@ -2,8 +2,11 @@ package aws
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
+	"time"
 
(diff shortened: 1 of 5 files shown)


andrii-i (approved):
LGTM, thank you @aws-prayags. 

Non-blocking comments / suggestions:
* Replace O(n log n) sort with O(n) linear scan in most recently registered instance finding logic in ssm_client.go LL128-137 
* In future PRs we could / should add defensive duplicate detection and cleanup on / before instance registration. This way duplicate instances won't have a chance to accumulate.
* No E2E tests, I'd also add UTs for more cases, for example a large one missing in my opinion is ophaned instances handling. Let's track the need to add both

PR #306, opened 2025-12-03
fix: discovery handler not properly registering connection resources

joshuatowner:
Fixes the discovery handling issues for workspace connections by turning of the built-in discovery handler of generic API, which causes a conflict. Also turns off the fairness controller, which required extra permissions to call flowschema & pollutes startup logs. we can consider re-adding in future.

this allows for the original CLI UX:
```
kubectl create -f - -o yaml <<EOF                                                                                                          
apiVersion: connection.workspace.jupyter.org/v1alpha1                                                                                                                                    kind: WorkspaceConnection
metadata:
 namespace: default     
spec:
 workspaceName: workspace-sample
 workspaceConnectionType: vscode-remote                                                                                                                                                  EOF
```

### DIFF
--- a/internal/extensionapi/server.go
+++ b/internal/extensionapi/server.go
@@ -211,6 +211,9 @@ func createRecommendedOptions(config *ExtensionConfig) *genericoptions.Recommend
 	recommendedOptions.SecureServing.ServerCert.CertKey.KeyFile = config.KeyPath
 	recommendedOptions.SecureServing.ServerCert.PairName = "tls"
 
+	// turn off advanced priority and fairness features that require extra permissions
+	recommendedOptions.Features.EnablePriorityAndFairness = false
+
 	return recommendedOptions
 }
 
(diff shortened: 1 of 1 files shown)


internal/extensionapi/server.go
```diff
@@ -211,13 +211,17 @@ func createRecommendedOptions(config *ExtensionConfig) *genericoptions.Recommend
...
+	recommendedOptions.Features.EnablePriorityAndFairness = false
+
 	return recommendedOptions
 }
 
 // createGenericAPIServer creates a GenericAPIServer from options
 func createGenericAPIServer(recommendedOptions *genericoptions.RecommendedOptions) (*genericapiserver.GenericAPIServer, error) {
 	// Create server config
 	serverConfig := genericapiserver.NewRecommendedConfig(codecs)
+	serverConfig.EnableDiscovery = false
```
aws-prayags: non-blocking - would be good to add a comment here that notes we've implemented our own discovery and this causes issue

aws-jasakshi (approved):

Issue #279, opened 2025-11-17
Access Strategy e2e tests


JGuinegagne:
Context: fast follow from #268 

Add e2e tests for `AccessStrategy`:
1. `Workspace` w/ `AccessStrategy` AND `desiredState=running` should:
    1.1. have its deployment modified
    1.2. create access resources
    1.3. template the access URL
    1.4. track access resources in `Workspace.status`
2. `Workspace` w/ `AccessStrategy` AND `desiredState=stopped` should not create access resources nor display access URL
3. Running `Workspace` updated to `desiredState=stopped` should delete the access resources and stop showing access URL
4. `Workspace` referencing an non-existent `AccessStrategy` should fail the admission webhook
5. `Workspace` that reference an `AccessStrategy` should:
    5.1. place a finalizer on `AccessStrategy`
    5.2. keep the finalizer on `desiredState=stopped`
    5.3. prevent deletion on `AccessStrategy`
    5.4. remove the finalizer on `delete Workspace` and allow deletion of `AccessStrategy` to proceed 
6. similar logic with multiple workspaces
7. `Workspace` that de-references an `AccessStrategy` on update should:
    7.1. delete its access resources
    7.2. stop tracking access resources in `Workspace.status`
    7.3. stop showing access URL
8. `AccessStrategy` that receives an update:
    8.1. adding an access resource --> should cause the access resource to be added for each referencing `Workspaces`
    8.2. updating an access resource --> should cause the access resource to be updated for each referencing `Workspaces`
    8.3. removing an access resource --> should cause the access resource to be deleted for each referencing `Workspaces`
    8.4. updating the access URL template --> should cause the `Workspace.status.accessUrl` to be modified
    8.5. updating an env var --> should cause the deployment of each underlying `Workspace` to be recreated to pick up the change

PR #307, opened 2025-12-03
test: e2e revamp, access strategy e2e

JGuinegagne:
This PR revamps the e2e tests:
1. add linting coverage for e2e tests
2. reorganize yaml file for resource creation under `test/e2e/static/<group>-<subgroup>/`
3. simplify tests utilities
4. rename test suites, update tests code to more readable and standard
5. remove irrelevant tests (e.g. add multiple finalizers; functionality to be tested is resource ownership/preservation
6. load application images into `kind` cluster
7. uses only application images loaded in cluster (to avoid live docker pull)
8. always create workspaces in `default` namespace, template & access strategies in `jupyter-k8s-shared` namespace
9. attempt to harden webhook registration test in main test suite

Also in this PR:
10. add e2e test coverage for `access strategy` (partially addressing #279 )
11. expand status e2e tests
12. rename 'resource' test suite > 'storage'
13. re-enable scheduling e2e tests

### Temporary hack
- `finch` based tests on cloud-desktop cannot perform file system operation due to `cgroups` permission access, skip tests if it detects that engine is `finch` (not the case in the github hooks)

### Testing
- ran each test suite independently

### DIFF
--- a/test/e2e/static/access-strategy/workspace-stopped-with-access-strategy.yaml
+++ b/test/e2e/static/access-strategy/workspace-stopped-with-access-strategy.yaml
@@ -11,8 +11,8 @@ spec:
     namespace: jupyter-k8s-shared
   resources:
     requests:
-      cpu: "100m"
-      memory: "128Mi"
+      cpu: 100m
+      memory: 128Mi
     limits:
-      cpu: "500m"
-      memory: "512Mi"
+      cpu: 500m
+      memory: 512Mi
--- a/test/e2e/static/access-strategy/workspace-with-access-strategy.yaml
+++ b/test/e2e/static/access-strategy/workspace-with-access-strategy.yaml
@@ -11,8 +11,8 @@ spec:
     namespace: jupyter-k8s-shared
   resources:
     requests:
-      cpu: "100m"
-      memory: "128Mi"
+      cpu: 100m
+      memory: 128Mi
     limits:
-      cpu: "500m"
-      memory: "512Mi"
+      cpu: 500m
+      memory: 512Mi
--- a/test/e2e/static/access-strategy/workspace-with-env-access-strategy.yaml
+++ b/test/e2e/static/access-strategy/workspace-with-env-access-strategy.yaml
@@ -11,8 +11,8 @@ spec:
     namespace: jupyter-k8s-shared
   resources:
     requests:
-      cpu: "100m"
-      memory: "128Mi"
+      cpu: 100m
+      memory: 128Mi
     limits:
-      cpu: "500m"
-      memory: "512Mi"
+      cpu: 500m
+      memory: 512Mi
--- a/test/e2e/static/access-strategy/workspace-with-nonexistent-access-strategy.yaml
+++ b/test/e2e/static/access-strategy/workspace-with-nonexistent-access-strategy.yaml
@@ -11,8 +11,8 @@ spec:
     namespace: jupyter-k8s-shared
   resources:
     requests:
-      cpu: "100m"
-      memory: "128Mi"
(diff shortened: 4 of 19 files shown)


test/e2e/e2e_suite_test.go
andrii-i: [Out of scope of this PR / not relevant to its changes] This `Eventually` block could / should do 1 cmd call instead of 3

andrii-i (approved):
Looks good, thank you for this work. A couple of non-blocking improvements to discuss below.

1. [Non-blocking] YAML quoting style 

This PR enforces use of unquoted strings for resource values in some yaml files (`cpu: 500m`):
* `test/e2e/static/storage-base/workspace-custom-mountpath.yaml`, 
* `test/e2e/static/storage-template/storage-template.yaml`

And also introduces some yaml files that use quoted strings (`cpu: "500m”`):
* `test/e2e/static/access-strategy/workspace-stopped-with-access-strategy.yaml`,
* `test/e2e/static/access-strategy/workspace-with-access-strategy.yaml`, 
* `test/e2e/static/access-strategy/workspace-with-env-access-strategy.yaml`, 
* `test/e2e/static/access-strategy/workspace-with-nonexistent-access-strategy.yaml`,
* `test/e2e/static/access-strategy/workspace-with-resources-access-strategy.yaml` 

Let’s choose a quoting convention to follow (I’m suggesting unquoted strings for simplicity) and consistently use it across the codebase in all example and test yaml files. This can be automated and enforced with yaml linters like `yamllint` or `kube-linter`. 

In this PR let's just make sure single convention is used across all yaml files modified or created in this PR.

2. [Optional] Inline parameters in e2e 

Should we define timeouts, polling intervals, and other inline numbers as constants in `test/e2e/constants.go`? 

If yes, we could either apply to new/modified files now (workspace_access_strategy_test.go, helpers_workspaces.go, helpers_volume.go, e2e_suite_test.go) or refactor for all e2e tests in a separate PR.

test/e2e/e2e_suite_test.go
JGuinegagne: Thanks @andrii-i 
- let me fix the quoting style.
- adding a yaml linter makes sense, we should do it but add it in a different that can be only linting change.
- moving timeouts/polling intervals to constants also make sense, same thing though: let's do it in a dedicated PR

test/e2e/e2e_suite_test.go
```diff
@@ -119,29 +121,68 @@ var _ = BeforeSuite(func() {
...
 		"Namespace not in Active phase")
 
 	By("labeling the namespace to enforce the restricted security policy")
-	cmd = exec.Command("kubectl", "label", "--overwrite", "ns", namespace,
+	cmd = exec.Command("kubectl", "label", "--overwrite", "ns", OperatorNamespace,
 		"pod-security.kubernetes.io/enforce=restricted")
 	_, err = utils.Run(cmd)
 	Expect(err).NotTo(HaveOccurred(), "Failed to label namespace with pod security policy")
 
+	By("installing Traefik CRDs for access strategy tests")
```
joshuatowner: nit: in future we should split all this up. maybe a traefik installation helper (same with cert manager, etc). so we can pick and choose for subsets of tests as well.

test/e2e/e2e_test.go
```diff
@@ -23,13 +22,6 @@ import (
 	"github.com/jupyter-infra/jupyter-k8s/test/utils"
 )
 
-// Test constants
```
joshuatowner: nit: what does this test do? is it just a catch all?

joshuatowner (approved):

PR #310, opened 2025-12-06
test: access strategy e2e coverage

JGuinegagne:
This PR closes the gap on access strategy end-to-end test:
- verify behavior when workspace status change while it references an access strategy
    - running > stopped
    - stopped > running
- verify access strategy deletion protection
    - when not referenced by any workspace
    - when referenced by running workspace
    - when referenced by stopped workspace
- verify behavior of updating a running workspace
   - to add access strategy ref
   - to remove access strategy ref
- verify behavior of updating an access strategy while referenced by workspaces
   - add access resources
   - update access resources
   - delete access resources
   - update env vars/ access URL

Addresses #279


andrii-i (approved):
Thank you @JGuinegagne. Everything looks good, a couple of comments / suggestions for your consideration:
1. This PR adds a large number of inline numbers for testing timeouts etc in `test/e2e/helpers_access_strategy.go` and `test/e2e/workspace_access_strategy_test.go`. In https://github.com/jupyter-infra/jupyter-k8s/pull/307#pullrequestreview-3540688453 we aligned that it would be nice to define timeouts, polling intervals, and other inline numbers as constants in `test/e2e/constants.go`. Should it be done in this PR or tracked as a follow-up issue to ensure it doesn't get lost? Non-blocking, for your consideration.
2. Should there also be a test / tests for multiple workspaces referencing same AccessStrategy and associated finalizers, workspace operations, AccessStrategy mutation propagation logic? This is point 6 from #279 which would be closed when this PR is merged. Non-blocking, but want to ensure we don't lose this coverage.

PR #320, opened 2026-01-14
feat: auto-rotate JWT key for authmiddleware

### LINKED ISSUE #228: OSS: JWT key rotation
## Problem
The JWT signing key for the [StandardSigner](https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/jwt/standard_signer.go#L13) does not rotate, which is not a security good practice.

### Threat vector
If an attacker breaks the JWT signing key, they can create their own cookie and gain access to Workspaces they shouldn't be allowed to.

## Solution
Implement a kubernetes native secret rotation of the JWT signing key. We will use a `CronJob` that updates the k8s secret regularly. The CronJob will a/ update the secret by adding a key: value, w/ key=UTC-timestamp, value=generated-signing-key, b/ remove stale keys. Each of the `authmiddleware` pods will A/ add a `key-timestamp` to the JWT, B/ if encounter a key timestamp it doesn't know, invalidate its cache, C/ watch the specific k8s secret to update its in-memory cache. 

### rotator package
Add a `rotator` package under `internal`, which updates a k8s secret w/ `crypto/rand`

### rotator image
Add [another image](https://github.com/jupyter-infra/jupyter-k8s/blob/main/images/authmiddleware/Dockerfile) for secret rotation that under uses a new `rotator` package

### guided deployment
Add the `CronJob` setup, with dedicated RBAC

### authmiddleware
Watch the secret value, update the `JWTHandler` that's used at runtime

JGuinegagne:
This PR implements an automatic rotation for the JWT seed used by the `authmiddleware`.

Addresses #228 

This change only impacts the oauth guided deployment since the hyperpod deployment uses KMS to sign the JWT embedded in the cookie.

### User experience
- no change, the rotation should be invisible to the users

### Mechanism
- add a `CronJob` resource running every 15 mins in the `jupyter-k8s-router` namespace
- the `CronJob` updates the `authmiddleware` secrets, and add keys
- `authmiddleware` pods watches the secret, and update its in-memory key set
- `authmiddleware` pods recognize up to 6 previous keys, but if it detects a previous key on `/verify` route, it issues it w/ the new key
-  `authmiddleware` pods enforce a 5s cooloff period before using a newly rotated key to ensure each of the pods got the memo that this new key exist

### Implementation
- added a new image in `images/rotator` (distroless, same `go mod`)
- added `internal/rotator`
- update `internal/authmiddleware` to a/ use manager (instead of standalone HTTP server), b/ handle key rotation, c/ configure the manager to watch the secret, d/ add parameters to `config`
- update `guided-charts/aws-traefik-dex` yaml file to a/ configure the new secret, b/ configure the `CronJob` (w/ rbac), c/ use the new parameters in `authmiddleware` deployment
- add `config-auth` dir that leverages `kustomize` to deploy a/ `authmiddleware`, b/ `rotator` against local `kind` cluster, and added related `Makefile` methods 

### Testing
- tested with the new `makefile` methods in `kind`
- wiped the `aws-traefik-dex` deployment on my EKS cluster, redeployed, tested manually end to end
    - accessed a new notebook (worked)
    - forced rotated the key, reloaded the tab, worked seamlessly
    - decoded the cookie and verified that it is now using the new key
    - waited for an automatic rotate, performed same verification


andrii-i (dismissed):
Looks good, thank you @JGuinegagne. Please find some non-blocking improvement suggestions below (good to be handled separately).

PR size makes sense as multi-key signing, rotator, and deployment via helm chart all had to be developed and tested together / are essential parts of the same feature.

Improvement suggestions:
* Graceful shutdown of the JWT auth HTTP server in `internal/authmiddleware/server.go`
* Use of `log` instead of `fmt.Printf` in `internal/rotator/rotator.go` for in-package consistency with `cmd/rotator/main.go`
* `jwt-signing-key` naming in `config-auth/authmiddleware/secret.yaml`

config-auth/README.md
```diff
@@ -0,0 +1,138 @@
+# Auth Components for Local Kind Testing
+
+This directory contains Kustomize manifests for deploying the authmiddleware and JWT rotator to a local Kind cluster for development and testing.
```
joshuatowner: hmm, should we consider how to structure this? maybe we can also vend as a subchart or something. would like to avoid fragmentation between tests and vended charts.

guided-charts/aws-hyperpod/templates/authmiddleware/deployment.yaml
```diff
@@ -68,6 +68,8 @@ spec:
         env:
           - name: PORT
             value: "8080"
+          - name: NAMESPACE
+            value: "{{ .Values.namespace }}"
```
joshuatowner: this should pull from Release.namespace not Values.namespace

internal/authmiddleware/http_server.go
```diff
@@ -0,0 +1,97 @@
...
+	logger         logr.Logger
+	runtimeClient  client.Client
+	standardSigner *jwt.StandardSigner
+	secretName     string
+	namespace      string
+}
+
+// NewHTTPServerRunnable creates a new HTTPServerRunnable.
+// If standardSigner is not nil, it will load the initial JWT signing keys before starting the server.
+func NewHTTPServerRunnable(
```
joshuatowner: feels unnecessary, why can't we do initial lookup as part of the initialization flow? is there a value to being in `Start`?

internal/jwt/standard_signer.go
```diff
@@ -102,3 +179,72 @@ func (s *StandardSigner) ValidateToken(tokenString string) (*Claims, error) {
...
+	s.keyAddedTimes = newKeyAddedTimes
+	s.latestKid = latestKid
+
+	return nil
+}
+
+// RetrieveInitialSecret loads the initial JWT signing keys from the Kubernetes secret.
+// This is called when the HTTP server starts to ensure keys are loaded before accepting requests.
+// The parseFunc parameter is a function that parses signing keys from a secret - it's injected
+// to avoid circular dependencies with the rotator package.
```
joshuatowner: if this is a concern maybe `jwt` package can contain all of the logic around generating and parsing the secrets, and keep rotator package to just call the common jwt package parsing.

joshuatowner (commented):
overall looks good just structural comments

joshuatowner (approved):

andrii-i (approved):

Issue #14, opened 2025-10-07
Consider `WorkspaceTemplate` env vars overwrite strategy


andrii-i:
`Env` is an interesting one. We need to think about this carefully.

Should we:
1. override the whole env (as you are doing)
2. allow merging user vars & template vars; if so how?
3. support admin to disallow passing certain env vars; if so how?
4. something else?

_Originally posted by @JGuinegagne in https://github.com/jupyter-ai-contrib/jupyter-k8s/pull/10#discussion_r2399031302_

JGuinegagne:
Let's evaluate whether this can go in before code freeze. Otherwise, I would vote to remove [EnvironmentVariables](https://github.com/jupyter-infra/jupyter-k8s/blob/060c606e7227322ac66b96991426453f9c71bb6c/api/v1alpha1/workspacetemplate_types.go#L71) attribute from Template for now.

JGuinegagne:
Addressed in:
- #324 
- #327

PR #324, opened 2026-02-24
feat: env variable in container config

henrywa2:
Adding environment support in container config. This partially addresses https://github.com/jupyter-infra/jupyter-k8s/issues/14.

Behavior:
- Users can input env variables the same way they would for pod containers
- Template can define default env variables
- If users input _anything_ as part of their workspace spec container config, the entire default container config in template will be fully overridden. 

Other options that were considered:
- Merging workspace spec env variables with template defaults while overriding template defaults for any env variables that exist in both. There might be cases where the admin wants some variables to always be in the template
- In future if there is a request, we can add `MergeEnvFromTemplate` field in workspace container config spec so users can choose whether they want to merge or override

Reasons to go with full override:
1. If there was env variable merging, there would be no way for DS to get rid of env variables on the template except setting to empty string
2. If DS creates workspace using the UI, there can be some logic where the template env variables are already filled in so DS can add to those if they want

### DIFF
--- a/test/e2e/workspace_env_test.go
+++ b/test/e2e/workspace_env_test.go
@@ -45,14 +45,6 @@ var _ = Describe("Workspace Environment Variables", Ordered, func() {
 				ConditionTrue,
 			)
 
-			By("verifying Available=True, Progressing=False, Degraded=False, Stopped=False")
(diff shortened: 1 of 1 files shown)


andrii-i (approved):
LGTM pending Jonathans comments implementation. Thanks for addressing unit test suggestion.

PR #326, opened 2026-03-04
feat: k8s native bearer tokens

JGuinegagne:
This PR implements the first step toward adopting a plugin architecture, which separates any aws-specific logic from the core operator. The `extensionapi`, which responds to `Create:Connection` API calls, can optionally sign JWT using a secret seed stored in a k8s secret, instead of relying on a KMS key. The JWT secret seed lives in a k8s secret in the controller namespace, and rotates automatically with a cronjob running on that same namespace. It follows the pattern of the authmiddleware JWT seed rotation implemented in #320 , but decouples from authmiddleware stack by adding an independent secret, cronjob and secrets watches.

### Next steps
1. create verification API
2. use verification API in authmiddleware
3. remove kms config from router helm chart
4. configure plugin sidecar for operator
5. migrate operator chart to use plugin sidecar

### User experience
- no change for workspace end-user, the code path to use the k8s-native JWT is controlled by the access strategy
- for admins, if `AccessStrategy.spec.CreateConnectionHandler` is set to `k8s-native`, then whenever a user calls `Create:Connection` for such a workspace, the `extensionapi` uses the k8s secret instead of AWS to sign the token

### Implementation
- add k8s secret `jupyter-k8s-extensionapi-secrets` to operator namespace
- add cronjob to operator namespace, w/ related RBAC
- add RBAC role/role-binding for controller pod
- update helm chart/kustomize
- update `makefile` methods to pass the new parameters
- migrate `deploy-kind` to use `helm` instead of `kustomize` for deployment
- update `extensionapi` logic to support several signing strategy (not just AWS)
- add secrets watch to `extensionapi` to receive updates

### Testing
- add E2E test coverage for `Create:Connection` that leverage the `k8s-native`signing strategy
- ran "FOCUS=Extension API" E2E tests
- deployed to local kind cluster and verified secret rotation
- deployed manually to AWS --> no issue

### DIFF
--- a/internal/extensionapi/server.go
+++ b/internal/extensionapi/server.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/jupyter-infra/jupyter-k8s/internal/aws"
 	"github.com/jupyter-infra/jupyter-k8s/internal/jwt"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	genericapiserver "k8s.io/apiserver/pkg/server"
@@ -24,6 +25,7 @@ import (
 	"k8s.io/apiserver/pkg/util/compatibility"
 	"k8s.io/client-go/kubernetes"
 	v1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	toolscache "k8s.io/client-go/tools/cache"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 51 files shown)


joshuatowner (commented):
overall looks good, very similar to middleware usage. lets discuss de-duplication if you feel it's not posssible.

internal/extensionapi/server.go
```diff
@@ -275,12 +311,46 @@ func SetupExtensionAPIServerWithManager(mgr ctrl.Manager, config *ExtensionConfi
...
+	signerFactory, err := createJWTCompositeSignerFactory(config)
 	if err != nil {
 		return err
 	}
 
+	// When using k8s-native signing, register secret watch and initial key loader
+	if config.JwtSecretName != "" {
+		composite := signerFactory.(*jwt.CompositeSignerFactory)
+		nativeFactory, _ := composite.GetFactory("k8s-native")
+		stdFactory := nativeFactory.(*jwt.StandardSignerFactory)
```
andrii-i: Nit: bare type assertions. Might not be worth following up now though as this will be removed in the planned de-duplication refactoring.

andrii-i (approved):
Thank you @JGuinegagne. Looks good.

Agreeing with Joshua on the deduplication point + please find a nit suggestion below.

joshuatowner (approved):

PR #334, opened 2026-03-10
fix: preserved annotation/label system prefix 

earaghbidikashani:
### Problem

Users could set arbitrary labels and annotations with the workspace.jupyter.org/ prefix on Workspace resources. This prefix is reserved for system-managed metadata (e.g., created-by, template-name), and allowing users to set arbitrary keys under it could cause confusion, conflicts with future system keys, or security issues (e.g., spoofing ownership).

Introduce a centralized registry of system-managed metadata keys with mutation policies, and enforce the reserved prefix at the webhook level.

Each system-managed key has a policy:
- `SetOnCreateOnly` — system sets it once at creation, immutable after (e.g., created-by)
- `SetAlways` — system may update it on every operation (e.g., last-updated-by, template labels, access strategy labels)

The registry lives in internal/controller/constants.go as SystemManagedMetadataKeys. When someone adds a new workspace.jupyter.org/ key to the operator, they must add it to this map — otherwise the webhook will reject it for non-admin users.

Admin users and the controller service account bypass these checks entirely.

### Behavior

On create (non-admin user):
- Any workspace.jupyter.org/ label or annotation not in the system-managed registry → rejected
- System-managed keys are allowed because the mutating webhook sets them before validation runs

On update (non-admin user):
- Adding an unknown workspace.jupyter.org/ key → rejected
- Changing a SetOnCreateOnly key (e.g., created-by) → rejected as immutable
- Removing a SetOnCreateOnly key → rejected
- Changing or removing a SetAlways key → allowed (system will overwrite)

On WorkspaceTemplate baseLabels:
- A CEL validation rule on the CRD rejects any baseLabels entry with the workspace.jupyter.org/ prefix, preventing templates from injecting reserved keys into workspaces

This replaces the previous hardcoded created-by immutability check with a generalized mechanism that covers all reserved-prefix keys.

### DIFF
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -122,6 +122,7 @@ const (
 var SystemManagedMetadataKeys = map[string]MetadataKeyPolicy{
 	AnnotationCreatedBy:             SetOnCreateOnly,
 	AnnotationLastUpdatedBy:         SetAlways,
+	PreemptionReasonAnnotation:      SetAlways,
 	LabelWorkspaceTemplate:          SetAlways,
 	LabelWorkspaceTemplateNamespace: SetAlways,
 	LabelAccessStrategyName:         SetAlways,
--- a/internal/webhook/v1alpha1/reserved_prefix_validator_test.go
+++ b/internal/webhook/v1alpha1/reserved_prefix_validator_test.go
@@ -180,5 +180,17 @@ var _ = Describe("Reserved Prefix Validator", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(Equal("annotation 'workspace.jupyter.org/created-by' is immutable"))
 		})
+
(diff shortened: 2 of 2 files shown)


test/e2e/webhook_reserved_prefix_test.go
```diff
@@ -0,0 +1,122 @@
...
+			)
+			_, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), "patch with non-reserved label should succeed")
+
+			By("verifying the label was applied")
+			value, err := kubectlGet("workspace", "valid-workspace", namespace,
+				"{.metadata.labels.environment}")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(value).To(Equal("staging"))
+		})
```
JGuinegagne: could you add a +ve and -ve case for annotations as well?

Issue #337, opened 2026-03-13
Controller-Plugin Interface


JGuinegagne:
The core operator is the HTTP client; the plugin is the HTTP server, running as a sidecar
container on `localhost:<port>`.

## 1. Interface Shape

Errors are returned as HTTP status codes with `{"error": "<message>"}`.

### JWT Signing (`/v1alpha1/jwt/`)

Routed by `AccessStrategy.Spec.CreateConnectionHandler` (e.g. `"aws"`).

#### `POST /v1alpha1/jwt/sign`

```json
// Request
{
  "user": "string",
  "groups": ["string"],
  "uid": "string",
  "extra": {"key": ["value"]},
  "path": "/workspaces/ns/ws",
  "domain": "example.com",
  "tokenType": "bootstrap",
  "signingType": "kms",
  "connectionContext": {"kmsKeyId": "arn:..."}
}

// Response 200
{ "token": "<signed-jwt-string>" }
```

`tokenType` is the token lifecycle stage (`"bootstrap"` or `"session"`) — determines which
endpoints accept the token. `signingType` is an optional hint about the signing mechanism
(e.g. `"kms"`, `"k8s-native"`); the plugin already knows how it signs, so this is metadata
for logging/auditing rather than a routing decision.

`connectionContext` is the opaque `AccessStrategy.Spec.CreateConnectionContext` map.

#### `POST /v1alpha1/jwt/verify`

```json
// Request
{ "token": "<signed-jwt-string>" }

// Response 200
{
  "claims": {
    "sub": "user",
    "groups": ["group1"],
    "uid": "uid",
    "extra": {"key": ["value"]},
    "path": "/workspaces/ns/ws",
    "domain": "example.com",
    "tokenType": "bootstrap",
    "exp": 1234567890
  }
}

// Response 401
{ "error": "invalid or expired token" }
```

---

### Remote Access (`/v1alpha1/remote-access/`)

Routed by `AccessStrategy.Spec.PodEventsHandler` for pod lifecycle operations, and by
`AccessStrategy.Spec.CreateConnectionHandler` for session creation.

#### `POST /v1alpha1/remote-access/initialize`

One-time idempotent provider resource creation, called at startup or sidecar readiness.

```json
// Request: {}
// Response 200: {}
```

#### `POST /v1alpha1/remote-access/register-node-agent`

Register a node agent in a workspace pod for remote access. Called when a pod becomes running.

```json
// Request
{
  "podUID": "string",
[...]

PR #338, opened 2026-03-17
feat: controller plugin interfaces

### LINKED ISSUE #339: Harden E2E: "External volumes" container-not-found timeout
**Ref:** [CI run (PR #338)](https://github.com/jupyter-infra/jupyter-k8s/actions/runs/23208145048/job/67449261354?pr=338)

## What happened

`Workspace Storage > External volumes > should mount external PVCs specified in volumes field`
timed out after 30s at `workspace_storage_test.go:268`.

The `touch` on the first external PVC (`/home/jovyan/data`) succeeded, but every subsequent
`kubectl exec … ls /home/jovyan/shared/test-file-1.txt` failed with
`container not found ("workspace")` — the container vanished between the two
`VerifyPodCanAccessExternalVolumes` calls (likely OCI/cgroup restart under CI memory
pressure). 5 additional specs were skipped due to `Ordered` cascade.

## Proposed fixes

1. **Increase `Eventually` timeout** in `VerifyPodCanAccessExternalVolumes`
   (`helpers_volume.go:168,183`) from 30s to 60-90s with a shorter poll interval (2s
   instead of 5s), giving the container time to recover from a restart.

2. **Log pod events and container status on failure** — add a `DeferCleanup` that runs
   `kubectl describe pod` and `kubectl get events` when a spec fails, so future flakes
   are diagnosable from CI logs alone.

3. **Log container restart count before exec** — emit
   `kubectl get pod <name> -o jsonpath='{.status.containerStatuses[*].restartCount}'`
   so we can correlate exec failures with restarts.

JGuinegagne:
This PR creates the base contract between the controller/extensionapi container and their cloud-provider plugins. In the P0 architecture, a cloud provider plugin runs as a sidecar within the pod of the controller/extensionapi. However this is extensible to separate deployments outside of the pod network.

Refer to #337

### Implementation
- define plugin interface in `api/plugin`
- add a generic client in `internal/pluginclient` to be used by the extensionapi / controller
- add a generic server in `internal/pluginserver` to be used by plugins
- basic implementation `pluginserver` as a `net/http` server

### Testing
- not wired up yet, rely on unit tests only

### DIFF
--- a/api/plugin/v1alpha1/error.go
+++ b/api/plugin/v1alpha1/error.go
@@ -1,0 +1,11 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package v1alpha1
+
+// ErrorResponse is the standard error response body returned by plugin endpoints.
+type ErrorResponse struct {
+	Error string `json:"error"`
+}
--- a/api/plugin/v1alpha1/jwt_types.go
+++ b/api/plugin/v1alpha1/jwt_types.go
@@ -17,6 +17,7 @@ type SignRequest struct {
 	Path              string              `json:"path"`
 	Domain            string              `json:"domain"`
 	TokenType         string              `json:"tokenType"`
+	SigningType       string              `json:"signingType,omitempty"`
 	ConnectionContext map[string]string   `json:"connectionContext,omitempty"`
 }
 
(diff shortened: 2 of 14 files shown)


andrii-i (commented):
Structure makes sense and is inline with #337. Please find two suggestions on code optimization below.

andrii-i (approved):

bhanderig (approved):
LGTM

PR #345, opened 2026-03-20
feat: plugin client-server scaffolding

JGuinegagne:
This PR implements the scaffolding of the plugin client and plugin server, without concrete method implementations yet.
- the extensionapi / controller will have n-instances of plugin clients, one for each plugin declared
- each plugin will consist in a sidecar running alongside the extensionapi/controller in its pod, and it will implement the plugin server
- communication between a plugin client and server will go over HTTP in the pod local network (localhost)

This PR just implements the basic HTTP server configuration, along with standard functionality such as JSON parsing/encoding, logging, error handling and connection retries.

### Implementation
- client logic in `internal/pluginclient` package
- server logic in `internal/pluginserver` package
- define shared headers to trace back the origin request (either an `extensionapi` call, a reconciliation loop interation or a pod event), and the current request (in case a single request makes several plugin calls)
- note: the plugin client retries only connection errors to the plugin; it delegates retry of all other AWS SDK operations to the plugins. Rationale: we want to avoid retry storms.

### Testing
- not wired up yet, rely on unit tests only

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3,7 +3,7 @@
 ## Project Overview
 
 Jupyter-k8s is a Kubernetes operator for Jupyter notebooks and other IDEs. It manages compute, storage, networking, and access control for multiple users in a secure, scalable, usable and flexible way.
-- the project is not live yet, do not worry about backward compatibility.
+- the project is live, be mindful of backward compatibility.
 
 ### Kubernetes Custom Resources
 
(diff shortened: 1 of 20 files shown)


henrywa2 (dismissed):

henrywa2 (dismissed):

andrii-i (dismissed):
LGTM, spec-conformant with #337, good test coverage. Please find two suggestions below for your consideration.

henrywa2 (dismissed):

andrii-i (approved):
Looks good, thank you @JGuinegagne.

henrywa2 (approved):

PR #354, opened 2026-04-06
feat: create conn handler map in access strategy

JGuinegagne:
First of 4 PRs to migrate all AWS calls to a sidecar running alongside the controller (a "plugin").

This first PR is very limited in scope, and modifies the shape of the CRD / plugin APIs:
1. introduces a new attribute to `WorkspaceAccessStrategy.spec`: `CreateConnectionHandlerMap`
2. add new attribute `ConnectionType` to the plugin API  `CreateSessionRequest`

### Future PRs
- **PR 2: Plugin infrastructure** — New packages (`internal/plugin/`, `internal/pluginadapters/`, `internal/awsadapter/`), `pluginclient`/`pluginserver` refactors (slog→logr, interface consolidation), `cmd/aws-plugin/main.go` + Dockerfile. All additive — nothing in the controller or extensionapi imports the new code yet.
- **PR 3 [BREAKING]: Controller + extension API wiring** — Connect plugin infra: `cmd/main.go` (plugin-endpoints flag), `internal/controller/` (pod_event_handler adapter dispatch), `internal/extensionapi/` (multi-plugin connection), `internal/authmiddleware/` cleanup, delete `internal/aws/`.
- **PR 4: Helm, charts, e2e** — Plugin sidecar Helm configuration, Makefile targets, guided-chart updates, sample workspaces, and e2e test fixtures.

In order to merge PR 2, 3, 4, we may want to create a feature branch.

### Testing
- w/ combined code changes of PR 1-4, I tested E2E against a hyperpod cluster: jupyterlab (web access), code editor (web access + remote access). All worked as expected.

PR #355, opened 2026-04-06
feat: plugin wiring

JGuinegagne:
PR 2a of plugin migration, which focuses on `plugin` core business logic:
1. creates shared interface in `internal/plugin/`
    1.1. `apis.go` --> API groups (remote access, jwt)
    1.2. `context_entry.go` --> utilities to retrieve the resolved context map from AccessStrategy [attribute](https://github.com/jupyter-infra/jupyter-k8s/blob/a73f6a9bf1a2af4d3cd611f88f5fb338a54199eb/api/v1alpha1/workspaceaccessstrategy_types.go#L118)
2. wire up APIs in `internal/pluginclient`
    2.1. use consistent `logr`
    2.2. pass `pluginRequestID` (each request), `originRequestID` (optional, if say passed from `extensionapi`)
    2.3. remove `jwt` client routes (not used)
3. wire up APIs in `internal/pluginserver`
    3.1. use consistent `logr`
    3.2. create default `NotImplemented` route handlers for `jwt` and `remote_connection`
    3.3. export port in `config`, read from env var
4.  concrete implementation in `internal/awsplugin/`
    4.1. add expected keys of `AccessStrategy.spec.ConnectionContext` / `.PodEventContext`, w/ default values
    4.2. add simple `server` that instantiates a `PluginServer` w/ an `AwsSsmRemoteAccessRoutes` implementation
    4.3. add `NewAWSRemoteAccessRoutes` struct in `serverroutes_remote_access` module
    4.4. port over (duplicate for now) `ssm_client` from `internal/aws`
5. aws-plugin image: `cmd/aws-plugin/main.go`
    5.1. add `main.go` as simple `ListenAndServe()` http server, w/ controller-provided logger
    5.2. add `Dockerfile` based on go1.24 distroless image

### Future PRs
- **PR 2b: Adapters** — Controller-side adapter layer (`internal/pluginadapters/`, `internal/awsadapter/`). Not wired into the controller yet.
- **PR 3: Controller + extension API wiring [BREAKING]** — Connect plugin infra: `cmd/main.go` (plugin-endpoints flag), `internal/controller/` (pod_event_handler adapter dispatch), `internal/extensionapi/` (multi-plugin connection), `internal/authmiddleware/` cleanup, delete `internal/aws/`.
- **PR 4: Helm, charts, e2e [RESTORING]** — Plugin sidecar Helm configuration, Makefile targets, guided-chart updates, sample workspaces, and e2e test fixtures.

### Testing
see #354


internal/awsplugin/initializer.go
```diff
@@ -0,0 +1,58 @@
...
+		ssmClient: ssmClient,
+	}, nil
+}
+
+// EnsureResourcesInitialized ensures SSH document is created (only once)
+func (r *ResourceInitializer) EnsureResourcesInitialized(ctx context.Context) error {
+	r.initOnce.Do(func() {
+		err := r.ssmClient.createSageMakerSpaceSSMDocument(ctx)
+		if err != nil {
+			r.initError = fmt.Errorf("failed to create SSH document: %w", err)
```
henrywa2, line 38: If this returns an error will it run again?

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// RegisterNodeAgent creates an SSM activation for a workspace pod so the SSM agent
+// sidecar can register as a managed instance.
+func (h *AWSRemoteAccessRoutes) RegisterNodeAgent(ctx context.Context, req *pluginapi.RegisterNodeAgentRequest) (*pluginapi.RegisterNodeAgentResponse, error) {
+	logger := log.FromContext(ctx).WithName("remote-access-handler")
+	pctx := req.PodEventsContext
+
+	// Extract IAM role from podEventsContext
+	iamRole := pctx[SSMManagedNodeRoleKey]
```
henrywa2, line 53: How will the controller get this IAM role to pass into the plugin? Would it make more sense to have this role live somewhere in the plugin itself, like an env variable? Similar comments for the other fields from `pctx`.

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// Resolve overridable constants from podEventsContext
+	managedByKey := SageMakerManagedByTagKey.ResolveStr(pctx)
+	managedByVal := SageMakerManagedByTagValue.ResolveStr(pctx)
+	clusterTagKey := SageMakerEKSClusterTagKey.ResolveStr(pctx)
+	instancePrefix := SSMInstanceNamePrefix.ResolveStr(pctx)
+
+	// Build tags for the SSM activation
+	clusterARN := os.Getenv("CLUSTER_ID")
```
henrywa2, line 65: This is being passed from both `pctx` and env?

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+// RegisterNodeAgent creates an SSM activation for a workspace pod so the SSM agent
+// sidecar can register as a managed instance.
+func (h *AWSRemoteAccessRoutes) RegisterNodeAgent(ctx context.Context, req *pluginapi.RegisterNodeAgentRequest) (*pluginapi.RegisterNodeAgentResponse, error) {
+	logger := log.FromContext(ctx).WithName("remote-access-handler")
+	pctx := req.PodEventsContext
+
+	// Extract IAM role from podEventsContext
+	iamRole := pctx[SSMManagedNodeRoleKey]
+	if iamRole == "" {
+		return nil, fmt.Errorf("%s is required in podEventsContext", SSMManagedNodeRoleKey)
```
henrywa2, line 55: Why only IAM role is required?

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+	clusterARN := os.Getenv("CLUSTER_ID")
+	tags := map[string]string{
+		"workspace-pod-uid":   req.PodUID,
+		"workspace-name":      req.WorkspaceName,
+		"workspace-namespace": req.Namespace,
+		managedByKey:          managedByVal,
+		clusterTagKey:         clusterARN,
+	}
+
+	instanceName := fmt.Sprintf("%s-%s", instancePrefix, req.PodUID)
```
henrywa2, line 74: Would it make more sense to use something more readable like `{workspaceName}-{workspaceNamespace}`?

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+	logger.Info("Deregistering node agent", "podUID", req.PodUID)
+
+	// DeregisterNodeAgentRequest has no context map, use defaults
+	podUIDTagKey := WorkspacePodUIDTagKey.Default
+	instancePrefix := SSMInstanceNamePrefix.Default
+
+	// Clean up managed instances
+	if err := h.ssmClient.CleanupManagedInstancesByPodUID(ctx, req.PodUID, podUIDTagKey); err != nil {
+		logger.Error(err, "Failed to cleanup managed instances", "podUID", req.PodUID)
+		// Continue to activation cleanup even if instance cleanup fails
```
henrywa2, line 106: Do we do anything to clean up instances if this fails?

internal/awsplugin/serverroutes_remote_access.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Find the managed instance by pod UID
+	instanceID, err := h.ssmClient.FindInstanceByPodUID(ctx, req.PodUID, podUIDTagKey)
+	if err != nil {
+		logger.Error(err, "Failed to find managed instance", "podUID", req.PodUID)
+		return nil, fmt.Errorf("failed to find managed instance for pod %s: %w", req.PodUID, err)
+	}
+
+	// Get SSM document name from connection context
+	documentName := cctx["ssmDocumentName"]
```
henrywa2, line 135: Shouldn't the aws plugin already have the documentName since it was created during initialization

henrywa2 (approved):

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: That gets passed from the `AccessStrategy.spec.PodEventContext`

I would keep this there rather than in the plugin, but we can discuss.

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: this one from env. I agree that it's a bit inconsistent w/ the IAM role. The thing w/ cluster-id is that it is unambiguously 1:1 w/ the cluster, as in w/ the plugin, whereas IAM roles could be 1:n

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: I think because we only need that value here.

What else should be required?

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: that change is just refactored from existing [code](https://github.com/jupyter-infra/jupyter-k8s/blob/a73f6a9bf1a2af4d3cd611f88f5fb338a54199eb/internal/aws/ssm_remote_access_strategy.go#L437)

Generally, this whole code should be revisited, but I would keep it as a separate concern

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: Same comment - this whole logic ought to be revisited, but this is just a refactor.

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: same argument, an SSM Document does not clearly map 1:1 to a cluster IMO

internal/awsplugin/initializer.go
JGuinegagne: It should, but that's beyond the scope of this PR/project which just refactors the AWS-specific code.

See [existing code](https://github.com/jupyter-infra/jupyter-k8s/blob/a73f6a9bf1a2af4d3cd611f88f5fb338a54199eb/internal/aws/initializer.go)

internal/awsplugin/serverroutes_remote_access.go
henrywa2: That makes sense but we also have logic in the plugin to create exactly one ssm document. But agree there could be more.

internal/awsplugin/serverroutes_remote_access.go
henrywa2: Don't know if anything else is needed. I don't see the use case where the customer would want to override something like `SageMakerEKSClusterTagKey`, but I guess it's not strictly required either.

internal/awsplugin/serverroutes_remote_access.go
henrywa2: What about `SageMakerEKSClusterTagKey`?

internal/awsplugin/serverroutes_remote_access.go
henrywa2: Makes sense to keep the possibility open for multiple IAM roles

internal/awsplugin/serverroutes_remote_access.go
JGuinegagne: thanks for the pointer, pure k8s-logic will live in the `internal/awsadapter` package; this package will remain in the operator, but WILL NOT make any aws API calls (just k8s native calls)

joshuatowner (approved):

PR #356, opened 2026-04-07
feat: plugin adapters

JGuinegagne:
PR 2b of the plugin migration story, which focuses on "adapters". An adapter is a package that handles specific business logic, including k8s operations, and delegate to a plugin to make cloud-provider calls, such as AWS API calls.

The idea is to temporarily keep in the controller the podEvent orchestration logic for SSM activation / registration etc.
We should revisit and simplify this story, and move most of this logic to the sidecar, however it would require new `extensionapi` APIs to retrieve the activation code. In between, we could use a build flag to ignore such packages; the key point is that `adapters` _must not_ add additional dependencies to the controller, ie would not bloat to `go mod` with dependencies like `github.com/aws/aws-sdk-go-v2` (+ all the transitive deps).

### Implementation
1. `internal/pluginadapters` package
    1.1. shared interface `PodEventPluginAdapter` with `HandlePodRunning` and `HandlePodDeleted` methods
    1.2. shared utilities to parse and resolve context from `AccessStrategy.spec.<>Context` ([example](https://github.com/jupyter-infra/jupyter-k8s/blob/cfc59d613ec2672562c75d1649c6d0e72df97b35/api/v1alpha1/workspaceaccessstrategy_types.go#L122))
    1.3. pod exec interface `PodExecInterface`, with shared pod/container utilities
2. `internal/awsadapter` package for an aws-specific adapter
    2.1. `NewAwsSsmPodEventAdapter` concrete implementation of `PodEventPluginAdapter` interface
    2.2. aws-specific context entries w/ defaults
3. minor changes to `internal/controller` to use the refactored `PodEventPluginAdapter` interface
4. remove `aws` handling in `internal/jwt`: codepath no longer used since #347

### Future PRs
- **PR 3: Controller + extension API wiring [BREAKING]** — Connect plugin infra: `cmd/main.go` (plugin-endpoints flag), `internal/controller/` (pod_event_handler adapter dispatch), `internal/extensionapi/` (multi-plugin connection), `internal/authmiddleware/` cleanup, delete `internal/aws/`.
- **PR 4: Helm, charts, e2e [RESTORING]** — Plugin sidecar Helm configuration, Makefile targets, guided-chart updates, sample workspaces, and e2e test fixtures.

### Testing
See #354

### DIFF
--- a/internal/awsadapter/context_entries.go
+++ b/internal/awsadapter/context_entries.go
@@ -1,0 +1,35 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+// Package awsadapter provides the SSM-based pod event adapter implementation.
+// It reads all configuration from the podEventsContext map defined in the AccessStrategy.
+package awsadapter
+
+import "github.com/jupyter-infra/jupyter-k8s/internal/plugin"
+
+// Pod event context keys resolved from the AccessStrategy's podEventsContext map.
+// These keys have no defaults — values must be provided in the AccessStrategy.
+var (
+	// ContextKeyPodUID is the context key for the pod UID (resolved from controller::PodUid()).
+	ContextKeyPodUID = plugin.ContextEntry{Key: "podUid"}
+	// ContextKeySsmManagedNodeRole is the context key for the IAM role used for SSM activation.
+	ContextKeySsmManagedNodeRole = plugin.ContextEntry{Key: "ssmManagedNodeRole"}
+	// ContextKeySidecarContainerName is the context key for the SSM agent sidecar container name.
+	ContextKeySidecarContainerName = plugin.ContextEntry{Key: "sidecarContainerName"}
+	// ContextKeyWorkspaceContainerName is the context key for the main workspace container name.
+	ContextKeyWorkspaceContainerName = plugin.ContextEntry{Key: "workspaceContainerName"}
+	// ContextKeyRegistrationStateFile is the context key for the state file path in the shared volume.
+	ContextKeyRegistrationStateFile = plugin.ContextEntry{Key: "registrationStateFile"}
+	// ContextKeyRegistrationMarkerFile is the context key for the legacy marker file path.
+	ContextKeyRegistrationMarkerFile = plugin.ContextEntry{Key: "registrationMarkerFile"}
+	// ContextKeyRegistrationScript is the context key for the SSM registration script path.
+	ContextKeyRegistrationScript = plugin.ContextEntry{Key: "registrationScript"}
+	// ContextKeyRemoteAccessServerScript is the context key for the remote access server startup script.
+	ContextKeyRemoteAccessServerScript = plugin.ContextEntry{Key: "remoteAccessServerScript"}
+	// ContextKeyRemoteAccessServerPort is the context key for the remote access server port.
+	ContextKeyRemoteAccessServerPort = plugin.ContextEntry{Key: "remoteAccessServerPort"}
+	// ContextKeyRegion is the context key for the AWS region used in SSM registration.
+	ContextKeyRegion = plugin.ContextEntry{Key: "region"}
+)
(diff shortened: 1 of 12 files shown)


internal/awsadapter/ssm_pod_event_adapter.go
```diff
@@ -0,0 +1,339 @@
...
+//   - (nil, nil) if state file doesn't exist (first time setup)
+//   - (nil, error) if state file exists but is corrupted
+func (s *AwsSsmPodEventAdapter) readRegistrationState(ctx context.Context, pod *corev1.Pod, sidecarContainer, stateFile string) (*RegistrationState, error) {
+	logger := logf.FromContext(ctx).WithValues("pod", pod.Name, "namespace", pod.Namespace)
+
+	cmd := []string{"cat", stateFile}
+	output, err := s.podExecUtil.ExecInPod(ctx, pod, sidecarContainer, cmd, "")
+	if err != nil {
+		// File doesn't exist - this is first time setup
+		logger.V(1).Info("State file not found, treating as first registration")
```
aws-jasakshi, line 215: if err is not nil, file not found is always the right reason ?
What if there are some other errors while running state file in sidecar container ?

internal/awsadapter/ssm_pod_event_adapter.go
```diff
@@ -0,0 +1,339 @@
...
+		"workspace", currentWorkspaceRestarts)
+
+	stateFile := ContextKeyRegistrationStateFile.ResolveStr(podEventsContext)
+
+	// Read state from shared volume
+	state, err := s.readRegistrationState(ctx, pod, sidecarContainer, stateFile)
+
+	// Check if setup is already in progress
+	if state != nil && state.SetupInProgress {
+		logger.V(1).Info("Setup already in progress by another event, skipping")
```
aws-jasakshi, line 103: nit - Should we add staleness check here ? what if controller crashes and this status is never updated

JGuinegagne: same comment - this logic needs a lot of work, but I'd like to keep it out of scope of this refactoring.

henrywa2 (approved):

internal/awsadapter/context_entries.go
```diff
@@ -0,0 +1,35 @@
...
+	ContextKeyPodUID = plugin.ContextEntry{Key: "podUid"}
+	// ContextKeySsmManagedNodeRole is the context key for the IAM role used for SSM activation.
+	ContextKeySsmManagedNodeRole = plugin.ContextEntry{Key: "ssmManagedNodeRole"}
+	// ContextKeySidecarContainerName is the context key for the SSM agent sidecar container name.
+	ContextKeySidecarContainerName = plugin.ContextEntry{Key: "sidecarContainerName"}
+	// ContextKeyWorkspaceContainerName is the context key for the main workspace container name.
+	ContextKeyWorkspaceContainerName = plugin.ContextEntry{Key: "workspaceContainerName"}
+	// ContextKeyRegistrationStateFile is the context key for the state file path in the shared volume.
+	ContextKeyRegistrationStateFile = plugin.ContextEntry{Key: "registrationStateFile"}
+	// ContextKeyRegistrationMarkerFile is the context key for the legacy marker file path.
```
henrywa2, line 25: If it's legacy, can it be removed during this migration?

JGuinegagne: Sure, let me see if it's still used:
- if no, will remove in the next PR
- if yes, I will update the comment

internal/awsadapter/ssm_pod_event_adapter.go
JGuinegagne: While I agree that this logic should be revisited/hardened, however this PR focuses on decoupling/rearchitecting. See existing code [here](https://github.com/jupyter-infra/jupyter-k8s/blob/cfc59d613ec2672562c75d1649c6d0e72df97b35/internal/aws/ssm_remote_access_strategy.go#L295)

aws-jasakshi (approved):

PR #358, opened 2026-04-10
feat: hpchart plugin migration

JGuinegagne:
This PR 4 of the plugin migration story, which provides configuration for the `aws-hyperpod` guided chart, along with `Makefile` methods to deploy, create sample workspaces, and generate bearer token URLs for both jupyterlab and code editor IDEs against an `aws-hyperpod` deployment.

### Implementation
1. update `dist/chart/template` after generation to add the plugin configuration
    1.1. expose a `controller: plugin: []` in `values.yaml`
3. update `hack` to handle the plugins
4. update `Makefile` methods:
    3.1. create the ecr repo for the aws plugin image if missing
    3.2. build & publish the aws plugin to ecr
    3.3. parametrize `deploy-aws-internal` method to pass the `plugin` config to helm

### Testing
- deployed the updated operator, including `controller`, `extensionapi` and `aws-plugin` to an hyperpod cluster
- used `make apply-sample-hyperpod WS_USER=admin` to create
    - public `jupyterlab` workspace
    - private `jupyterlab` workspace
    - public `code-editor` workspace
    - private `code-editor` workspace
- used `make bearer-token WS_NAME=<>` to generate URL, and verify access on webbrowser
- used `make vscode-token WS_NAME=<>` to generate VS Code URL, and verify access on my desktop
- switched users (ie used different iam role), then used `make apply-sample-hyperpod WS_USER=parkeruser1`
- verified access (webbrowser and VS Code remote access) on own workspaces
- verified access denied on `make bearer-token WS_NAME=<>` and `make vscode-token WS_NAME=<>` for private workspaces owned by `admin` user
- verified access (webbrowser and VS Code remote access) on admin public workspaces

### DIFF
--- a/dist/chart/templates/manager/manager.yaml
+++ b/dist/chart/templates/manager/manager.yaml
@@ -126,6 +126,18 @@ spec:
           ports:
             - containerPort: {{ .port }}
               protocol: TCP
+          {{- if .healthcheckCommand }}
+          livenessProbe:
+            exec:
+              command: {{ .healthcheckCommand | toJson }}
+            initialDelaySeconds: 5
+            periodSeconds: 10
+          readinessProbe:
+            exec:
+              command: {{ .healthcheckCommand | toJson }}
+            initialDelaySeconds: 2
+            periodSeconds: 5
+          {{- end }}
           {{- if .env }}
           env:
             {{- range $key, $value := .env }}
--- a/dist/chart/templates/manager/manager.yaml
+++ b/dist/chart/templates/manager/manager.yaml
@@ -61,6 +61,9 @@ spec:
             {{- if .Values.workspacePodWatching.enable }}
             - "--enable-workspace-pod-watching"
             {{- end}}
+            {{- if .Values.controller.plugins }}
+            - "--plugin-endpoints={{ range $i, $p := .Values.controller.plugins }}{{ if $i }},{{ end }}{{ $p.name }}=http://localhost:{{ $p.port }}{{ end }}"
+            {{- end}}
           command:
             - /manager
           image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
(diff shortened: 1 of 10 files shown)


guided-charts/aws-hyperpod/values.yaml
```diff
@@ -3,6 +3,10 @@
 
 namespace: jupyter-k8s-system
 
+# AWS configuration
+aws:
+  region: ""
```
bhanderig, line 8: nit: If someone deploys without setting the region, the podEventsContext will have region:  "" which could cause subtle runtime failures. Can we add a `required` check here or someplace else is making sure region is always populated?

guided-charts/aws-hyperpod/templates/hyperpod-access-strategy.yaml
```diff
@@ -61,12 +61,28 @@ spec:
...
+    podUid: "extensionapi::PodUid()"
     ssmDocumentName: "{{ .Values.remoteAccess.ssmDocumentName }}"
+    port: "2222"
+  podEventsContext:
+    podUid: "controller::PodUid()"
+    region: "{{ .Values.aws.region }}"
+    ssmManagedNodeRole: "{{ .Values.remoteAccess.ssmManagedNodeRole }}"
+    sidecarContainerName: "ssm-agent-sidecar"
+    workspaceContainerName: "workspace"
+    registrationStateFile: "/opt/amazon/sagemaker/workspace/.ssm-registration-state.json"
```
bhanderig: nit: These seems to be duplicate values that also appear in the sidecar container spec below (e.g.,
   the readinessProbe checks /tmp/ssm-registered). If these ever diverge, it'll be a hard-to-debug issue. Could we extract these to values.yaml or at least adding a comment noting the coupling

dist/chart/templates/manager/manager.yaml
```diff
@@ -114,6 +117,27 @@ spec:
               readOnly: true
             {{- end }}
           {{- end }}
+        {{- range .Values.controller.plugins }}
```
bhanderig, line 120: Nit: Plugin sidecar containers have no health probes. Since this is the first sidecar added to the controller pod, consider adding at least a liveness probe (e.g. TCP on {{ .port }}) so the pod restarts if the plugin crashes, rather than silently losing plugin functionality.

bhanderig (approved):

henrywa2 (approved):

PR #360, opened 2026-04-14
feat: add default secondary storage

earaghbidikashani:
## Add `defaultVolumes` to WorkspaceTemplate

   Adds a new `defaultVolumes` field to `WorkspaceTemplateSpec` that lets platform admins define default secondary volumes (e.g., shared FSx datasets) for workspaces using a template.

   Follows the existing "replace if empty" defaulting pattern: if a workspace specifies no volumes, it inherits the template's `defaultVolumes`. If the workspace defines its own volumes, template defaults are not applied.

   ### Changes

   - New `defaultVolumes` field on `WorkspaceTemplateSpec` (reuses existing `VolumeSpec` type, max 10 items)
   - New `volume_defaulter.go` webhook defaulter, registered after storage defaults
   - Unit tests for the defaulter
   - E2E tests covering inheritance and override behavior
   - PVC cleanup added to template test teardown

### DIFF
--- a/dist/chart/templates/crd/workspace.jupyter.org_workspacetemplates.yaml
+++ b/dist/chart/templates/crd/workspace.jupyter.org_workspacetemplates.yaml
@@ -2058,6 +2058,34 @@ spec:
                       type: string
                   type: object
                 type: array
+              defaultVolumes:
+                description: |-
+                  DefaultVolumes specifies default additional volumes for workspaces using this template
+                  Volumes are applied during defaulting only if the workspace does not specify any volumes
+                  Each volume references a pre-existing PVC by name in the workspace's namespace
+                items:
+                  description: VolumeSpec defines a volume to mount from an existing
+                    PVC
+                  properties:
+                    mountPath:
+                      description: MountPath is the path where the volume should be
+                        mounted (Unix-style path, e.g. /data)
+                      type: string
+                    name:
+                      description: Name is a unique identifier for this volume within
+                        the pod (maps to pod.spec.volumes[].name)
+                      type: string
+                    persistentVolumeClaimName:
+                      description: PersistentVolumeClaimName is the name of the existing
+                        PVC to mount
+                      type: string
+                  required:
+                  - mountPath
+                  - name
+                  - persistentVolumeClaimName
+                  type: object
+                maxItems: 10
+                type: array
               description:
                 description: Description provides additional information about this
                   template
--- a/internal/webhook/v1alpha1/volume_defaulter.go
+++ b/internal/webhook/v1alpha1/volume_defaulter.go
@@ -1,0 +1,20 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package v1alpha1
+
+import (
+	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+)
+
+// applyVolumeDefaults applies default volumes from template to workspace
+func applyVolumeDefaults(workspace *workspacev1alpha1.Workspace, template *workspacev1alpha1.WorkspaceTemplate) {
+	if len(workspace.Spec.Volumes) > 0 || len(template.Spec.DefaultVolumes) == 0 {
+		return
+	}
+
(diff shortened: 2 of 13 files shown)


test/e2e/workspace_template_test.go
```diff
@@ -907,6 +907,101 @@ var _ = Describe("Workspace Template", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(jupyterLabValue).To(Equal("yes"))
 		})
+
+		It("should inherit default volumes from template", func() {
```
JGuinegagne, line 913: non-blocking: At a 1000 lines agent/LLM start having trouble reading files, we should consider breaking down this file. Could you create a GH issue to track it please?

JGuinegagne (dismissed):
Few high-level questions:
1. what happens if a template specifies an additional volume, and two workspaces reference the template?
2. do we need to verify that a workspace can read/write to the additional volume?
3. same question for a volume shared by 2 workspaces (defaulted from template)

If this is covered already by volume test, could you add an inline comment explaining that?

test/e2e/workspace_template_test.go
earaghbidikashani: sure: #362

earaghbidikashani:
>what happens if a template specifies an additional volume, and two workspaces reference the template?

Both workspaces get the same defaultVolumes copied into their spec.volumes at admission time by the mutating webhook.

> do we need to verify that a workspace can read/write to the additional volume?

This behavior is unchanged, the operator doesn't verify read/write access for secondary volumes today. If there is permission missing, the operator will get errors from the CSI drivers and show them in the logs. 

> same question for a volume shared by 2 workspaces (defaulted from template)

If you're asking about the operator, the behavior is same as the secondary storage. If you're asking about the e2e tests, I can add a test scenario that a template is with PVCs is used by 2 workspaces.

dist/chart/templates/crd/workspace.jupyter.org_workspacetemplates.yaml
```diff
@@ -2058,6 +2058,34 @@ spec:
                       type: string
                   type: object
                 type: array
+              defaultVolumes:
+                description: |-
+                  DefaultVolumes specifies default additional volumes for workspaces using this template
```
henrywa2, line 2063: nit: add some punctuation at the end of the line

internal/webhook/v1alpha1/volume_defaulter.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// applyVolumeDefaults applies default volumes from template to workspace
+func applyVolumeDefaults(workspace *workspacev1alpha1.Workspace, template *workspacev1alpha1.WorkspaceTemplate) {
+	if len(workspace.Spec.Volumes) > 0 || len(template.Spec.DefaultVolumes) == 0 {
+		return
+	}
+
+	workspace.Spec.Volumes = make([]workspacev1alpha1.VolumeSpec, len(template.Spec.DefaultVolumes))
+	copy(workspace.Spec.Volumes, template.Spec.DefaultVolumes)
```
henrywa2, line 19: Any reason we need a copy here? Would there be an issue to just do `workspace.Spec.Volumes = template.Spec.DefaultVolumes`?

henrywa2 (dismissed):

JGuinegagne (dismissed):

henrywa2 (approved):

internal/webhook/v1alpha1/volume_defaulter.go
earaghbidikashani: It's a defensive copy so mutating the workspace's volumes can't accidentally modify the template's. Same pattern as other defaulters in the codebase (e.g., DeepCopy() in lifecycle defaults).

dist/chart/templates/crd/workspace.jupyter.org_workspacetemplates.yaml
earaghbidikashani: similar pattern to existing descriptions

JGuinegagne (approved):

PR #365, opened 2026-04-20
feat: init container support

earaghbidikashani:
## Add Init Containers support to Workspace and WorkspaceTemplate CRDs

   ### Summary

   Adds the ability for users to specify init containers on workspaces, with template-level defaults and an allow/deny toggle for admins.

   ### Changes

   - **`Workspace.spec.initContainers`** — New field (`[]corev1.Container`, max 10) allowing users to define init containers that run before the workspace container starts.
   - **`WorkspaceTemplate.spec.defaultInitContainers`** — Admins can define default init containers inherited by workspaces that don't specify their own.
   - **`WorkspaceTemplate.spec.allowInitContainers`** — Boolean toggle (defaults `true`). When `false`, workspaces referencing the template are rejected if they specify init containers.
   - **Deployment builder** — Workspace init containers are set on the pod spec before access strategy init containers are appended, so the final order is: workspace → access strategy.

   ### Files changed

   | Area | Files |
   |------|-------|
   | API types | `workspace_types.go`, `workspacetemplate_types.go`, `zz_generated.deepcopy.go` |
   | Webhook defaulting | `initcontainer_defaulter.go` (new), `template_defaulter.go` |
   | Webhook validation | `initcontainer_validator.go` (new), `template_validator.go`, `validation_types.go` |
   | Controller | `deployment_builder.go` |
   | Unit tests | `initcontainer_defaulter_test.go`, `initcontainer_validator_test.go`, `deployment_builder_test.go` |
   | E2E tests | `workspace_initcontainer_test.go` (new), static fixtures in `test/e2e/static/initcontainer-base/` |
   | Generated | CRD manifests, Helm chart |

   ### Testing

   - Unit tests cover defaulting (inherit from template, no-op when workspace specifies own), validation (allow/deny/nil), and deployment builder (init containers appear in pod spec).
   - E2E tests cover: workspace with init containers reaches Available, template default inheritance, and webhook rejection when `allowInitContainers: false`.


api/v1alpha1/workspace_types.go
```diff
@@ -187,6 +187,12 @@ type WorkspaceSpec struct {
 	// Overrides template defaults when specified
 	// +optional
 	ContainerSecurityContext *corev1.SecurityContext `json:"containerSecurityContext,omitempty"`
+
+	// InitContainers specifies init containers to run before the workspace container starts
+	// When a template is used, template's DefaultInitContainers are applied if workspace has none
+	// +kubebuilder:validation:MaxItems=10
```
henrywa2, line 192: How was this decided?

henrywa2 (dismissed):

api/v1alpha1/workspacetemplate_types.go
```diff
@@ -145,6 +145,18 @@ type WorkspaceTemplateSpec struct {
 	// +optional
 	DefaultContainerSecurityContext *corev1.SecurityContext `json:"defaultContainerSecurityContext,omitempty"`
 
+	// DefaultInitContainers specifies default init containers for workspaces using this template
```
joshuatowner, line 148: will we need to support for human: setting an init container at a template level and not allowing user to override? not sure if they'll be happy w/ potentially any container/img being added at workspace level.

in general- i think restricting init containers may need a bit more thought (resources, images, etc). lets make sure by default we aren't bypassing the console's ability to set resource limits for instance.

joshuatowner (changes requested):
- plz validate it works OK with the built in init containers with SSM

internal/webhook/v1alpha1/initcontainer_validator.go
```diff
@@ -0,0 +1,33 @@
...
+	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+)
+
+// validateInitContainers checks if init containers are allowed by template
+func validateInitContainers(initContainers []corev1.Container, template *workspacev1alpha1.WorkspaceTemplate) *TemplateViolation {
+	if len(initContainers) == 0 {
+		return nil
+	}
+
+	if template.Spec.AllowInitContainers != nil && !*template.Spec.AllowInitContainers {
```
henrywa2: Is "allow by default" the behavior we have for custom images as well?

api/v1alpha1/workspace_types.go
earaghbidikashani: there is no k8 limit, but I wanted to set a similar limit to secondary volume just to have some limit, we can raise if there is need later

api/v1alpha1/workspacetemplate_types.go
earaghbidikashani: the implementation for this follows the same pattern as secondary volume. 
For the scenario you mentioned: if a template sets `defaultInitContainers` + `allowInitContainers: false`, the user can't add their own, but the template defaults still apply.

Here is the overall behavior: 
   | Template setting | Workspace specifies init containers | Result |
   |---|---|---|
   | `allowInitContainers: true` (or nil/default) | Yes | Accepted — user's init containers are used |
   | `allowInitContainers: true` (or nil/default) | No | Template's `defaultInitContainers` are inherited |
   | `allowInitContainers: false` | Yes | Rejected by webhook |
   | `allowInitContainers: false` | No | Rejected by webhook if there is any defaults in template|

I think this approach applies to humain and overall customer use cases. Is there any scenario that you think this approach is lacking in?

internal/webhook/v1alpha1/initcontainer_validator.go
earaghbidikashani: images is set to false as default but secondary volume is set to true by default.

earaghbidikashani:
> plz validate it works OK with the built in init containers with SSM

@joshuatowner Pods support multiple init containers natively. This implementation sets workspace-level init containers first, then the access strategy init containers (including the SSM one) are appended after in.
We verified in kind and e2e tests. We will also test this e2e in addon release process

internal/webhook/v1alpha1/initcontainer_validator.go
henrywa2: What is the reason we decided to make it different?

api/v1alpha1/workspacetemplate_types.go
JGuinegagne: hum, to stay closer to the precedent of [AllowCustomImages](https://github.com/jupyter-infra/jupyter-k8s/blob/5f0b508b46151c2ceaa20527caf4800e941a8669/api/v1alpha1/workspacetemplate_types.go#L44), how about `allowCustomInitContainers`?

We could then default to `false` (consistent w/ `allowCustomImages`), and if set to `false`, reject anything that does not match the template exactly?

JGuinegagne (changes requested):
Consider using `allowCustomInitContainers` in template for consistency w/ `allowCustomImages`

internal/webhook/v1alpha1/initcontainer_validator.go
earaghbidikashani: discussed offline the default behavior

api/v1alpha1/workspacetemplate_types.go
earaghbidikashani: agreed, will implement `allowCustomInitContainers`

internal/webhook/v1alpha1/initcontainer_validator.go
```diff
@@ -13,21 +13,39 @@ import (
...
+	if template.Spec.AllowCustomInitContainers != nil && *template.Spec.AllowCustomInitContainers {
+		return nil
+	}
+
+	// Check if the init containers are exactly the template defaults (set by defaulter)
+	if len(initContainers) == len(template.Spec.DefaultInitContainers) {
+		allMatch := true
+		for i := range initContainers {
+			if initContainers[i].Name != template.Spec.DefaultInitContainers[i].Name ||
+				initContainers[i].Image != template.Spec.DefaultInitContainers[i].Image {
```
JGuinegagne: hum, only check name and image? Shouldn't we run a deep check here?

Also, does the order of init containers matter? otherwise shouldn't we match by container name, and then run a deep check?

JGuinegagne (changes requested):
We should run deeper checks than just name + image.

Also, consider whether init containers order matter, as in is the following valid?
- workspace.spec.initContainers = [a, b, c]
- template.spec.defaultInitContainers = [b, a, c]

internal/webhook/v1alpha1/initcontainer_validator.go
earaghbidikashani: Sounds good, I'll do a deep equal comparison which checks every field. 

Matching is done by container name so order is not enforced during validation. The runtime order is whatever is in the workspace spec (which the defaulter sets in template order).

henrywa2: from docs:
> If you specify multiple init containers for a Pod, kubelet runs each init container sequentially

so if workspace spec is [a, b, c] and template is [b, a, c], that should count as different and workspace should override.  Is that the case for your chage?

api/v1alpha1/workspacetemplate_types.go
henrywa2: > if set to false, reject anything that does not match the template exactly?

This might be too restrictive? With `allowCustomImages` false, users can still change their container with diff command/args/env variables. With `allowCustomInitContainers` false, users will not be able to change anything?

internal/webhook/v1alpha1/initcontainer_validator_test.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			template.Spec.AllowCustomInitContainers = nil
+			initContainers := []corev1.Container{
+				{Name: "template-init", Image: "evil:latest", Command: []string{"echo"}},
+			}
+			violation := validateInitContainers(initContainers, template)
+			Expect(violation).NotTo(BeNil())
+		})
+
+		It("should allow template defaults in different order", func() {
```
JGuinegagne: let's align on that one. That's essentially saying that init container compose operator is commutative; as in `InitContainerA o InitContainerB (state) == InitContainerB o InitContainerA (state)`.

internal/webhook/v1alpha1/initcontainer_validator.go
JGuinegagne: +1, if we consider that init container mutates the "state" of the container, allowing to change order means that we consider this state mutation is commutative. That doesn't seem right

earaghbidikashani: Sounds good, I'll do a deep equal comparison which checks every field by index so order matters

internal/webhook/v1alpha1/initcontainer_validator_test.go
earaghbidikashani: Sure following up to compare order as well

api/v1alpha1/workspacetemplate_types.go
earaghbidikashani: Yes I think that should be the case
The main container has additional guardrails (resourceBounds, allowedImages, envRequirements) that constrain what users can do even when they customize it. 

Init containers don't have those layers, so a full match is the only way to prevent abuse without building an equivalent validation framework for init containers.

JGuinegagne (approved):

henrywa2 (approved):

PR #367, opened 2026-04-23
feat: access readiness probes

### LINKED ISSUE #363: Bad Gateway on workspace open immediately after creation
# Issue: "Bad Gateway" on workspace open immediately after creation

## Problem

When a user opens a workspace URL right after creation ("Create and Open" workflow),
they see a **502 Bad Gateway**. The workspace becomes accessible after a few seconds.

There is a propagation delay between creating a Traefik IngressRoute in etcd and
Traefik actually serving the route. The controller marks the workspace `Available`
as soon as the IngressRoute object exists, without verifying the route is live.

Existing TODO: `state_machine.go:259` — `// TODO: add probe and requeue`

## Proposed Solution: `accessProbe`

Add an optional `accessProbe` field to `WorkspaceAccessStrategySpec`. The controller
performs the probe after creating access resources and before setting `Available=True`.
Strategies without a probe keep current behavior.

Timing fields (`initialDelaySeconds`, `periodSeconds`, `timeoutSeconds`,
`failureThreshold`) follow the naming and semantics of
[`corev1.Probe`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#probe-v1-core)
(used by `Pod.spec.containers[].readinessProbe`).

### API Change

```go
type WorkspaceAccessStrategySpec struct {
    // ... existing fields ...

    // AccessProbe defines how the controller verifies that access resources are
    // serving traffic. If not set, access resources are considered ready as soon
    // as they exist in the API server.
    // +optional
    AccessProbe *AccessProbe `json:"accessProbe,omitempty"`
}

type AccessProbe struct {
    // HTTPGet specifies an HTTP GET to perform against the access path.
    HTTPGet *AccessHTTPGetProbe `json:"httpGet,omitempty"`

    // Number of seconds after the Workspace would otherwise be marked as Available before accesss probes are initiated. Default: 0.
    // +optional
    InitialDelaySeconds int32 `json:"initialDelaySeconds,omitempty"`
    // How often (in seconds) to perform the probe. Default to 2 seconds. Minimum value is 1.
    // +optional
    PeriodSeconds int32 `json:"periodSeconds,omitempty"`
    // Number of seconds after which the probe times out. Defaults to 5 second. Minimum value is 1.
    // +optional
    TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
    // Minimum consecutive failures before giving up and marking the workspace as Degraded.
    // The controller stops probing after this threshold; a new reconciliation (e.g. from a
    // Workspace or AccessStrategy update) restarts the probe from scratch.
    // Unlike corev1.Probe, there is no successThreshold — the probe is a one-time gate
    // that passes on the first successful response. Defaults to 30. Minimum value is 1.
    // Default rationale: with periodSeconds=2, 30 failures = 60 seconds. This covers
    // most ingress controller disruptions (pod restart, reschedule to another node with
    // image pull) while failing fast enough to surface real misconfigurations.
    // +optional
    FailureThreshold int32 `json:"failureThreshold,omitempty"`
}

type AccessHTTPGetProbe struct {
    // URLTemplate is a Go template resolving to the URL to probe.
    // Variables: .Workspace, .AccessStrategy, .Service
    URLTemplate string `json:"urlTemplate"`

    // FailureStatusCodes that indicate the route is not yet ready. Default: [502, 503].
    // +optional
    FailureStatusCodes []int `json:"failureStatusCodes,omitempty"`
}
```

### Reconciliation Change

```
1. Deployment ready? Service ready? ──(both yes)──►
2. Create access resources in etcd (EnsureAccessResourcesExist)
3. If accessProbe is defined:
   a. HTTP GET the resolved URL
   b. If response ∈ failureStatusCodes:
      - If under failureThreshold → Progressing=True, reason=AccessNotReady, requeue after PeriodSeconds
      - If failureThreshold exceeded → Progressing=False, Degraded=True, Available=False. Stop probing.
        A subsequent reconciliation (Workspace or AccessStrategy update) restarts the probe.
   c. If success → continue
4. Set Available=True
```

### Sample Usage

```yaml
kind: WorkspaceAccessStrategy
metadata:
  name: oauth-access-strategy
spec:
  accessProbe:
    httpGet:
      urlTemplate: "https://$DOMAIN/workspaces/{{ .Workspace.Namespace }}/{{ .Workspace.Name }}/"
      failureStatusCodes: [502, 503]
    periodSeconds: 2
    failureThreshold: 30
  accessResourceTemplates:
    # ... IngressRoute templates ...
```

## Alternatives Considered

**Client-side loading page** — auth middleware returns a loading page on 502 with
auto-refresh. Better UX but doesn't fix status accuracy; only works through auth
middleware.

**Fixed delay** — hardcoded requeue delay after creating access resources. Too short
fails, too long wastes time. Not configurable per environment.

**Watch IngressRoute status** — poll `.status` on the IngressRoute. Traefik's CRD
doesn't expose a reliable readiness condition, and this is Traefik-specific.

JGuinegagne:
This PR adds support to probe route readiness in `AccessStrategy` instead of marking the `Workspace` available immediately after the `accessResources` have been created in ETCD.

Closes: https://github.com/jupyter-infra/jupyter-k8s/issues/363

### Implementation
1. add `AccessStrategy.spec.AccessStartupProbe` attribute
2. add access probe related attribute `Workspace.status`
    2.1. `AccessStartupProbeSucceeded` boolean
    2.2. `AccessStartupProbeFailures` int
    2.3. `LastAccessStartupProbeTime` time
3. update controller logic to handle access startup probe

### Testing
1. add unit tests to new controller logic
2. add unit tests to state machine using envtest framework (controller coverage 52% -> 69%)
3. add E2E tests for access strategy (directly against the workspace service)
4. deployed to controller w/ an `aws-oidc` chart deployment, tested E2E
5. deployed the controller w/ an `aws-hyperpod` chart deployment, tested E2E

For 4/ and 5/, refer to https://github.com/jupyter-infra/jupyter-k8s-aws/pull/8


internal/controller/access_startup_prober.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/http"
+	"time"
+
+	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// AccessStartupProberInterface allows mocking in tests
+type AccessStartupProberInterface interface {
```
henrywa2, line 20: `AccessStartupProbe` or `AccessStartupProber`?

internal/controller/access_startup_prober.go
```diff
@@ -0,0 +1,123 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return false, fmt.Errorf("failed to create probe request: %w", err)
+	}
+
+	resp, err := p.client.Do(req)
+	if err != nil {
+		logger.V(1).Info("Access startup probe connection failed", "url", url, "error", err)
+		return false, nil
```
henrywa2, line 91: Do we want to return an error here?

readiness-probes-plan.md
```diff
@@ -0,0 +1,259 @@
...
+
+### 4. Access startup prober — `internal/controller/access_startup_prober.go` (new)
+
+`AccessStartupProber` struct with `net/http` client. Two methods:
+- `ResolveProbeURL` — resolves `urlTemplate` via the existing template engine
+- `Probe` — single HTTP GET, returns `(ready bool, err error)`. Success = status 200–399
+  or in `additionalSuccessStatusCodes`.
+
+HTTP client config: timeout from `timeoutSeconds`, redirect following disabled (so 302 is
+observed directly as a 3xx success, not followed), TLS `InsecureSkipVerify: true` for
```
henrywa2: Did this make it in?

internal/controller/access_startup_prober.go
JGuinegagne: nice catch, let me fix this and add a unit test

readiness-probes-plan.md
JGuinegagne: oops, let me uncommit the agent planning file and confirm that `InsecureSkipVerify` was removed :)

internal/controller/access_startup_prober.go
JGuinegagne: `AccessStartupProber` is correct, look at the interface, it has a `probe` methods (as in a verb)

internal/controller/access_startup_prober.go
JGuinegagne: Actually no this is correct, if we fail to send the HTTP request, it should be recorded as a probe failure and return `false, nil`. Added an inline comment as I agree this is confusing for the reader.

internal/controller/state_machine_access.go
```diff
@@ -70,3 +77,131 @@ func (sm *StateMachine) ReconcileAccessForDesiredStoppedStatus(ctx context.Conte
...
+	failureThreshold := resolveFailureThreshold(probe)
+	periodSeconds := resolvePeriodSeconds(probe)
+
+	// Already exceeded threshold on a previous reconcile — stay Degraded
+	// without re-probing. The counter is cleared when the workspace stops
+	// or the access strategy is removed, which restarts the probe.
+	if workspace.Status.AccessStartupProbeFailures != nil &&
+		*workspace.Status.AccessStartupProbeFailures >= failureThreshold {
+		return &ProbeResult{Status: ProbeFailureThresholdExceeded}, nil
+	}
```
andrii-i, line 129: The `FailureThreshold` doc comment says updating the AccessStrategy restarts the probe:
https://github.com/jupyter-infra/jupyter-k8s/blob/a684f08f088f11b37cb132ac4ea7ffa128f42205/api/v1alpha1/workspaceaccessstrategy_types.go#L102-L103

But this early return means it doesn't: once threshold is exceeded the workspace stays `Degraded` regardless of what changed.

So fixing a bad `urlTemplate` or correcting a misconfigured domain in the AccessStrategy won't help as the workspace won't retry. Recovery requires stop+restart.

Probe state should reset here when `failures >= threshold`. Since `UpdatePermanentDegradedRunningStatus` stops requeuing, reaching this point while `Degraded` only happens on an external change, which is exactly when a retry makes sense.

andrii-i (dismissed):
Thank you @JGuinegagne, looks good. Please find one suggestion re probe reset on AccessStrategy update below.

internal/controller/state_machine_access.go
JGuinegagne: Thanks Andrii, you're right, once the probe fails it won't restart again on ws/as updates, and the only way is then to stop the workspace. I think this is okay to have to restart workspaces that failed the access probe.

The reconciliation loop will clear the status element tracking the failed probe in `state_machine.ReconcileAccessForDesiredStoppedStatus()`. I do agree though that CRD docstring is misleading, let me fix it to explain that the workspace needs to be restart.

There is another edge case that is quite subtle to address: what happens if the `AccessStrategy` gets updated and the workspaces that use it have passed the access probe? In some cases at least, the access probe ought to restart. The access strategy admission hook already has logic to trigger updates on Workspaces, we should use that to invalidate the probe status, but let's do that in a separate PR. I'll create a GH issue to track this.

andrii-i: Makes sense, let's do that, thank you.

henrywa2 (approved):

andrii-i (approved):

PR #371, opened 2026-04-27
feat: kiro and cursor support

earaghbidikashani:
Summary

  Extends the WorkspaceConnection API to support Kiro and Cursor as remote connection types, and opens up the *-remote pattern so future IDEs can be added without code changes.

  Motivation

  The extension API previously hardcoded vscode-remote as the only remote connection type. Adding any new IDE required changes to the connection type constants, validation logic, and handler dispatch. This change makes the remote connection path generic so that new IDEs only require
  a YAML change in the WorkspaceAccessStrategy's createConnectionHandlerMap.

  Changes

  api/connection/v1alpha1/workspace_connection_types.go

  - Added ConnectionTypeKiroRemote (kiro-remote) and ConnectionTypeCursorRemote (cursor-remote) constants

  internal/extensionapi/serverroute_connection.go

  - Validation now accepts web-ui, known remote types (vscode-remote, kiro-remote, cursor-remote), and any string matching the *-remote pattern. Bare -remote and non-matching strings are rejected.
  - Handler dispatch collapsed from a per-type switch to two paths: web-ui (bearer token) vs everything else (plugin handler map lookup). This removes the hardcoded vscode-remote case.
  - For unknown *-remote types, the response includes an HTTP Warning: 299 header indicating the type is not officially supported.
  - Response workspaceConnectionType now echoes back the actual requested type instead of hardcoding vscode-remote.
  - Added isRemoteConnectionType and isKnownRemoteType helpers.

  internal/extensionapi/serverroute_connection_test.go

  - Updated validation tests to cover kiro-remote, cursor-remote, unknown *-remote (e.g. windsurf-remote), and rejection of bare -remote
  - Added unit tests for isRemoteConnectionType and isKnownRemoteType

  Backward compatibility

  Fully backward compatible. Existing vscode-remote and web-ui requests work identically. No CRD schema changes — the connection type field was already an unvalidated string; the enum was only enforced in Go code.

  What this does NOT change

  - Plugin-side URL scheme resolution (the plugin already receives connectionType via CreateSessionRequest and can use it to pick the right deep-link URL)
  - CRD schema or generated manifests
  - WorkspaceAccessStrategy structure


api/connection/v1alpha1/workspace_connection_types.go
```diff
@@ -17,6 +17,10 @@ const (
 
 	// ConnectionTypeVSCodeRemote represents VSCode remote connection type
 	ConnectionTypeVSCodeRemote = "vscode-remote"
+	// ConnectionTypeKiroRemote represents Kiro remote connection type
```
henrywa2: I'd lean towards getting rid of these. The goal of making it `*-remote` is that we don't have to update this enum with every new IDE. If anything, we can include it somewhere in AWS docs.

edwardps: +1

edwardps (commented):
Please update the help content to describe how *-remote works. Customer is responsible to provide prefix corresponding to IDE which is controlled by IDE. We can also provide the famous example for vscode, cursor, kiro.

api/connection/v1alpha1/workspace_connection_types.go
earaghbidikashani: sounds good

internal/extensionapi/serverroute_connection.go
```diff
@@ -265,26 +265,23 @@ func (s *ExtensionServer) HandleConnectionCreate(w http.ResponseWriter, r *http.
...
+	default:
+		// All *-remote types go through the plugin handler map
 		pluginName, action, found := resolveConnectionHandler(accessStrategy, connectionType)
 		if !found {
 			WriteKubernetesError(w, http.StatusBadRequest, fmt.Sprintf("connection type %q is not configured for this workspace", connectionType))
 			return
 		}
 		if pluginName == handlerK8sNative {
-			WriteKubernetesError(w, http.StatusBadRequest, "k8s-native handler is not supported for vscode-remote connections")
+			WriteKubernetesError(w, http.StatusBadRequest, fmt.Sprintf("k8s-native handler is not supported for %q connections", connectionType))
```
henrywa2: nit: this could be a bit more clear that k8s-native is not supported for _any_ `*-remote` connections

internal/extensionapi/serverroute_connection.go
```diff
@@ -314,6 +311,11 @@ func (s *ExtensionServer) HandleConnectionCreate(w http.ResponseWriter, r *http.
 	}
 }
 
+// isRemoteConnectionType checks if the connection type follows the *-remote pattern.
+func isRemoteConnectionType(connectionType string) bool {
+	return strings.HasSuffix(connectionType, "-remote") && len(connectionType) > len("-remote")
```
henrywa2: nit: maybe regex match would be more clear here? I'm ok with this too as it's pretty simple. To me it's a bit easier to understand than `len(connectionType) > len("-remote")` but don't really care either way

henrywa2 (dismissed):

internal/extensionapi/serverroute_connection.go
earaghbidikashani: sure I can make that change

internal/extensionapi/serverroute_connection.go
earaghbidikashani: I tested this regex, also this shouldn't make a difference practically so I'll keep it but I can change it in the future if needed

henrywa2 (dismissed):

aakashmandavilli96 (approved):

internal/extensionapi/serverroute_connection.go
edwardps: nit: can we add validation to error out when special letters are included? we do the validation as much as possible. 

e.g. ^[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*-(remote|score)$

Matches
vscode-remote
my-vscode-remote
team-cursor-score
kiro-remote

Does NOT match
vscode-remote-ssh ❌
my--vscode-remote ❌ (double hyphen invalid)
-vscode-remote ❌

internal/extensionapi/serverroute_connection.go
```diff
@@ -324,12 +326,16 @@ func validateWorkspaceConnectionRequest(req *connectionv1alpha1.WorkspaceConnect
...
+
+	// Accept web-ui, known remote types, and any *-remote pattern
+	switch {
+	case connectionType == connectionv1alpha1.ConnectionTypeWebUI:
+		// valid
+	case isRemoteConnectionType(connectionType):
+		// valid — known or unknown remote types are accepted
 	default:
-		return fmt.Errorf("invalid workspaceConnectionType: '%s'. Valid types are: 'vscode-remote', 'web-ui'", req.Spec.WorkspaceConnectionType)
+		return fmt.Errorf("invalid workspaceConnectionType: '%s'. Must be 'web-ui' or follow the '{ide}-remote' pattern (e.g. 'vscode-remote')", connectionType)
```
edwardps: can we add kiro and cursor case as well?

edwardps (commented):
It's very close.

internal/extensionapi/serverroute_connection.go
earaghbidikashani: makes sense, will add

internal/extensionapi/serverroute_connection.go
earaghbidikashani: sure

internal/extensionapi/serverroute_connection.go
```diff
@@ -265,26 +266,23 @@ func (s *ExtensionServer) HandleConnectionCreate(w http.ResponseWriter, r *http.
...
+	default:
+		// All *-remote types go through the plugin handler map
 		pluginName, action, found := resolveConnectionHandler(accessStrategy, connectionType)
 		if !found {
 			WriteKubernetesError(w, http.StatusBadRequest, fmt.Sprintf("connection type %q is not configured for this workspace", connectionType))
 			return
 		}
 		if pluginName == handlerK8sNative {
-			WriteKubernetesError(w, http.StatusBadRequest, "k8s-native handler is not supported for vscode-remote connections")
+			WriteKubernetesError(w, http.StatusBadRequest, fmt.Sprintf("k8s-native handler is not supported for remote connections (requested %q)", connectionType))
```
JGuinegagne, line 277: non-blocking: generally reflecting user input in error messages is considered a security risk. The codebase already did it on line 286, so it's not a blocker, just for your awareness though.

JGuinegagne (approved):

edwardps (approved):

PR #374, opened 2026-04-28
fix: pass template defaults for primary storage

### LINKED ISSUE #372: Workspace primary storage mounted to an unexpected path
Hi,
I have a workspace configured to use both primary EBS storage and FSx storage, as follows:

```
  storage:
    mountPath: /home/jovyan
    size: 5Gi
    storageClassName: sagemaker-spaces-default-storage-class
  templateRef:
    name: jupyterlab-smd-slurm-custom
  volumes:
  - mountPath: /fsx
    name: secondary-storage-0
    persistentVolumeClaimName: fsx-claim
```

As you can see, the `mountPath` gets set, for some reason, to `/home/jovyan`, which I did not configure.

Indeed, the configuration in my template is:

```
  primaryStorage:
    defaultMountPath: /mnt/ebs-volume
    defaultSize: 5Gi
    defaultStorageClassName: sagemaker-spaces-default-storage-class
    maxSize: 100Gi
    minSize: 100Mi
```

Looks like the `defaultMountPath` configuration is not applied.

earaghbidikashani:
When a `WorkspaceTemplate` specifies `primaryStorage.defaultMountPath` (e.g., `/mnt/ebs-volume`), workspaces using that template always get `/home/jovyan` instead. The same issue affects `defaultSize`.

   ### Root Cause

   `StorageSpec.MountPath` and `StorageSpec.Size` had `+kubebuilder:default` markers in the CRD. CRD defaults are applied by the API server **before** the mutating webhook runs. By the time the storage defaulter checks `if workspace.Spec.Storage.MountPath == ""`, the field is
   already `/home/jovyan` — so the template's `defaultMountPath` is never applied.

   ### Fix

   Remove the `+kubebuilder:default` markers from `StorageSpec.Size` and `StorageSpec.MountPath`. The template defaulter and controller fallbacks already handle all scenarios.

   ### Behavior Matrix

   | # | Template? | Template has primaryStorage? | User specifies storage? | User sets mountPath/size? | Before (CRD defaults): After API server | After (no CRD defaults): Persisted on object |
   |---|---|---|---|---|---|---|
   | 1 | Yes | Yes | No | No | Storage is nil (no storage key submitted) | Template values persisted. Same behavior as before. |
   | 2 | Yes | Yes | Yes | Yes | User values kept as-is | User values persisted. Same behavior as before. |
   | 3 | Yes | Yes | Yes | No | CRD pre-fills /home/jovyan + 10Gi before webhook runs | Template values persisted. This is the bug fix — previously CRD defaults won over template. |
   | 4 | Yes | No primaryStorage | Yes | No | CRD pre-fills /home/jovyan + 10Gi before webhook runs | Empty fields on the persisted object. Works at runtime because the controller (pvc_builder.go) has hardcoded fallbacks to /home/jovyan + 10Gi. |
   | 5 | No | N/A | Yes | Yes | User values kept as-is | User values persisted. Same behavior as before. |
   | 6 | No | N/A | Yes | No | CRD pre-fills /home/jovyan + 10Gi before webhook runs | Same as row 4. Empty fields on the persisted object, controller fills them at runtime. |
   | 7 | No | N/A | No | No | Storage is nil | No storage created, no PVC. Same behavior as before. |

 fix #372

### DIFF
--- a/api/v1alpha1/workspace_types.go
+++ b/api/v1alpha1/workspace_types.go
@@ -44,12 +44,10 @@ type StorageSpec struct {
 	// Size specifies the size of the persistent volume
 	// Supports standard Kubernetes resource quantities (e.g., "10Gi", "500Mi", "1Ti")
 	// Integer values without units are interpreted as bytes
-	// Defaults to 10Gi (applied by webhook, not CRD default, to allow template overrides)
 	Size resource.Quantity `json:"size,omitempty"`
 
 	// MountPath specifies where to mount the persistent volume in the container
 	// Default is /home/jovyan (jovyan is the standard user in Jupyter images)
-	// Defaults applied by webhook, not CRD default, to allow template overrides
 	MountPath string `json:"mountPath,omitempty"`
 }
 
--- a/api/v1alpha1/workspace_types.go
+++ b/api/v1alpha1/workspace_types.go
@@ -44,12 +44,12 @@ type StorageSpec struct {
 	// Size specifies the size of the persistent volume
 	// Supports standard Kubernetes resource quantities (e.g., "10Gi", "500Mi", "1Ti")
 	// Integer values without units are interpreted as bytes
-	// +kubebuilder:default="10Gi"
+	// Defaults to 10Gi (applied by webhook, not CRD default, to allow template overrides)
 	Size resource.Quantity `json:"size,omitempty"`
 
 	// MountPath specifies where to mount the persistent volume in the container
 	// Default is /home/jovyan (jovyan is the standard user in Jupyter images)
-	// +kubebuilder:default="/home/jovyan"
+	// Defaults applied by webhook, not CRD default, to allow template overrides
 	MountPath string `json:"mountPath,omitempty"`
 }
 
--- a/config/crd/bases/workspace.jupyter.org_workspaces.yaml
+++ b/config/crd/bases/workspace.jupyter.org_workspaces.yaml
@@ -1992,7 +1992,6 @@ spec:
                     description: |-
                       MountPath specifies where to mount the persistent volume in the container
                       Default is /home/jovyan (jovyan is the standard user in Jupyter images)
-                      Defaults applied by webhook, not CRD default, to allow template overrides
                     type: string
                   size:
                     anyOf:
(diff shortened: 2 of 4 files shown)


joshuatowner (dismissed):

joshuatowner (commented):
approval contingent on henry low level check

api/v1alpha1/workspace_types.go
```diff
@@ -44,12 +44,12 @@ type StorageSpec struct {
...
 	// Supports standard Kubernetes resource quantities (e.g., "10Gi", "500Mi", "1Ti")
 	// Integer values without units are interpreted as bytes
-	// +kubebuilder:default="10Gi"
+	// Defaults to 10Gi (applied by webhook, not CRD default, to allow template overrides)
 	Size resource.Quantity `json:"size,omitempty"`
 
 	// MountPath specifies where to mount the persistent volume in the container
 	// Default is /home/jovyan (jovyan is the standard user in Jupyter images)
-	// +kubebuilder:default="/home/jovyan"
+	// Defaults applied by webhook, not CRD default, to allow template overrides
```
henrywa2: nit: is this line elsewhere in the CRD? IMO the CRD doesn't need to mention the webhook. Similar comment in other places

earaghbidikashani: makes sense will fix

henrywa2:
quick question, what happens if only one of mountpath/size is specified but not the other? Will the controller fill in the other one, or only if both are empty?

earaghbidikashani:
> quick question, what happens if only one of mountpath/size is specified but not the other? Will the controller fill in the other one, or only if both are empty?

the webhook applies defaults one by one for storage fields, same logic is applied for fallback logic

JGuinegagne (approved):
LGTM, but could you either a/ add E2E tests to assert this behavior or b/ cut a GH issue and track it by EOW?

earaghbidikashani:
> LGTM, but could you either a/ add E2E tests to assert this behavior or b/ cut a GH issue and track it by EOW?

sure: https://github.com/jupyter-infra/jupyter-k8s/issues/376

joshuatowner (approved):

PR #375, opened 2026-04-28
fix: dns cache issues

### LINKED ISSUE #369: Access startup probe defeated by DNS negative caching
## Problem

When a workspace is created, the access startup probe fires before the wildcard
DNS record has propagated to CoreDNS. CoreDNS caches the NXDOMAIN response, and
every subsequent probe retry resolves against the stale negative cache entry,
burning through all 30 attempts against a cached failure.

Observed on EKS with `*.parker.jggg.people.aws.dev` — `admin-code-editor-private`
failed 100% reproducibly while 3 sibling workspaces (created milliseconds earlier)
passed. Restarting CoreDNS to flush the cache fixed it immediately.

## Root cause

1. Probe fires ~1s after access resources are created (before DNS propagates)
2. CoreDNS returns NXDOMAIN and caches it (default negative TTL: 30s)
3. Probe retries every 2s — all subsequent lookups hit the cached NXDOMAIN
4. 30 failures x 2s = 60s, but the negative cache TTL may outlast the entire window
5. Workspace goes Degraded permanently

### 1. Separate DNS failure budget (recommended)

Detect `net.DNSError` in the probe response. Still count DNS failures, but use a
separate, higher threshold (e.g. 60) or longer period (e.g. 5s) so the probe
outlasts the negative cache TTL without inflating `failureThreshold` for HTTP errors.

### 2. Log DNS failures at info level

Currently the actual error is logged at `V(1)` (debug). DNS-specific failures
should log at info with a distinct message like
`"Access startup probe: DNS not yet resolved"` so operators can diagnose without
bumping log verbosity.

### 3. Back off on consecutive DNS failures

After N consecutive DNS failures, increase the probe interval (e.g. 2s -> 5s -> 10s)
to avoid reinforcing the negative cache with rapid retries.

### 4. Document CoreDNS negative cache TTL

Note in the guided chart docs that CoreDNS negative cache TTL should be kept low
(5-10s) for clusters using access startup probes with dynamic DNS.

## Reproduction

1. Deploy controller + hyperpod chart on EKS
2. Create 4 sample workspaces simultaneously
3. One or more will hit NXDOMAIN on first probe and fail all 30 retries
4. `kubectl rollout restart deployment/coredns -n kube-system` + recreate fixes it

JGuinegagne:
This PR adds exponential backoff to the `AccessStartupProbe` defined in `AccessStrategy`. By default, the last 10 probes use exponential backoff, using the `periodSeconds` of the `accessStrategy.spec.accessStartupProbe`.

Closes: #369 

Related PR: https://github.com/jupyter-infra/jupyter-k8s-aws/pull/13

### Challenge
- CoreDNS caches DNS resolution in the cluster, which can lead to apparent probe failures
- CoreDNS config is cluster-specific, the operator has no control over it
- if we hit the same cache record too frequently, we risk extending the cache TTL
- this leads to conflicting constraints: on the one hand we want to try fast enough to allow workspace to get to `Available` as fast as possible, on the other hand backoff enough on DNS failure to allow the cached record to invalidate

### Implementation
1. log DNS failure at info level (instead of debug) for traceability
2. change `workspace.status.lastAccessStartupProbeTime` > `workspace.status.earliestNextProbeTime` to support exponential backoff
3. add logic to increase the retry time exponentially for the last 10 probes
4. update default `failureThreshold` to 20, leading to 10 fast polls followed by exponential backoff

### Testing
- unit tests
- E2E tests
- see https://github.com/jupyter-infra/jupyter-k8s-aws/pull/13

### DIFF
--- a/api/v1alpha1/workspace_types.go
+++ b/api/v1alpha1/workspace_types.go
@@ -246,11 +246,11 @@ type WorkspaceStatus struct {
 	// +optional
 	AccessStartupProbeFailures *int32 `json:"accessStartupProbeFailures,omitempty"`
 
-	// LastAccessStartupProbeTime records when the last access startup probe
-	// was executed. Used to enforce minimum spacing (periodSeconds) between
-	// probes across reconcile cycles.
+	// EarliestNextProbeTime is the earliest wall-clock time at which the next
+	// access startup probe may fire. Set by the controller after each probe
+	// attempt to enforce spacing; survives watch-triggered re-reconciliations.
 	// +optional
-	LastAccessStartupProbeTime *metav1.Time `json:"lastAccessStartupProbeTime,omitempty"`
+	EarliestNextProbeTime *metav1.Time `json:"earliestNextProbeTime,omitempty"`
 
 	// Conditions represent the current state of the Workspace resource.
 	// Each condition has a unique type and reflects the status of a specific aspect of the resource.
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -86,7 +86,16 @@ const (
 	// DefaultAccessStartupProbeTimeoutSeconds is the default probe timeout in seconds
 	DefaultAccessStartupProbeTimeoutSeconds = 5
 	// DefaultAccessStartupProbeFailureThreshold is the default max consecutive failures
-	DefaultAccessStartupProbeFailureThreshold = 30
+	DefaultAccessStartupProbeFailureThreshold = 20
+
+	// ProbeBackoffThreshold is the number of retries (counting from the end)
+	// that use exponential backoff. The first max(0, failureThreshold - this)
+	// retries use the configured periodSeconds; the last 10 use backoff.
+	ProbeBackoffThreshold = 10
+	// ProbeBackoffMaxRetrySeconds caps the exponential backoff interval.
+	// 302s comfortably exceeds any standard CoreDNS negative cache TTL
+	// (default 30s, SOA-driven up to 300s).
+	ProbeBackoffMaxRetrySeconds = 302
 
 	// MinimalRequeueDelay is the delay for near-immediate requeue
 	MinimalRequeueDelay = 10 * time.Millisecond
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -765,8 +765,8 @@ func (in *WorkspaceStatus) DeepCopyInto(out *WorkspaceStatus) {
 		*out = new(int32)
 		**out = **in
 	}
-	if in.LastAccessStartupProbeTime != nil {
-		in, out := &in.LastAccessStartupProbeTime, &out.LastAccessStartupProbeTime
+	if in.EarliestNextProbeTime != nil {
+		in, out := &in.EarliestNextProbeTime, &out.EarliestNextProbeTime
 		*out = (*in).DeepCopy()
 	}
 	if in.Conditions != nil {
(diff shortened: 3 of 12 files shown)


api/v1alpha1/workspace_types.go
```diff
@@ -246,11 +246,11 @@ type WorkspaceStatus struct {
...
 
-	// LastAccessStartupProbeTime records when the last access startup probe
-	// was executed. Used to enforce minimum spacing (periodSeconds) between
-	// probes across reconcile cycles.
+	// EarliestNextProbeTime is the earliest wall-clock time at which the next
+	// access startup probe may fire. Set by the controller after each probe
+	// attempt to enforce spacing; survives watch-triggered re-reconciliations.
 	// +optional
-	LastAccessStartupProbeTime *metav1.Time `json:"lastAccessStartupProbeTime,omitempty"`
+	EarliestNextProbeTime *metav1.Time `json:"earliestNextProbeTime,omitempty"`
```
henrywa2, line 251: Bit concerned about using wall clock here. Have you considered cases like daylight savings, desync between machines etc and made sure there are no edge cases that could cause issues?

internal/controller/constants.go
```diff
@@ -86,7 +86,16 @@ const (
...
+	DefaultAccessStartupProbeFailureThreshold = 20
+
+	// ProbeBackoffThreshold is the number of retries (counting from the end)
+	// that use exponential backoff. The first max(0, failureThreshold - this)
+	// retries use the configured periodSeconds; the last 10 use backoff.
+	ProbeBackoffThreshold = 10
+	// ProbeBackoffMaxRetrySeconds caps the exponential backoff interval.
+	// 302s comfortably exceeds any standard CoreDNS negative cache TTL
+	// (default 30s, SOA-driven up to 300s).
+	ProbeBackoffMaxRetrySeconds = 302
```
henrywa2, line 98: seems like 5 mins is a long time to wait for a connection to be ready

api/v1alpha1/workspace_types.go
JGuinegagne: That's an absolute time, and it's compiled `metav1.NewTime(time.Now().Add(delay))`. Not sure how `go`/`k8s` handle time exactly, but such operations should be resilient to date idiosyncrasies. Let me research though.

internal/controller/constants.go
JGuinegagne: Yeah I agree. We can reduce it to 120s or even 60s if you d like, it's just that we don't have control over the CoreDNS cache TTL setting.

api/v1alpha1/workspace_types.go
JGuinegagne: According to Claude, it's wall clock time-independent (UTC), the only wrinkle is NTP (internal node clock), but that is too marginal a difference to matter.

- UTC, not local time — Go and Kubernetes both use UTC internally, so DST is irrelevant. Henry's DST concern doesn't apply.                  
- NTP adjustments are tiny — modern NTP uses slew (gradual rate adjustment), not step. Typical drift is milliseconds. Even a rare NTP step would be a few seconds at most.                                                                                                              
- Worst case is benign — clock jumps forward → one probe fires early, hits DNS cache, wastes one retry. Clock jumps backward → probe is delayed by the jump amount. Neither is catastrophic; the backoff just sets a new deadline on the next attempt.
- Probe spacing is best-effort — it's an optimization to avoid hammering a stale DNS cache, not a correctness invariant. Off-by-seconds doesn't matter when we're dealing with 30-300s cache TTLs.

henrywa2 (approved):

PR #377, opened 2026-04-29
chore: ci release

JGuinegagne:
This PR adds a release github workflow which publishes 4 packages to `ghcr` of `jupyter-infra`: 
1. operator image to `jupyper-k8s`
2. authmiddleware image to `jupyter-k8s-authmiddleware`
3. rotator image to `jupyter-k8s-rotator` 
4. operator chart to `charts/jupyter-k8s`

### Implementation
1. add workflow triggered jobs:
    1.1. `release-stage-images.yml` > build & push each docker image to `staging/` in `ghcr` of `jupyter-infra`
    1.2. `release-stage-chart.yml` > build & push chart to `staging/charts/jupyter-k8s` (using `oci://`)
    1.3. `release-e2e.yml` > start a `kind` cluster, install the staged chart (1.2) which pins (1.1) images, run the E2E tests
    1.4. `promote-images.yml` > copy img from `staging/` to `` in `ghcr`
    1.5. `promote-chart.yml` > copy chart from `staging/charts/<chart>` to `charts/<chart>` in `ghcr` (using `oci://`)
2. add `release.yml` workflow in `.github/release.yml` which takes version to release (semver) as arg
3. add `Makefile` method for running E2E tests against staging
4. add `.github/AGENT.md` to explain how to test the CI
5. add pointer to `.github/AGENT.md` in root `CLAUDE.md`

### Testing
- pushed to another branch of the repo, run each workflow separately
- ran a release workflow, see [results](https://github.com/jupyter-infra/jupyter-k8s/actions/runs/25119405189)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,153 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version (e.g. v0.2.0)'
+        required: true
+      dry_run:
+        description: 'Push to staging + run e2e, skip promotion and release'
+        type: boolean
+        default: false
+
+env:
+  CHART_NAME: jupyter-k8s
+
+jobs:
+  validate:
+    name: Validate
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.validate.outputs.version }}
+      chart_version: ${{ steps.validate.outputs.chart_version }}
+      short_sha: ${{ steps.validate.outputs.short_sha }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Validate inputs
+        id: validate
+        run: |
+          VERSION="${{ inputs.version }}"
+
+          # Validate semver format
+          if ! echo "$VERSION" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$'; then
+            echo "::error::Invalid version format: $VERSION (expected vX.Y.Z or vX.Y.Z-prerelease)"
+            exit 1
+          fi
+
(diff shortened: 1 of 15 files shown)


henrywa2 (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+          fi
+
+          # Chart version strips the v prefix
+          CHART_VERSION="${VERSION#v}"
+
+          # Check tag doesn't already exist
+          if git rev-parse "refs/tags/$VERSION" >/dev/null 2>&1; then
+            echo "::error::Tag $VERSION already exists"
+            exit 1
+          fi
```
andrii-i, line 46: `actions/checkout@v4` defaults to a shallow clone without tags (https://github.com/actions/checkout#usage). Since no remote tags are present locally, `git rev-parse "refs/tags/$VERSION"` always returns exit code 1 (not found), the if body that would block the release is never entered, and the duplicate-tag check silently passes even when the tag already exists on origin.

This means a duplicate release version could make it all the way through stage + e2e + promote before failing at `git push` wasting a lot of CI time. Fix is to add `fetch-tags: true` to the checkout step so that remote tags are available locally for the uniqueness check, `git rev-parse` returns exit code 0 and fails fast when necessary.

andrii-i (dismissed):
Looks good, one note inline: the tag uniqueness check in validate would fail silently. Non-blocking as the pipeline still fails safely at git push but worth fixing to fail fast.

.github/workflows/release.yml
JGuinegagne: Great catch, thanks Andrii!

henrywa2 (approved):

andrii-i (approved):

PR #396, opened 2026-06-09
feat: access strategy template-induced hold

### LINKED ISSUE #386: Add validating webhook to reject deletion of in-use AccessStrategies
## Problem

When a user deletes a `WorkspaceAccessStrategy` that is still in use, `kubectl delete` returns immediately but the resource sits in "Terminating" state until the controller clears the finalizer. There is no upfront feedback explaining why the deletion is blocked.

Additionally, templates can reference access strategies via `spec.defaultAccessStrategy`. Deleting an access strategy referenced by a template should also be prevented.

## Current behavior

1. Workspace mutating webhook adds a finalizer to the referenced access strategy.
2. `kubectl delete` marks the resource for deletion (sets `DeletionTimestamp`).
3. The resource hangs in "Terminating" until no workspaces reference it.
4. No error message tells the user what's happening.

## Desired behavior

A validating webhook on `WorkspaceAccessStrategy` intercepts `DELETE` and rejects if:
- Any active workspace references the access strategy (same namespace or via shared namespace).
- Any template references it in `spec.defaultAccessStrategy`.

The webhook should return a clear error, e.g.:
> cannot delete: access strategy "web-access" is referenced by 3 workspaces and 1 template in namespace "team-notebooks"

The finalizer remains as a safety net (webhook uses `failurePolicy: Ignore`).

## Implementation notes

- Follow the same pattern as `WorkspaceTemplateCustomValidator`.
- Use `failurePolicy: Ignore` so cluster operations aren't blocked if the operator is unavailable.
- Only intercept `DELETE` (no validation needed on create/update).
- Check both workspace references and template `spec.defaultAccessStrategy` references.
- The controller's finalizer reconciliation logic remains unchanged.

## Acceptance criteria

- [ ] Webhook rejects deletion of access strategies referenced by active workspaces
- [ ] Webhook rejects deletion of access strategies referenced by templates
- [ ] Error message includes count and namespace of referencing resources
- [ ] `failurePolicy: Ignore` so cluster isn't blocked if webhook is down
- [ ] Finalizer logic unchanged (safety net)
- [ ] Unit tests for the webhook

JGuinegagne:
This PR adds lifecycle controls for access strategies to ensure that `Template.spec.AccessStrategyRef` do not have stale references. This enforces similar safety measure as for direct references between workspace and access strategy.

Closes: #386

Next step: implement shared namespace rules as in #349 .

### Customer experience
1. create a template referencing a missing access strategy (incorrect name or namespace) --> template webhook rejects
2. update a template to reference a missing access strategy (incorrect name or namespace) --> template webhook rejects
3. delete access strategy while a template references it --> finalizer holds deletion

### Implementation
1. on template admission webhook
    1.1. if ref an access strategy that does NOT exist, rejects
    1.2. if ref an access strategy, add a/ a finalizer to the AS, b/ tracking labels to the template
2. in the template reconciliation loop
    2.1. reconcile its own access strategy labels (on template) and adds - but does not remove - the finalizer (on access strategy)
3. the access strategy controller
    3.1. watches templates, excludes deleting template, include template that reference an access strategy (both before & after update), enqueues the access strategy for reconciliation
    3.2. this mechanism allows n templates to reference one access strategy w/ a single finalizer
4. in the access strategy reconciliation loop
    4.1. lookup workspaces and templates that may reference the access strategy, reconcile the access strategy finalizers 
5. add new finalizer for "referenced-by-template": `workspace.jupyter.org/accessstrategy-template-protection`

### Testing
- add E2E tests for template webhook admission
- add E2E tests for template-induced finalizers on access strategies
- add E2E tests for deletion hold on access strategies w/ template(s); both delete & de-reference
- add E2E tests for deletion hold on access strategies w/ mixed of workspace & template reference
- expand E2E tests for deletion hold on access strategies w/ multiple workspaces


henrywa2 (approved):

henrywa2:
LGTM. Any risk in the finalizers not being cleaned up and preventing access strategy deletion?

internal/controller/workspaceaccessstrategy_controller.go
```diff
@@ -52,53 +54,108 @@ func (r *WorkspaceAccessStrategyReconciler) Reconcile(ctx context.Context, req c
...
+// concurrent writer cannot make us clobber an unrelated change.
+func (r *WorkspaceAccessStrategyReconciler) reconcileFinalizers(
+	ctx context.Context,
+	logger logr.Logger,
+	accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy,
+	hasWorkspaces bool,
+	hasTemplates bool) (bool, error) {
 
-		// Use the safe utility to remove finalizer (handles conflicts)
-		err := workspace.SafelyRemoveFinalizerFromAccessStrategy(ctx, logger, r.Client, accessStrategy, false)
```
earaghbidikashani, line 91: can we remove `SafelyRemoveFinalizerFromAccessStrategy` function now?

earaghbidikashani (approved):

internal/controller/workspaceaccessstrategy_controller.go
JGuinegagne: wdym Eilia, this method was removed, wasn't it?

JGuinegagne:
> LGTM. Any risk in the finalizers not being cleaned up and preventing access strategy deletion?

In theory, if the operator is down or fails on the delete template/deref access strategy event.
However template have the same mechanism and this hasn't been an issue so far, so I would say the benefit of preventing stale refs offset the risk.

PR #398, opened 2026-06-11
feat: template access strategy ref rules

JGuinegagne:
This PR adds the same namespace consistency rules to `Template.spec.defaultAccessStrategy` field as to `Workspace.spec.accessStrategy`, namely either a/ same namespace as the template or b/ shared namespace. If `Template.spec.defaultAccessStrategy.namespace` is omitted, then defaults to the template's namespace.

This consistency enforcement is necessary because a Template configures a workspace, and these same rules apply at workspace admission time. If a template in namespace A referenced an access strategy in namespace B (!= from the shared namespace), then any workspace that try to use it would fail at admission time.

Refer to #349 for workspace <-> access strategy related rules.

Fast-follow from #396 

### User experience

- no change, but same backward incompatibility risk if existing users already referenced cross-namespaces AccessStrategy (deemed acceptable)

### Implementation

1. on create Template:
    1.1. if spec.defaultAccessStrategy is Nil -> OK
    1.2. if spec.defaultAccessStrategy.namespace == "" -> OK (interpreted as template namespace)
    1.3. if spec.defaultAccessStrategy.namespace == metadata.namespace -> OK
    1.4. if spec.defaultAccessStrategy.namespace == shared-namespace -> OK
    1.5. otherwise deny
2. on update Template:
    2.1. always allow to remove a defaultAccessStrategy

### Testing
- added E2E test cases for `defaultAccessStrategy` admission flow in create Template

### DIFF
--- a/internal/webhook/v1alpha1/template_webhook.go
+++ b/internal/webhook/v1alpha1/template_webhook.go
@@ -26,11 +26,15 @@ var templatelog = logf.Log.WithName("workspacetemplate-resource")
 
 // SetupWorkspaceTemplateWebhookWithManager registers the webhook for WorkspaceTemplate in the manager.
 func SetupWorkspaceTemplateWebhookWithManager(mgr ctrl.Manager, defaultTemplateNamespace string) error {
+	accessStrategyValidator := NewAccessStrategyValidator(defaultTemplateNamespace)
 	return ctrl.NewWebhookManagedBy(mgr).For(&workspacev1alpha1.WorkspaceTemplate{}).
 		WithValidator(&WorkspaceTemplateCustomValidator{
-			accessStrategyValidator: NewAccessStrategyValidator(defaultTemplateNamespace),
+			accessStrategyValidator: accessStrategyValidator,
 		}).
-		WithDefaulter(&WorkspaceTemplateCustomDefaulter{client: mgr.GetClient()}).
+		WithDefaulter(&WorkspaceTemplateCustomDefaulter{
+			client:                  mgr.GetClient(),
+			accessStrategyValidator: accessStrategyValidator,
+		}).
 		Complete()
 }
 
@@ -48,7 +52,8 @@ func SetupWorkspaceTemplateWebhookWithManager(mgr ctrl.Manager, defaultTemplateN
 // NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
 // as this struct is used only for temporary operations and does not need to be deeply copied.
 type WorkspaceTemplateCustomDefaulter struct {
-	client client.Client
+	client                  client.Client
+	accessStrategyValidator *AccessStrategyValidator
 }
 
 var _ webhook.CustomDefaulter = &WorkspaceTemplateCustomDefaulter{}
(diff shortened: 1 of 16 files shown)


aws-jasakshi (dismissed):

henrywa2 (dismissed):

internal/webhook/v1alpha1/template_webhook.go
```diff
@@ -109,7 +115,12 @@ func (v *WorkspaceTemplateCustomValidator) ValidateCreate(ctx context.Context, o
 	}
 	templatelog.Info("Validation for WorkspaceTemplate upon creation", "name", template.GetName())
 
-	// No special validation needed on create
+	// Enforce that the referenced access strategy is in an allowed namespace, so the template
+	// cannot make referencing workspaces fail their own admission webhook.
+	if err := v.accessStrategyValidator.ValidateCreateTemplate(template); err != nil {
```
andrii-i, line 129: Non-blocking: should the namespace scope be checked before the finalizer add in `Default()` so that a rejected template create does not modify an AccessStrategy in a namespace it is not allowed to reference?

test/e2e/template_namespace_scope_test.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Note: the cross-namespace rejection here is distinct from the "access strategy does not exist"
+// rejection covered in template_access_strategy_test.go. To prove the scope rule (not the existence
+// check, which the mutating webhook applies first) is what rejects, BeforeAll creates the team-b
+// access strategy so the reference resolves and only the scope rule can fail it.
+var _ = Describe("Template Namespace Scope", Ordered, func() {
+	const (
+		groupDir           = "template"
+		subgroup           = "scope"
+		workspaceNamespace = "default"
```
andrii-i: Nit: this holds the template's namespace, so `templateNamespace` would fit better and be more correct than `workspaceNamespace`

andrii-i (dismissed):

internal/webhook/v1alpha1/template_webhook.go
```diff
@@ -88,15 +90,19 @@ func (d *WorkspaceTemplateCustomDefaulter) Default(ctx context.Context, obj runt
 	return nil
 }
 
-// +kubebuilder:webhook:path=/validate-workspace-jupyter-org-v1alpha1-workspacetemplate,mutating=false,failurePolicy=ignore,sideEffects=None,groups=workspace.jupyter.org,resources=workspacetemplates,verbs=update,versions=v1alpha1,name=vworkspacetemplate-v1alpha1.kb.io,admissionReviewVersions=v1,serviceName=jupyter-k8s-controller-manager,servicePort=9443
+// +kubebuilder:webhook:path=/validate-workspace-jupyter-org-v1alpha1-workspacetemplate,mutating=false,failurePolicy=ignore,sideEffects=None,groups=workspace.jupyter.org,resources=workspacetemplates,verbs=create;update,versions=v1alpha1,name=vworkspacetemplate-v1alpha1.kb.io,admissionReviewVersions=v1,serviceName=jupyter-k8s-controller-manager,servicePort=9443
```
andrii-i, line 102: Non-blocking. Note that `docs/source/dive-deeper/webhooks/template-validation.md` says this webhook fires only on `UPDATE` (it now also fires on `CREATE`) and never blocks updates (it now rejects cross-namespace `defaultAccessStrategy` refs). `shared-namespace.md` also describes the namespace rule for workspace refs only, while templates now follow the same rule.

JGuinegagne: nice catches, thanks Andrii. Let me address them before merging.

andrii-i (approved):
Thank you @JGuinegagne.

aws-jasakshi (approved):

PR #399, opened 2026-06-12
feat: add WorkspaceIntegrationStrategy CRD with template resolution

HariniNarayanan:
Adds the WorkspaceIntegrationStrategy CRD: an admin-installed, reusable strategy that declaratively injects runtime capabilities (sidecars, volumes, env vars) into a workspace pod with dynamic template resolution. Customers reference it from Workspace.spec.integrationStrategy with name/value parameters.

- CRD types + IntegrationStrategyRef on Workspace.spec (extensible name/value parameter list with ParametersMap helper)
- Two-pass template resolution engine: {{ .Workspace.* }}, {{ .Parameters.* }}, and {{ resource "{jsonpath}" }} against a looked-up resource
- Controller wiring: resolution runs in the canonical deployment build path; watch + label-sync + event handler mirror the access strategy pattern
- Vetted Ray connector sample, e2e tests, generated CRDs

Note: resourceLookup performs dynamic GETs for arbitrary GVKs, but the controller RBAC currently grants only ray.io/rayclusters (get,list) since Ray is the only shipped integration in v1. Looking up other resource kinds will require additional RBAC grants; broadening this (e.g. configurable/aggregated RBAC) is deferred to a follow-up.


HariniNarayanan:
We're changed the design since this PR - closing this

PR #401, opened 2026-06-14
fix: set all conditions coherently in UpdateDeletingStatus

bhanderig:
## Summary

- Add `ConditionTypeDeleting` and `ReasonDeletionInProgress` constants to `conditions.go`
- Rewrite `UpdateDeletingStatus` to set all 5 conditions coherently (matching the pattern used by `UpdateRunningStatus`, `UpdateStoppedStatus`, etc.)
- Remove direct `meta.SetStatusCondition` usage which only appended `Deleting=True` without clearing `Available=True`

## Problem

When deleting a workspace, the controller set `Deleting=True` via `meta.SetStatusCondition` without touching other conditions. This left `Available=True` alongside `Deleting=True`, causing the UI to display both "Running" and "Deleting" simultaneously.

## Fix

`UpdateDeletingStatus` now follows the same pattern as other status methods — explicitly sets all conditions with appropriate reasons:
- `Available=False` (reason: DeletionInProgress)
- `Progressing=False` (reason: DeletionInProgress)
- `Degraded=False` (reason: NoError)
- `Stopped=False` (reason: DeletionInProgress)
- `Deleting=True` (reason: DeletionInProgress)

## Test plan

- [x] Deploy updated controller to EKS cluster
- [x] Create workspace, wait for Running, delete it
- [x] Verify `kubectl get workspace <name> -o json` shows coherent conditions (Available=False, Deleting=True)
- [x] Verify UI shows "Deleting" status (not "Running")

### DIFF
--- a/internal/controller/state_machine_deletion_test.go
+++ b/internal/controller/state_machine_deletion_test.go
@@ -98,6 +98,18 @@ var _ = Describe("ReconcileDeletion", func() {
 		availableCond = findCondition(ws.Status.Conditions, ConditionTypeAvailable)
 		Expect(availableCond).NotTo(BeNil())
 		Expect(availableCond.Status).To(Equal(metav1.ConditionFalse))
+
+		progressingCond := findCondition(ws.Status.Conditions, ConditionTypeProgressing)
+		Expect(progressingCond).NotTo(BeNil())
+		Expect(progressingCond.Status).To(Equal(metav1.ConditionFalse))
+
+		degradedCond := findCondition(ws.Status.Conditions, ConditionTypeDegraded)
+		Expect(degradedCond).NotTo(BeNil())
+		Expect(degradedCond.Status).To(Equal(metav1.ConditionFalse))
+
+		stoppedCond := findCondition(ws.Status.Conditions, ConditionTypeStopped)
+		Expect(stoppedCond).NotTo(BeNil())
+		Expect(stoppedCond.Status).To(Equal(metav1.ConditionFalse))
 	})
 
 	It("should skip when no finalizer is present", func() {
--- a/internal/controller/state_machine_deletion_test.go
+++ b/internal/controller/state_machine_deletion_test.go
@@ -87,9 +87,9 @@ var _ = Describe("ReconcileDeletion", func() {
 		_, err := sm.ReconcileDeletion(ctx, ws)
 		Expect(err).NotTo(HaveOccurred())
 
-		// Reload and verify conditions
-		Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(ws), ws)).To(Succeed())
-
+		// After ReconcileDeletion, the workspace is modified in-place with updated conditions.
+		// The object may be garbage-collected (finalizer removed + DeletionTimestamp set),
+		// so we verify the in-memory object rather than reloading from the API.
 		deletingCond := findCondition(ws.Status.Conditions, ConditionTypeDeleting)
 		Expect(deletingCond).NotTo(BeNil())
 		Expect(deletingCond.Status).To(Equal(metav1.ConditionTrue))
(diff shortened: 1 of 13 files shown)


JGuinegagne (commented):
logic LGTM, but please:
- add unit tests to the `state_machine` (this is non-trivial, but your agent should be able to follow precedent)
- try adding E2E tests [workspace_status_test](https://github.com/jupyter-infra/jupyter-k8s/blob/main/test/e2e/workspace_status_test.go) --> as in, a/ transition (check all the statuses), b/ ultimately get deleted.

internal/controller/state_machine_deletion_test.go
```diff
@@ -0,0 +1,151 @@
...
+		// The object may be garbage-collected (finalizer removed + DeletionTimestamp set),
+		// so we verify the in-memory object rather than reloading from the API.
+		deletingCond := findCondition(ws.Status.Conditions, ConditionTypeDeleting)
+		Expect(deletingCond).NotTo(BeNil())
+		Expect(deletingCond.Status).To(Equal(metav1.ConditionTrue))
+		Expect(deletingCond.Reason).To(Equal(ReasonDeletionInProgress))
+
+		availableCond = findCondition(ws.Status.Conditions, ConditionTypeAvailable)
+		Expect(availableCond).NotTo(BeNil())
+		Expect(availableCond.Status).To(Equal(metav1.ConditionFalse))
```
JGuinegagne, line 100: non-blocking: shall we check `Progressing`, `Stopped` and `Degraded` while we are at it?

test/e2e/workspace_status_test.go
```diff
@@ -371,6 +371,52 @@ var _ = Describe("Workspace Status", Ordered, func() {
...
+			By("waiting for Deleting condition to become True")
+			WaitForWorkspaceToReachCondition(
+				deletionWorkspace,
+				statusTestNamespace,
+				ConditionTypeDeleting,
+				ConditionTrue,
+			)
+
+			By("verifying Available=False while Deleting=True")
+			VerifyWorkspaceConditions(deletionWorkspace, statusTestNamespace, map[string]string{
```
JGuinegagne, line 409: callout: let's try this logic, but it may lead to race condition if the pods are deleted too fast and the workspace is removed from ETCD. If that happens, we can harden to not fail if the workspace is missing.

test/e2e/workspace_status_test.go
```diff
@@ -371,6 +371,52 @@ var _ = Describe("Workspace Status", Ordered, func() {
...
+				ConditionTypeProgressing: ConditionFalse,
+				ConditionTypeDegraded:    ConditionFalse,
+				ConditionTypeStopped:     ConditionFalse,
+				ConditionTypeDeleting:    ConditionTrue,
+			})
+
+			By("waiting for workspace to be fully deleted")
+			Eventually(func() bool {
+				_, err := kubectlGet("workspace", deletionWorkspace, statusTestNamespace, "{.metadata.name}")
+				return err != nil
```
JGuinegagne: hum, can we assert that we eventually hit a `NotFound` error, not just any error?

Ideally create a utility in [workspace_helpers](https://github.com/jupyter-infra/jupyter-k8s/blob/main/test/e2e/helpers_workspaces.go)

JGuinegagne (dismissed):

earaghbidikashani:
Quick comment, we want to make sure that only one condition is true at all times. 

Can you add fifth entry for the remaining statuses please? 
https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/controller/status_manager.go#L229

bhanderig:
> Quick comment, we want to make sure that only one condition is true at all times.
> 
> Can you add fifth entry for the remaining statuses please? https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/controller/status_manager.go#L229

nice callout, updated

earaghbidikashani (approved):

JGuinegagne (approved):

Issue #267, opened 2025-11-14
[hyperpod-remote-access] enable ssh server, ssm agent logs to be uploaded to Cloudwatch for troubleshooting


aws-prayags:

JGuinegagne:
Obsolete here after the vendor-neutral refactor: SSM-agent/CloudWatch log shipping is a plugin-sidecar and guided-chart concern. Please retarget to https://github.com/jupyter-infra/jupyter-k8s-aws.

PR #410, opened 2026-06-16
feat: automated roborev code review

### LINKED ISSUE #400: Add automated roborev PR reviews
## Problem

There's potential to save time for contributors and reviewers in this repo by automatically reviewing new PRs and surfacing problems to address before the maintainer first pass.

[roborev](https://github.com/roborev-dev/roborev) provides the review engine: its CI mode (`roborev ci review`) runs the configured review types, synthesizes the results, and posts one PR comment, with no daemon or database. It does not provide the infrastructure: a runtime with roborev and its agent dependencies installed, model access, and credentials in CI.

Note that following the standard OSS model, contributions come from forks, and GitHub gives fork-triggered `pull_request` workflows no secrets and no OIDC token. So the review job needs one of the privileged triggers that run with base-repo credentials, which is what shapes the workflow design below.

## Proposed solution

Automatically review new PRs: one roborev review comment per PR revision, posted after the author stops pushing. Reviews run in GitHub Actions with Claude on Amazon Bedrock as the model backend. Bedrock is invoked on demand through a single IAM role assumed via GitHub OIDC, so no API keys are stored anywhere: not in repo secrets, not in the image, not in the account.

The AWS side reuses the CI infrastructure patterns from [jupyter-deploy](https://github.com/jupyter-infra/jupyter-deploy), which already provisions OIDC roles and pre-built ECR images for its E2E pipeline.

Architecture:

1. Review workflow in this repo. Triggered on `pull_request_target` for PR open, reopen, ready for review, and push; unlike `pull_request`, this trigger runs with base-repo credentials for fork PRs. The review fires per PR revision and runs in parallel with CI, independent of its outcome. roborev's CI mode is stateless, so the workflow owns the once-per-revision semantics: a per-PR concurrency group cancels any queued or in-progress review when a new push arrives, and the job verifies the revision is still the PR head before reviewing, so only the latest revision gets reviewed. The job assumes the review IAM role via OIDC, pulls the review image from ECR, and runs `roborev ci review` against the PR's ref range; then roborev posts the comment. A `workflow_dispatch` input reviews any PR on demand.

   The job never executes PR code: it checks out the base branch, fetches the PR head as git objects, and passes roborev a ref range. If a malicious diff prompt-injects the review agent, the worst it can do is spend model tokens and post a misleading comment; the role and the workflow's `GITHUB_TOKEN` allow nothing else.

2. Review image, published by jupyter-deploy. A container with roborev, the claude-code agent, and their dependencies, all version-pinned, built and pushed to ECR the same way [e2e-build-image.yml](https://github.com/jupyter-infra/jupyter-deploy/blob/556593df48c4fc880a05a72f3a8512ec1bb73eeb/.github/workflows/e2e-build-image.yml) publishes the E2E images. This repo's job only assumes a role, docker pulls, docker runs.

3. IAM template, in jupyter-deploy. A template lib with the same structure as [jupyter-infra-tf-aws-iam-ci](https://github.com/jupyter-infra/jupyter-deploy/tree/main/libs/jupyter-infra-tf-aws-iam-ci), deployed via `jd init` / `jd up`. It creates the ECR repository and one review role: trust scoped to a GitHub Actions environment (`review`) for an explicit list of repos, permissions limited to Bedrock model invocation and pulling the review image, hardened with the same deny statements as the existing e2e and release roles.

4. Review policy and local setup, in this repo, to get roborev reviews locally after every commit. This allows contributors to iterate on and improve their work locally before pushing and triggering CI review. Concretely: a committed `.roborev.toml` with the review configuration (review types, reasoning level, severity floor), shared by CI and local runs, and an AGENT.md section documenting the standard roborev flow: install roborev, run `roborev init` to set up the post-commit hook, then every commit gets reviewed in the background; `roborev show HEAD` shows findings, `roborev refine` runs the iterative review-fix loop. Local reviews use whatever coding agent the contributor already has (claude-code, codex, gemini, etc.). Local review is opt-in.

## Alternatives considered

Managed review products (CodeRabbit, GitHub Copilot code review, Amazon Q). Rejected on the constraints above: inference runs on vendor endpoints and accounts (CodeRabbit routes code through its own model accounts), and policy, prompts, and engine internals are not ours to modify. The durable asset to own is the trust architecture; the engine should be swappable.

claude-code-action on Bedrock. Meets the own-endpoint constraint, but the action process that reads the untrusted diff also holds the GitHub token, mitigated by tool allowlists rather than eliminated. The chosen design keeps the token out of the diff-processing context entirely: the container prints the review, a runner step posts it.

Per-run approval gates for fork PRs. A collaborator-gated or label-gated privileged workflow re-adds a human to every revision, the latency this system exists to remove, and GitHub's guidance documents the approve-then-push race. Reviewing PR code as data under `pull_request_target` (base-branch workflow copy, head fetched as git objects, never executed) removes the need for a per-run gate; the residual worst case is a wasted review comment, bounded by a role that can only pull one image and invoke Bedrock.

Stored model API keys. Rejected in favor of OIDC: per-run credentials that expire within the hour and exist nowhere at rest.

andrii-i:
Closes #400. Adds automated roborev review for jupyter-k8s: one review comment per PR revision, Claude on Amazon Bedrock, and no API keys stored anywhere. Because fork PRs get no secrets or OIDC token under `pull_request`, the review runs from `pull_request_target` with base-repo credentials and never executes the PR's code. Depends on jupyter-infra/jupyter-deploy#267, which publishes the review image and provisions the IAM role this workflow assumes, so merge and deploy that side first.

Review workflow (`.github/workflows/roborev-review.yml`)
- Triggers on `pull_request_target` (opened, reopened, ready for review, synchronize) and `workflow_dispatch`.
- Never runs PR code: checks out the trusted base, fetches the PR head as git objects into `refs/roborev/head` without checking it out, and passes roborev a `merge-base..head` ref range.
- Once per revision: a per-PR concurrency group cancels superseded runs, and a stale-head guard re-checks the live PR head before spending any tokens.
- Assumes the review role via OIDC, pulls the review image from ECR, and runs `roborev ci review`; roborev posts the comment. A commit status tracks progress and is always resolved, so a cancelled, failed, or smoke run never leaves a dangling pending status.
- Inert until the repo variables `REVIEW_RUN_ROLE_ARN`, `REVIEW_IMAGE`, and `REVIEW_BEDROCK_MODEL` are set (optional `REVIEW_AWS_REGION`): the job is skipped otherwise, posting no status.

Local review (`.roborev.toml`, `Makefile`, `AGENT.md`, `CONTRIBUTING.md`, `README.md`)
- `.roborev.toml` holds the shared review policy (claude-code agent, `default` and `security` review types).
- `make review` runs a one-shot local review of the branch against main; `make review-setup` installs the opt-in post-commit hook. Local review is optional; CI reviews every PR regardless.

### Testing

- Traced the commit-status lifecycle across every path (fetch failure, cancel, timeout, smoke, stale-head skip, unset-vars skip): the pending status and the `if: always()` resolver share the same guards, so no path leaves a dangling status.
- Confirmed the inert path is a clean job-level skip (no status, no error) when the three repo variables are unset.
- `make -n review` / `make -n review-setup` expand to the expected roborev commands; the targets guard with an actionable message when roborev is not installed.
- Verified against the installed roborev that the `[ci]` matrix (`default` + `security`) is consumed by `roborev ci review` while the local `roborev review` runs a single review; the docs describe this rather than claim local and CI findings match.

The workflow is intentionally inert until the three repo variables are set, so it merges safely before the infrastructure exists. Once #267 is deployed and the variables are set, a `workflow_dispatch` run with `smoke: true` validates the OIDC -> ECR -> Bedrock chain (it pulls the image and asserts a Bedrock round-trip) before the first real review.


.github/workflows/roborev-review.yml
```diff
@@ -0,0 +1,190 @@
...
+            -f description="review running" > /dev/null
+
+      - name: Smoke test (Bedrock reachability)
+        if: steps.pr.outputs.skip != 'true' && inputs.smoke
+        env:
+          REVIEW_IMAGE: ${{ vars.REVIEW_IMAGE }}
+        run: |
+          set -euo pipefail
+          out=$(docker run --rm \
+            -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN \
```
JGuinegagne: oof, we shouldn't pass creds in plain text like this, and anyway you need to pick up rotated creds.

Mount the creds dir please, example [here](https://github.com/jupyter-infra/jupyter-deploy/blob/3810f1c919873c63e7f93bb14385bab03b0c8b58/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/image/docker-compose.yml#L18)

.github/workflows/roborev-review.yml
```diff
@@ -0,0 +1,190 @@
...
+          git fetch --no-tags origin "+${HEAD_SHA}:refs/roborev/head" 2>/dev/null \
+            || git fetch --no-tags origin "pull/${PR}/head:refs/roborev/head"
+          base_sha=$(git merge-base "$BASE_SHA" "$HEAD_SHA")
+
+          docker run --rm \
+            -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN \
+            -e AWS_REGION -e AWS_DEFAULT_REGION \
+            -e CLAUDE_CODE_USE_BEDROCK=1 \
+            -e ANTHROPIC_MODEL="${{ vars.REVIEW_BEDROCK_MODEL }}" \
+            -e GH_TOKEN \
```
JGuinegagne: same comment, don't pass the GH_TOKEN in plain text, use a mount please.

JGuinegagne (changes requested):
Watch out for credentials / token security and creds rotation

.github/workflows/roborev-review.yml
andrii-i: A staging step now writes the OIDC session creds to `~/.aws/credentials`, and both runs mount that read-only instead of passing the keys with `-e`. No AWS secrets are in the container environment anymore.

These creds are issued once per run and outlive it: the OIDC assume-role gives session creds valid for 1h against a 30-min job cap. `configure-aws-credentials` runs once, so nothing refreshes them mid-run and the mounted file holds the same credentials for the whole run. There's no rotation to pick up: a CI runner has no instance or workload identity to refresh from, and the credential outlives the run regardless.

.github/workflows/roborev-review.yml
andrii-i: Took this further than a mount: the gh token never enters the review container. roborev can print the review to stdout instead of posting it (logs go to stderr), so the container processes the untrusted diff with only the Bedrock creds, and the runner posts the captured output with `gh pr comment` using the `GITHUB_TOKEN` already in the runner env. Nothing gh-related is passed to the container, env or mount.

A config mount is doable but couples to gh's internals: the file has to match gh's current on-disk format (the `users:` map keyed by the bot account, not a top-level `oauth_token`, which gh tries to migrate), so it pins to the gh version in the image. Posting from the runner sidesteps that and, more to the point, keeps the token out of the untrusted container entirely.

JGuinegagne (approved):

bhanderig (approved):

PR #411, opened 2026-06-17
fix: idle shutdown logic

### LINKED ISSUE #409: Idle detection: eliminate image dependency, support base URLs, make response format configurable
## Problem

The current idle detection implementation has three design issues that prevent it from working
with standard JupyterLab images:

### 1. Requires `curl` inside the workspace container

The operator execs `curl` into the workspace pod to probe the HTTP endpoint. This fails for
any image that doesn't bundle curl (including the jk8s reference image `jupyter-uv`, which
uses `python3.12-bookworm-slim`).

```
"curl": executable file not found in $PATH
```

The operator retries every 5 minutes indefinitely — idle shutdown never triggers.

Requiring specific binaries in user images is fragile and undocumented. Users shouldn't need
to install CLI tools just to satisfy operator internals.

### 2. Detection path doesn't account for application base URL

When an access strategy routes traffic to the workspace under a path prefix
(e.g., `/workspaces/<namespace>/<name>/`), the application's endpoints live under that prefix.
JupyterLab's `/api/status` is actually at `/workspaces/default/my-ws/api/status`.

The detection `httpGet.path` is used verbatim — there's no mechanism to incorporate the
per-workspace routing prefix. A static path in the WorkspaceTemplate can never work for all
workspaces since each has a different base URL.

### 3. Hardcoded response format

The Go struct expects exactly `{"lastActiveTimestamp": "..."}`:

```go
type EndpointIdleResponse struct {
    LastActivity string `json:"lastActiveTimestamp"`
}
```

But JupyterLab's built-in `/api/status` returns:

```json
{"connections": 0, "kernels": 0, "last_activity": "2026-06-11T22:41:34.836944Z", "started": "..."}
```

Users must create a custom server extension just to bridge the field name — that's not
reasonable for a stock JupyterLab deployment.

### A. Direct HTTP calls from the operator (`transport: network`)

Replace `kubectl exec curl` with a direct Go `net/http` call from the operator pod.
The workspace Pod IP is already available at idle-check time (`pod.Status.PodIP`).

```yaml
detection:
  httpGet:
    path: /api/status
    port: 8888
    transport: network       # new; "podExec" kept as default for backward compat
```

- `network`: operator calls `http://<podIP>:<port><resolvedPath>` directly
- `podExec` (default): legacy behavior — exec curl inside the container

This eliminates the implicit dependency on workspace image contents.

**Schema implementation:** `IdleHTTPGetAction` embeds `corev1.HTTPGetAction` inline via
`json:",inline"`, adding `transport` and `lastActivityTimestamp` as sibling fields. This
preserves wire-compatibility with existing stored CRs automatically (same JSON field
names/types for `path`/`port`/`scheme`). A reflect-based CI test guards against future
upstream field-name collisions.

### B. `applicationBasePath` on AccessStrategy (materialized to workspace status)

The access strategy determines the routing prefix. Add a field so the operator can resolve it:

```yaml
kind: WorkspaceAccessStrategy
spec:
  applicationBasePath: "/workspaces/{{.Workspace.Namespace}}/{{.Workspace.Name}}/"
```

- Resolved using the existing `ResolveTemplateURL` engine (same as `accessURLTemplate` and
  `mergeEnv` — exposes `.Workspace`, `.AccessStrategy`, `.Service`).
- Defaults to `/` when absent (port-forward, no access strategy).

**Implementation — materialized to `workspace.status.applicationBasePath`:**

The resolved base path is stored in `workspace.Status.ApplicationBasePath` during the
access-resources reconcile step — the same pass that sets `status.accessURL` and rolls
the deployment. The idle checker reads from this status field (not from the live
AccessStrategy).

Rationale: if an admin updates `applicationBasePath` on the AccessStrategy but the workspace
hasn't been reconciled yet, the pod is still serving on the old path. Dynamically resolving
the new path at idle-check time would produce a 404 against the still-running old pod. The
status field always reflects what was configured in the same reconcile that updated the
deployment — it stays synchronized with what the pod is actually serving.

At idle-check time: `workspace.Status.ApplicationBasePath + httpGet.path` → full URL path.

The detection config stays application-relative:
```yaml
detection:
  httpGet:
    path: /api/status    # relative to the app root
```

Separation of concerns:
- **Workspace/Template** declares: "where does the app expose its status endpoint?"
- **AccessStrategy** declares: "where does this app live in the routing tree?"

### C. Configurable last-activity timestamp parsing

Add a `lastActivityTimestamp` section to `httpGet`:

```yaml
detection:
  httpGet:
    path: /api/status
    port: 8888
    lastActivityTimestamp:
      responseBodyPath: last_activity     # dot-separated path into JSON response body
      format: RFC3339                     # default; "unix" for epoch seconds
```

- `responseBodyPath` — dot-separated path to the timestamp value in the JSON response body.
  Default: `lastActiveTimestamp` (backward compat).
- `format` — how to parse it. Default: `RFC3339`. Support `unix` for epoch-based responses.
- Implementation: replace the hardcoded struct with generic JSON unmarshaling + dot-path extraction.

### End state

With all three changes, a stock JupyterLab image works out of the box:

```yaml
# AccessStrategy
spec:
  applicationBasePath: "/workspaces/{{.Workspace.Namespace}}/{{.Workspace.Name}}/"

# WorkspaceTemplate
spec:
  defaultIdleShutdown:
    enabled: true
    idleTimeoutInMinutes: 60
    detection:
      httpGet:
        path: /api/status
        port: 8888
        transport: network
        lastActivityTimestamp:
          responseBodyPath: last_activity
```

No curl. No custom extensions. No per-workspace path hacks.

### D. Configurable poll interval (test infrastructure)

The hardcoded 5-minute `IdleCheckInterval` becomes a startup flag `--idle-check-interval`
(default `5m`) with a corresponding Helm value (`controller.idleCheckInterval`). This enables
E2E testing of idle shutdown in seconds rather than waiting 5+ minutes per check cycle.

## Migration / Backward Compatibility

All changes are additive — existing WorkspaceTemplates and AccessStrategies that work
today must continue to work without modification.

### `transport` field

Default: **`podExec`** (preserves current behavior).

Rationale: images like SageMaker Distribution already have `curl` installed and serve
`/api/idle` with the expected contract. Changing the default to `network` would break setups
where the app binds only to `127.0.0.1` (unreachable from the operator pod via Pod IP).
Users opt into `network` explicitly.

When `transport` is omitted, behavior is identical to today.

### `lastActivityTimestamp` field

Default when omitted: operator parses `lastActiveTimestamp` from the JSON root (current behavior).

- `lastActivityTimestamp.responseBodyPath` defaults to `lastActiveTimestamp`
- `lastActivityTimestamp.format` defaults to `RFC3339`

Existing images that return `{"lastActiveTimestamp": "2024-01-01T00:00:00Z"}` continue to work
with no config change.

### `applicationBasePath` field

Default when omitted on AccessStrategy: `/` (no prefix applied).

When the field is absent, the operator uses `httpGet.path` verbatim — identical to current
behavior. Existing setups where the app serves at root (or users have already configured the
full path including base URL) are unaffected.

### Summary

| Field | Default | Effect on existing configs |
|-------|---------|---------------------------|
| `transport` | `podExec` | No change — same curl-in-pod behavior |
| `lastActivityTimestamp.responseBodyPath` | `lastActiveTimestamp` | No change — same JSON field parsed |
| `lastActivityTimestamp.format` | `RFC3339` | No change — same time format |
| `applicationBasePath` | `/` | No change — path used verbatim |
| `--idle-check-interval` | `5m` | No change — same poll cadence |

Zero-config migration: existing CRDs without the new fields behave exactly as before.

## Design decisions

- **Embed, don't redeclare:** `IdleHTTPGetAction` embeds `corev1.HTTPGetAction` with
  `json:",inline"` — `path`/`port`/`scheme` are the actual upstream fields. Wire-compat for
  existing CRs is automatic. A reflect-based CI test fails if a future `k8s.io/api` bump
  introduces a field whose JSON tag collides with `transport` or `lastActivityTimestamp`.

- **`transport` naming (not `source`/`via`):** avoids colliding with the k8s probe handler
  name "exec"; follows gRPC/Envoy/containerd precedent. Values `podExec` (curl-in-pod) and
  `network` (direct HTTP to pod IP).

- **`lastActivityTimestamp` (not `response`):** collision-proof against upstream; precisely names
  what the block describes rather than implying generic response formatting. Sub-fields
  `responseBodyPath` (dot-separated path into the JSON response body, avoids ambiguity with the
  parent `httpGet.path` URL path) and `format` (RFC3339/unix).

- **Materialized `applicationBasePath` (not dynamic pull):** the status field is updated in the
  same reconcile pass that rolls the deployment, so it always reflects what the pod is actually
  configured to serve. A dynamic pull from the live AccessStrategy could produce a stale mismatch
  (new path resolved against an old pod still serving on the previous path → 404).

- **Template engine reuse:** `applicationBasePath` uses the existing `ResolveTemplateURL` engine
  with `.Workspace`/`.AccessStrategy`/`.Service` context — same as `accessURLTemplate` and
  `mergeEnv`. One engine, one convention.

- **Future extensibility (YAGNI — not built now):**
  - Non-JSON HTTP bodies: add `bodyFormat` enum under `httpGet` (KEDA precedent).
  - gRPC/exec handlers: add sibling `detection.exec: { command, lastActivityTimestamp }` or
    `detection.grpc: { ... }`. Each handler owns its own parsing contract.

## Scope

- CRD schema changes (AccessStrategy + IdleDetectionSpec + WorkspaceStatus)
- Controller: add `transport: network` path (net/http to pod IP) alongside existing `podExec` path
- Controller: materialize resolved `applicationBasePath` to workspace status during access reconcile
- Controller: idle checker reads base path from status, composes full URL
- Controller: generic JSON field extraction for last-activity timestamp parsing
- Controller: configurable poll interval (startup flag + Helm value)
- Webhook: validate new fields (enum markers handle most; template-parse `applicationBasePath`)
- Documentation: update idle-shutdown page
- Reference image: remove curl if/when added as interim fix
- E2E: test with stock JupyterLab `/api/status` (no custom extension)

JGuinegagne:
This PR extends the idle shutdown logic to a/ ensure compatibility with OSS `jupyterlab` images by extracting the timestamp from the response with `jsonpath`, b/ support direct HTTP GET to the workspace service instead of relying exclusively on making a `curl` call via `podExec` , c/ support workspaces tied to an access strategy that modifies their app base path, and d/ support configuring the timeout check interval in the operator (for testing purposes mainly).

Closes: #409 

### User experience

**Workspace spec for idle shutdown via network for plain JupyterLab (new)**
```yaml
idleShutdown:
  enabled: true
  idleTimeoutInMinutes: 60
  detection:
    httpGet:
      path: /api/status
      port: 8888
      transport: network
      lastActivityTimestamp:
        responseBodyPath: last_activity
        format: RFC3339
```

**Workspace spec for idle shutdown via podExec for plain JupyterLab**
```yaml
idleShutdown:
  enabled: true
  idleTimeoutInMinutes: 60
  detection:
    httpGet:
      path: /api/status
      port: 8888
      transport: podExec
      lastActivityTimestamp:
        responseBodyPath: last_activity
        format: RFC3339
```

### Implementation
- update `workspace.spec.idleShutdown.detection.httpGet` to:
     - build a mixin new type that merges `corev1.HTTPGetAction`
     - adds a `transport` field - enum: `{network|podExec}` default: `podExec`
     - adds a `lastActivityTimestamp` field of type:
         - `responseBodyPath`: 
         - `format`: enum: `{RFC3339|unix}`, default: `RFC3339`
- add `applicationBasePathTemplate` to `accessstrategy.spec`
- add `applicationBasePath` to `workspace.status`
- add an `idleShutdown.checkInterval` flag to the operator helm chart (default to `5m`)
- add an `--idle-check-interval` flag (default to `5m` if missing) to the controller manager
- pass down the `idleCheckInterval` to the workspace controller, use it to configure the `WorkspaceIdleChecker`
- update workspace controller to:
    - add `checkViaNetwork()` method to `WorkspaceIdleChecker`
    - add `NetworkHTTPGetDetector` construct to `idle_detectors`
    - implement a `CheckIdle` which sends an HTTP GET request to the workspace service
    - add utilities to parse the JSON response based on the `lastActivityTimestamp.responseBodyPath` field value of the workspace
    - resolve the application base path of a workspace from `accessstrategy.spec.applicationBasePathTemplate` and set it in the `workspace.status.applicationBasePath` (when running only)
    - pass down the workspace service to the idle config handler in the state machine of the workspace reconciliation loop
    - use the `WorkspaceIdleChecker.idleCheckInterval` to schedule the next reconciliation in the state machine

### Testing
Added comprehensive E2E tests for idle shutdown:
- Network transport stops idle workspaces (with and without access strategy)
- PodExec transport stops idle workspaces
- Long timeout workspaces keep running (proves the check fires but respects the timeout)
- Active workspaces with real JupyterLab + keepalive pod stay running (proves the timestamp parsing actually works end-to-end)


api/v1alpha1/workspace_types.go
```diff
@@ -89,7 +89,43 @@ type IdleShutdownSpec struct {
...
+	HTTPGet *IdleHTTPGetAction `json:"httpGet,omitempty"`
+}
+
+// IdleHTTPGetAction extends corev1.HTTPGetAction with transport and response parsing options.
+type IdleHTTPGetAction struct {
+	corev1.HTTPGetAction `json:",inline"`
+
+	// Transport selects how the operator reaches the endpoint.
+	// "podExec" executes curl inside the workspace container (legacy).
+	// "network" makes a direct HTTP call from the operator to the pod IP.
```
andrii-i: `network` always calls the Service ClusterIP (idle_checker.go:77). Should this say "...to the workspace Service's ClusterIP."?

bhanderig (commented):
Solid PR overall, few inline comments.

internal/controller/state_machine.go
```diff
@@ -349,19 +349,11 @@ func (sm *StateMachine) handleIdleShutdownForRunningWorkspace(
...
-
-	result, err := sm.idleChecker.CheckWorkspaceIdle(ctx, workspace, idleConfig)
+	// No explicit pod-readiness check needed here: this code only runs after
+	// IsDeploymentAvailable()==true, which guarantees at least one ready pod.
+	// If a pod crashes between the deployment check and the idle probe, the
+	// probe returns ShouldRetry=true and we retry next cycle.
+	result, err := sm.idleChecker.CheckWorkspaceIdle(ctx, workspace, service, idleConfig)
 	if err != nil {
 		if !result.ShouldRetry {
 			logger.Error(err, "Permanent failure checking idle status, disabling idle shutdown for this workspace")
```
bhanderig, line 359: **[Observability]** On permanent failure (e.g., 404), idle shutdown silently stops working — only this log line fires, with no requeue. A cluster admin wouldn't know idle shutdown is broken unless they read operator logs.

Consider also emitting a K8s Warning Event here:
```go
sm.recorder.Event(workspace, corev1.EventTypeWarning, "IdleShutdownDisabled",
    fmt.Sprintf("Idle shutdown permanently disabled: %v", err))
```
This makes the failure visible via `kubectl describe workspace`. A status Condition (e.g., `IdleShutdownDegraded`) would be even better for programmatic alerting, but an Event is a low-cost first step.

internal/controller/idle_detectors.go
```diff
@@ -112,48 +118,190 @@ func (h *HTTPGetDetector) CheckIdle(ctx context.Context, workspaceName string, p
...
+
+// --- network transport (direct HTTP from operator) ---
+
+// NetworkHTTPGetDetector implements HTTP endpoint checking via direct network call
+type NetworkHTTPGetDetector struct {
+	httpClient *http.Client
+}
+
+// NewNetworkHTTPGetDetector creates a new NetworkHTTPGetDetector with a shared http.Client
+func NewNetworkHTTPGetDetector() *NetworkHTTPGetDetector {
```
bhanderig: **[Efficiency + Hardening]** Two concerns with the http.Client lifecycle:

1. **Per-check allocation:** `createIdleDetectorImpl` → `NewNetworkHTTPGetDetector()` is called every idle check cycle, allocating a fresh `http.Client` with its own connection pool. Connections are never reused across checks. Consider promoting the client to a long-lived field on `WorkspaceIdleChecker` so TCP connections to frequently-probed services are reused.

2. **No TLS story:** If a user sets `scheme: HTTPS` (valid per the CRD schema since `corev1.URIScheme` allows it), the default transport will verify the server certificate against system CAs. In-cluster services with self-signed certs (e.g., from cert-manager) would fail with a TLS error, and idle shutdown would silently retry forever without stopping the workspace. Consider either:
   - Documenting that `scheme: HTTPS` is unsupported for idle detection, or
   - Adding a `tlsInsecureSkipVerify` field (or custom CA bundle) if HTTPS is a real use case later.

internal/controller/idle_checker.go
```diff
@@ -29,52 +30,88 @@ type IdleCheckResult struct {
...
+	client        client.Client
+	checkInterval time.Duration
 }
 
-// NewWorkspaceIdleChecker creates a new WorkspaceIdleChecker instance
-func NewWorkspaceIdleChecker(k8sClient client.Client) *WorkspaceIdleChecker {
+// NewWorkspaceIdleChecker creates a new WorkspaceIdleChecker instance.
+// If checkInterval is zero or negative, DefaultIdleCheckInterval is used.
+func NewWorkspaceIdleChecker(k8sClient client.Client, checkInterval time.Duration) *WorkspaceIdleChecker {
+	if checkInterval <= 0 {
```
bhanderig: **[Hardening]** No minimum floor on `checkInterval` — values like `1s` or `100ms` are accepted and would cause the operator to reconcile + HTTP-probe every running workspace at that rate. At scale (100+ workspaces), this could saturate the controller's work queue and generate significant network traffic.

Consider a hard floor:
```go
const MinIdleCheckInterval = 5 * time.Second

if checkInterval < MinIdleCheckInterval {
    checkInterval = DefaultIdleCheckInterval
}
```
Or at minimum, add a warning comment in `values.yaml` documenting the minimum sensible value for production use.

andrii-i (dismissed):
Looks good, the CRD / API definitions make sense. 

Please find one field-doc accuracy suggestion below.

internal/controller/idle_checker.go
JGuinegagne: Yes, let me add a 1s minimum.

internal/controller/idle_detectors.go
JGuinegagne: Really, it doesn't reuse the client? Boy that's bad, let me confirm this.

Let's worry about in-cluster TLS later, it's a whole story on its own.

api/v1alpha1/workspace_types.go
JGuinegagne: thanks, stale comment, will update!

earaghbidikashani:
Overall CRD changes look good in workspace

QQ: What about workspace template?  `HTTPGetAction` is potentially something that admins would want to control and control data scienstists access to modify

Also curious to see if this is compatible with Grafana for a GPU based idle detection

internal/controller/idle_detectors.go
```diff
@@ -112,48 +118,190 @@ func (h *HTTPGetDetector) CheckIdle(ctx context.Context, workspaceName string, p
...
+		return &IdleCheckResult{IsIdle: false, ShouldRetry: false}, fmt.Errorf("httpGet config is nil")
+	}
+
+	scheme := string(httpGetConfig.Scheme)
+	if scheme == "" {
+		scheme = "http"
+	}
+	port := httpGetConfig.Port.String()
+	fullPath := resolveIdlePath(workspace.Status.ApplicationBasePath, httpGetConfig.Path)
+	url := fmt.Sprintf("%s://%s:%s%s", scheme, host, port, fullPath)
```
andrii-i: The probe URL is built by concatenating `host` with the workspace-controlled `httpGet.path` (`fmt.Sprintf("%s://%s:%s%s", ...)`), and `path` has no leading-slash validation. Two consequences:

1. Host-pinning bypass: when `applicationBasePath` is empty (the default / no-access-strategy case), `resolveIdlePath` returns `path` verbatim, so `path: "@10.0.0.5/x"` yields `http://<clusterip>:8888@10.0.0.5/x`, which Go parses with `Host=10.0.0.5` (the ClusterIP collapses into userinfo). A workspace editor could steer the operator's GET to any host it can reach. It's blind and GET-only with no headers (so no IMDSv2), so the practical impact is limited, but it defeats the ClusterIP pinning this code intends.
2. Malformed URL: `path: "api/status"` (no leading slash) yields `http://<host>:8888api/status`, which fails to parse, returns `ShouldRetry:false`, and silently disables idle shutdown for that workspace.

Both go away if the URL is built from components instead of concatenated, e.g. `(&url.URL{Scheme: scheme, Host: net.JoinHostPort(host, port), Path: fullPath}).String()`, optionally plus a CRD `Pattern=^/` on the path. Same builder is at `:90` (podExec) and the join is in `resolveIdlePath` (`:294`).

dist/chart/templates/manager/manager.yaml
```diff
@@ -79,6 +79,7 @@ spec:
         {{- if .Values.controller.plugins }}
         - "--plugin-endpoints={{ range $i, $p := .Values.controller.plugins }}{{ if $i }},{{ end }}{{ $p.name }}=http://localhost:{{ $p.port }}{{ end }}"
         {{- end }}
+        - "--idle-check-interval={{ .Values.idleShutdown.checkInterval }}"
```
andrii-i, line 83: This renders `--idle-check-interval` unconditionally, unlike the sibling duration flags just above (`--jwt-ttl`, `--new-key-use-delay`) which are guarded with `{{- if }}`. If an operator clears `idleShutdown.checkInterval` or sets it without a unit (e.g. `5`), the flag renders as `--idle-check-interval=` or `=5`, `flag.Parse` (ExitOnError) fails, and the whole controller crashloops before the `<=0` fallback in `NewWorkspaceIdleChecker` can apply. Could this be guarded like its siblings so an empty value falls back to the default? `{{- if .Values.idleShutdown.checkInterval }}` (the real edit is in `hack/apply-helm-patches.sh`, then regenerate).

andrii-i (commented):
A couple of other things worth looking at from a second pass. Details inline.

JGuinegagne:
> Overall CRD changes look good in workspace

> QQ: What about workspace template? `HTTPGetAction` is potentially something that admins would want to control and control data scienstists access to modify

That's covered, the template type for `idleShutdown` config inherits the workspace type.

> Also curious to see if this is compatible with Grafana for a GPU based idle detection

I don't know about that one, this is too vague. Do you have specifics on how this is implemented, and in that case we should have an issue that track E2E testing it.

internal/controller/idle_detectors.go
JGuinegagne: thanks Andrii, I will migrate to use `url.URL{}` utility.

Not adding `Pattern=^/` because that would override the `corev1.HTTPGetAction` shape.

internal/controller/idle_detectors.go
JGuinegagne: confirmed, very nice catch! I fixed the issue, thanks @bhanderig

andrii-i (approved):
Thank you @JGuinegagne.

bhanderig (approved):

PR #417, opened 2026-06-23
feat: workspace readiness probe

### LINKED ISSUE #416: Workspace pods receive traffic before JupyterLab is listening → transient 502s
## Summary

When a workspace is freshly created, the first requests routed to it can fail
with `502 Bad Gateway`. The workspace pod's IP joins the Service endpoints as
soon as the container process starts, but before the JupyterLab server has
bound its port. Traefik then tries to proxy to a port that isn't accepting
connections yet and returns a 502.

## Evidence

Observed during an EKS OIDC end-to-end run. Traefik logs the failure as a
backend dial error, not an auth-layer error:

```
502 Bad Gateway  error="dial tcp <pod-ip>:8888: connect: connection refused"
```

The matching workspace pod's logs show JupyterLab only finished binding its
port several seconds after the request arrived; the auth components
(oauth2-proxy, authmiddleware, Dex) were healthy throughout. The container has
**no readiness probe**, so the pod becomes an eligible Service endpoint the
moment it starts — well before the application is actually serving.

## Impact

- Transient and self-correcting: a retry a few seconds later succeeds.
- Surfaces as 502s to any client that hits a workspace immediately after
  creation (including users clicking through right after "create").
- Adds noise to e2e runs, which currently paper over it with client-side retries.

## Recommended fix

Gate the workspace pod from receiving traffic until the JupyterLab server is
actually accepting connections, rather than as soon as the container starts.

Note this is not a gap in the existing `accessStartupProbe` — that probe works
and does its job: it confirms the access *route* is wired up. But it
necessarily stops at the AuthN/AuthZ boundary (the operator cannot authenticate
through it), so it cannot observe whether the container itself has come up and
bound its port. We need an additional form of readiness — likely a
container-level readiness signal — to confirm the backend is actually serving
before the pod is marked available and made eligible to receive proxied traffic.

This eliminates the race at its source and lets the platform return a proper
"not ready yet" response instead of a hard 502 against a refused connection.

We intentionally leave the specific mechanism to the maintainers.

JGuinegagne:
This PR adds  `Workspace.spec.ReadinessProbe` field so that the workspace deployment only identify a pod as `ready` when it has passed the probe, which in turns translate to the workspace controller to wait for the probe to pass before setting the `workspace.status.condition[Available]` to `True`. Also add `defaultReadinessProbe` field in `WorkspaceTemplate.spec`.

Closes: #416 

### Workspace CRD field
```yaml
spec:
  readinessProbe:
    tcpSocket:
      port: 8888
    initialDelaySeconds: 2
    periodSeconds: 3
    failureThreshold: 30
```

### WorkspaceTemplate CRD field
```yaml
spec:
  defaultReadinessProbe:
    tcpSocket:
      port: 8888
    initialDelaySeconds: 2
    periodSeconds: 3
    failureThreshold: 30
```

### Implementation
- reuses `corev1.Probe` for `workspace.spec.readinessProbe` and `workspaceTemplate.spec.defaultReadinessProbe`
- controller injects `readinessProbe` to `deployment.spec.template.spec.containers[0].readinessProbe`
- workspace webhook uses `workspaceTemplate.spec.defaultReadinessProbe` field of referenced template when `workspace.spec.readinessProbe` is not set

### Testing
- add E2E test case for workspace w/ readinessProbe (TCP probe) for `jupyterlab-uv` image
- add E2E test cases for workspace inheritance of readinessProbe set in the template


bhanderig (approved):
LGTM

internal/controller/deployment_builder.go
```diff
@@ -252,7 +252,10 @@ func (db *DeploymentBuilder) buildPrimaryContainer(workspace *workspacev1alpha1.
 			},
 		},
 		Resources: resources,
-		// TODO: Add probes
+		// ReadinessProbe gates the pod from Service endpoints until the IDE is
+		// actually accepting connections, preventing transient 502s on startup.
+		// nil when neither workspace nor template configures one (opt-in via config).
+		ReadinessProbe: workspace.Spec.ReadinessProbe,
```
earaghbidikashani, line 258: we can potentially add startup probe and liveness probe in the same fashion

test/e2e/workspace_readiness_probes_test.go
```diff
@@ -0,0 +1,83 @@
...
+})
+
+func deleteResourcesForReadinessProbeTest() {
+	By("cleaning up workspaces")
+	cmd := exec.Command("kubectl", "delete", "workspace", "--all", "-n", "default",
+		"--ignore-not-found", "--wait=true", "--timeout=120s")
+	_, _ = utils.Run(cmd)
+
+	By("cleaning up standalone PVCs")
+	cmd = exec.Command("kubectl", "delete", "pvc", "--all", "-n", "default",
```
earaghbidikashani, line 77: I forgot we deleted the resource on every test suit. We should consider parallelizing these

api/v1alpha1/workspacetemplate_types.go
```diff
@@ -137,6 +137,12 @@ type WorkspaceTemplateSpec struct {
 	// +optional
 	DefaultLifecycle *corev1.Lifecycle `json:"defaultLifecycle,omitempty"`
 
+	// DefaultReadinessProbe specifies the default readiness probe for the main workspace
+	// container for workspaces using this template.
+	// Applied only if the workspace does not specify its own readiness probe.
+	// +optional
+	DefaultReadinessProbe *corev1.Probe `json:"defaultReadinessProbe,omitempty"`
```
earaghbidikashani, line 144: Should we consider adding probe validation at admission webhook level? If it's a valid probe?

earaghbidikashani (approved):
LGTM, I don't think we have a pattern of validating fields themselves (like validating if the secondary storage passed in is a valid volume) but it would be pretty easy to do for probes. We can add it in the future if necessary 

Let's discuss adding default readiness probes for the default juptyer and code editor templates in aws helm chart

internal/controller/deployment_builder.go
JGuinegagne: we could, but let's apply yagni here!

test/e2e/workspace_readiness_probes_test.go
JGuinegagne: could you add an issue to track this?

api/v1alpha1/workspacetemplate_types.go
JGuinegagne: Same thing, unless we have a clear case of needing it, I don't think we should add it now. In my mind, end-users shouldn't have any incentive to mess w/ this field and therefore defaulting ought to be enough. We could add an `allowCustomReadinessProbe` boolean flag, but again that increase our testing surface for no immediate customer benefit

PR #418, opened 2026-06-24
feat: add WorkspaceIntegrationTemplate CRD and integration API types

HariniNarayanan:
## New types (`api/v1alpha1`)

- **`WorkspaceIntegrationTemplate`** — admin-authored blueprint (`displayName`, `parameters`, `resourceRefs`, `shareProcessNamespace`, `deploymentModifications`, `statusProbe`). Controller-less, no status subresource. Mirrors `WorkspaceTemplate`. `resourceRefs` use the KRO external-reference shape (name handle + `apiVersion`/`kind` + `metadata`).
  - **`spec.parameters[]`** (name only) — the **declared parameter contract**: the names the template consumes via `{{ .Parameters.X }}`. It's the single source of truth both webhooks validate against (an undeclared reference is an author error; a workspace must supply every declared parameter). All declared parameters are required — no optionals/defaults. Mirrors the `EnvRequirements` declared-contract pattern on `WorkspaceTemplate`.
- **`Workspace.spec.integrationTemplateRefs[]`** (`MaxItems=1`) — the integration a workspace attaches: a `templateRef` (name/namespace) plus the supplied `parameters` (name/value).
- **`Workspace.status.resolvedIntegrations[]`** — the operator's private **freeze store** for the values-in-status design. Each entry holds the frozen resolution `values` plus the two re-resolve triggers:
  - `parametersVersion` — hash of the templateRef identity + user parameters (changes when the user switches clusters or edits a parameter).
  - `observedIntegrationTemplateVersion` — the resolved template's `UID.Generation` (changes when the admin edits/replaces the template).

  Re-resolution fires when **either** trigger changes; external drift in the referenced resource touches neither, so it never re-resolves and never rolls the pod.
- **`Workspace.status.integrationStatuses[]`** — operator-observed per-integration readiness (`name` + coarse `state` + `conditions`), following the KRO instance-status shape and the k8s `pod.status.containerStatuses` naming convention.

This is **pure API surface** — no controller logic. The state/condition value constants live in the controller package (added in the status-probe PR), mirroring how the `Workspace` type keeps its vocabulary in `internal/controller`. There is **no** `WorkspaceIntegration` child CRD — the old admission-child design is replaced by the values-in-status freeze.

## Files

- Hand-authored: `api/v1alpha1/workspace_types.go`, `api/v1alpha1/workspaceintegrationtemplate_types.go`.
- Generated (controller-gen) + dist-chart sync — **reviewers can skip these**, reproduce with `make manifests`: `api/v1alpha1/zz_generated.deepcopy.go`, the CRD YAML under `config/crd/bases/`, and the chart copies under `dist/chart/templates/crd/`.

## Testing

`go build` / `go vet` / unit suite green; `make manifests` reproduces the committed CRDs with no drift.

First commit of a stacked series — kept small for quicker review.

Workspace CRD - from live HyperPod cluster:
```
kind: Workspace
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"workspace.jupyter.org/v1alpha1","kind":"Workspace","metadata":{"annotations":{},"name":"ray-demo-workspace","namespace":"ray-hp-demo"},"spec":{"desiredStatus":"Running","image":"public.ecr.aws/sagemaker/sagemaker-distribution:latest-cpu","integrationTemplateRefs":[{"name":"ray-integration","parameters":[{"name":"rayClusterName","value":"demo-cluster"}]}],"resources":{"limits":{"cpu":"1","memory":"8Gi"},"requests":{"cpu":"250m","memory":"8Gi"}}}}
    workspace.jupyter.org/created-by: arn:aws:sts::659069595882:assumed-role/Admin/harinnar-Isengard
    workspace.jupyter.org/last-updated-by: system:serviceaccount:jupyter-k8s-system:jupyter-k8s-controller-manager
  creationTimestamp: "2026-07-15T07:05:54Z"
  finalizers:
  - workspace.jupyter.org/workspace-protection
  generation: 1
  labels:
    workspace.jupyter.org/access-strategy-name: hyperpod-access-strategy
    workspace.jupyter.org/access-strategy-namespace: jupyter-k8s-system
    workspace.jupyter.org/template-name: sagemaker-jupyter-template
    workspace.jupyter.org/template-namespace: jupyter-k8s-system
  name: ray-demo-workspace
  namespace: ray-hp-demo
  resourceVersion: "95222392"
  uid: 61ffe4ae-bb85-4b69-b95a-e2e0578dc90e
spec:
  accessStrategy:
    name: hyperpod-access-strategy
    namespace: jupyter-k8s-system
  accessType: OwnerOnly
  appType: jupyterlab
  containerConfig:
    command:
    - /opt/amazon/sagemaker/workspace/bin/entrypoint-workspace-jupyterlab
  desiredStatus: Running
  displayName: ""
  image: public.ecr.aws/sagemaker/sagemaker-distribution:latest-cpu
  integrationTemplateRefs:
  - name: ray-integration
    parameters:
    - name: rayClusterName
      value: demo-cluster
  ownershipType: OwnerOnly
  podSecurityContext:
    fsGroup: 1000
  resources:
    limits:
      cpu: "1"
      memory: 8Gi
    requests:
      cpu: 250m
      memory: 8Gi
  serviceAccountName: default
  storage:
    mountPath: /home/sagemaker-user
    size: 5Gi
    storageClassName: sagemaker-spaces-default-storage-class
  templateRef:
    name: sagemaker-jupyter-template
    namespace: jupyter-k8s-system
status:
  conditions:
  - lastTransitionTime: "2026-07-15T07:11:04Z"
    message: Workspace is ready
    reason: ResourcesReady
    status: "True"
    type: Available
  - lastTransitionTime: "2026-07-15T07:11:04Z"
    message: Workspace is ready
    reason: ResourcesReady
    status: "False"
    type: Progressing
  - lastTransitionTime: "2026-07-15T07:11:04Z"
    message: No errors detected
    reason: NoError
    status: "False"
    type: Degraded
  - lastTransitionTime: "2026-07-15T07:11:04Z"
    message: Workspace is running
    reason: DesiredStateRunning
    status: "False"
    type: Stopped
  - lastTransitionTime: "2026-07-15T07:11:04Z"
    message: Workspace is running
    reason: DesiredStateRunning
    status: "False"
    type: Deleting
  deploymentName: workspace-ray-demo-workspace
  integrationStatuses:
  - conditions:
    - lastTransitionTime: "2026-07-15T15:22:11Z"
      message: ""
      reason: Ready
      status: "True"
      type: Ready
    name: ray-integration
    state: Ready
  resolvedIntegrations:
  - name: ray-integration
    observedIntegrationTemplateVersion: 620e769b-4213-496e-96c8-8c4b362d7058.3
    parametersVersion: b3c5f3115d3176eaf7a78353501116b6
    values:
      rayCluster|{.spec.headGroupSpec.template.spec.containers[0].image}: public.ecr.aws/sagemaker/sagemaker-distribution:latest-cpu
      rayCluster|{.status.head.serviceName}: demo-cluster-head-svc
  serviceName: workspace-ray-demo-workspace-service

```

WorkspaceIntegrationTemplate
```
kind: WorkspaceIntegrationTemplate
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"workspace.jupyter.org/v1alpha1","kind":"WorkspaceIntegrationTemplate","metadata":{"annotations":{},"name":"ray-integration","namespace":"ray-hp-demo"},"spec":{"deploymentModifications":{"podModifications":{"additionalContainers":[{"args":["echo ray start --address={{ resource \"rayCluster\" \"{.status.head.serviceName}\" }}:6379 --block \u0026\u0026 sleep infinity"],"command":["/bin/sh","-c"],"image":"{{ resource \"rayCluster\" \"{.spec.headGroupSpec.template.spec.containers[0].image}\" }}","name":"ray-sidecar"}],"primaryContainerModifications":{"mergeEnv":[{"name":"RAY_HEAD_SERVICE","valueTemplate":"{{ resource \"rayCluster\" \"{.status.head.serviceName}\" }}"},{"name":"RAY_CLUSTER_NAME","valueTemplate":"{{ .Parameters.rayClusterName }}"}]}}},"displayName":"Ray Cluster Integration (HyperPod demo)","parameters":[{"name":"rayClusterName"}],"resourceRefs":[{"apiVersion":"ray.io/v1","kind":"RayCluster","metadata":{"name":"{{ .Parameters.rayClusterName }}","namespace":"{{ .Workspace.Namespace }}"},"name":"rayCluster"}],"shareProcessNamespace":true,"statusProbe":{"exec":{"command":["ray","status","--address","{{ resource \"rayCluster\" \"{.status.head.serviceName}\" }}:6379"]},"periodSeconds":5}}}
  creationTimestamp: "2026-07-15T07:03:23Z"
  generation: 3
  name: ray-integration
  namespace: ray-hp-demo
  resourceVersion: "95222308"
  uid: 620e769b-4213-496e-96c8-8c4b362d7058
spec:
  deploymentModifications:
    podModifications:
      additionalContainers:
      - args:
        - echo ray start --address={{ resource "rayCluster" "{.status.head.serviceName}"
          }}:6379 --block && sleep infinity
        command:
        - /bin/sh
        - -c
        image: '{{ resource "rayCluster" "{.spec.headGroupSpec.template.spec.containers[0].image}"
          }}'
        name: ray-sidecar
      primaryContainerModifications:
        mergeEnv:
        - name: RAY_HEAD_SERVICE
          valueTemplate: '{{ resource "rayCluster" "{.status.head.serviceName}" }}'
        - name: RAY_CLUSTER_NAME
          valueTemplate: '{{ .Parameters.rayClusterName }}'
  displayName: Ray Cluster Integration (HyperPod demo)
  parameters:
  - name: rayClusterName
  resourceRefs:
  - apiVersion: ray.io/v1
    kind: RayCluster
    metadata:
      name: '{{ .Parameters.rayClusterName }}'
      namespace: '{{ .Workspace.Namespace }}'
    name: rayCluster
  shareProcessNamespace: true
  statusProbe:
    exec:
      command:
      - ray
      - status
      - --address
      - '{{ resource "rayCluster" "{.status.head.serviceName}" }}:6379'
    periodSeconds: 30

````

### CODE REVIEW

File: api/v1alpha1/workspaceintegration_types.go [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+// A WorkspaceIntegration is the per-integration INSTANCE of a WorkspaceIntegrationTemplate,
+// analogous to how a Workspace is the instance of a WorkspaceTemplate. This mirrors the
+// Workspace/WorkspaceTemplate precedent exactly: the instance reuses the template's own field
+// types (DeploymentModifications, ShareProcessNamespace, StatusProbe) verbatim, with every
+// template expression already substituted to a literal value (no {{ }} remains). There is no
+// bespoke "resolved" envelope -- just as Workspace.spec.image carries the resolved value of
+// WorkspaceTemplate.spec.defaultImage, these fields carry the resolved values of the template's
+// corresponding fields.
+//
+// TemplateRef + WorkspaceRef are the INPUTS, written by the workspace controller when it creates
```
psnilesh: Discussed offline; the core motivation here is to freeze the rendered `WorkspaceIntegrationTemplate` at the time of Workspace admission (through a mutating webhook). Any modifications to resources referenced by the `WorkspaceIntegrationTemplate` after initial Workspace admission would not automatically reconcile. Any resource modifications that break the integration will be surfaced by integration status.

config/samples/workspace_integration_template_ray.yaml
```diff
@@ -0,0 +1,217 @@
...
+  # Each entry has an `id` handle, referenced below as {{ resource "<id>" "<jsonpath>" }}.
+  resourceRefs:
+    - id: rayCluster                      # handle used in template expressions below
+      apiVersion: ray.io/v1               # KubeRay RayCluster API group/version
+      kind: RayCluster
+      name: "{{ .Parameters.rayClusterName }}"        # customer-supplied on the workspace
+      # Namespace of the RayCluster. The UI pre-fills rayClusterNamespace with the workspace's own
+      # namespace (the HyperPod single-tenant-namespace default), but the customer can point at a
+      # RayCluster in another namespace by overriding this parameter.
+      namespace: "{{ .Parameters.rayClusterNamespace }}"
```
psnilesh: Let us double check how much flexibility we need in terms of namespaces with PM once again.

config/samples/workspace_integration_template_ray.yaml
```diff
@@ -0,0 +1,217 @@
...
+  # integrationRefs is a list, so a workspace can attach several templates; each entry resolves and
+  # applies independently. Parameters are a name/value list (extensible per-entry; values are strings
+  # interpolated into the template's expressions). The UI pre-fills rayClusterNamespace with the
+  # workspace's namespace.
+  integrationRefs:
+    - name: ray-integration
+      parameters:
+        - name: rayClusterName
+          value: demo-cluster
+        - name: rayClusterNamespace
```
psnilesh: Don't fully recollect if we discussed RBAC yet but I reckon we will need some [SubjectAccessReviews](https://kubernetes.io/docs/reference/kubernetes-api/definitions/subject-access-review-v1-authorization/) before admission to ensure the creator of Workspace has permissions to access these RayClusters to avoid privilege escalation.  Lets follow up on this.

api/v1alpha1/workspace_types.go
```diff
@@ -232,6 +232,63 @@ type WorkspaceSpec struct {
...
+	// Namespace where the WorkspaceIntegrationTemplate is located
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// Parameters provided by the user for template expression resolution.
+	// Each parameter is referenced in template expressions as {{ .Parameters.<Name> }}.
+	// Names must be unique.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
```
JGuinegagne, line 278: should we limit the size of this list? Maybe to 10 to start?

api/v1alpha1/workspace_types.go
```diff
@@ -327,6 +394,36 @@ type WorkspaceStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
+// IntegrationStatus reports the operator-observed readiness of a single integration.
+// Follows the k8s condition idiom (ready + reason + message) rather than a phase enum.
+type IntegrationStatus struct {
```
JGuinegagne, line 442: just wondering: how about following KRO [precedent](https://kro.run/docs/concepts/instances#understanding-status) for status reporting:

```yaml
name: string
state: string
conditions: metav1.Condition
```

Then we can use standard k8s condition:
- `Ready`
- `ResourcesReady`
- `Degraded`

api/v1alpha1/workspaceintegration_types.go
```diff
@@ -0,0 +1,112 @@
+/*
```
JGuinegagne: wait...what is this custom resource?

That would mean 1:i relationship w/ workspace right, as in 1 workspace _has many_ workspace integration resources.

I don't think we need it at all, why can't the workspace reconciliation loop handle this?

api/v1alpha1/workspaceintegrationtemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+	ID string `json:"id"`
+
+	// APIVersion of the target resource (e.g., "ray.io/v1")
+	APIVersion string `json:"apiVersion"`
+
+	// Kind of the target resource (e.g., "RayCluster")
+	Kind string `json:"kind"`
+
+	// Name supports template expressions: {{ .Workspace.Name }}, {{ .Parameters.X }}
+	Name string `json:"name"`
```
JGuinegagne, line 39: per KRO [precedent](https://kro.run/docs/concepts/rgd/resource-definitions/external-references), consider:
```yaml
name: string
apiVersion: string
kind: string
metadata: {
  name: string
  namespace: string
}
```

api/v1alpha1/workspaceintegrationtemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// +kubebuilder:validation:MaxLength=253
+	DisplayName string `json:"displayName"`
+
+	// ResourceRefs defines the resources to fetch at resolution time. Each entry is
+	// addressed in template expressions by its id: {{ resource "<id>" "<jsonpath>" }}.
+	// +optional
+	// +listType=map
+	// +listMapKey=id
+	// +kubebuilder:validation:MaxItems=1
+	ResourceRefs []ResourceRef `json:"resourceRefs,omitempty"`
```
JGuinegagne: 1. that should NOT be optional, omit empty, right?
2. min size should be 1, shouldn't it?
3. max size should be 1 (for now)

api/v1alpha1/workspaceintegrationtemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// and environment); injected containers should run as non-root.
+	// +optional
+	ShareProcessNamespace *bool `json:"shareProcessNamespace,omitempty"`
+
+	// DeploymentModifications defines modifications to apply to workspace deployments
+	// with template expressions in string fields
+	// +optional
+	DeploymentModifications *DeploymentModifications `json:"deploymentModifications,omitempty"`
+
+	// StatusProbe is an optional, admin-authored probe the operator runs periodically
```
JGuinegagne: remove the `admin-authored` please; we shouldn't leak our interpretation of who should use the construct to the API definition. Such "usage recommendations" belong to the documentation, [example](https://jupyter-k8s.readthedocs.io/en/latest/concepts/access-strategies/index.html)

config/samples/workspace_integration_template_ray.yaml
```diff
@@ -0,0 +1,217 @@
...
+      cpu: "2"
+      memory: "8Gi"
+  # All the customer configures: the integration template name + which Ray cluster (and its namespace).
+  # integrationRefs is a list, so a workspace can attach several templates; each entry resolves and
+  # applies independently. Parameters are a name/value list (extensible per-entry; values are strings
+  # interpolated into the template's expressions). The UI pre-fills rayClusterNamespace with the
+  # workspace's namespace.
+  integrationRefs:
+    - name: ray-integration
+      parameters:
```
JGuinegagne: how are these created? do we assume they exist in the cluster?

config/samples/workspace_integration_template_ray.yaml
```diff
@@ -0,0 +1,217 @@
...
+kind: Workspace
+metadata:
+  labels:
+    app.kubernetes.io/name: jupyter-k8s
+    app.kubernetes.io/managed-by: kustomize
+  name: alice-notebook
+spec:
+  displayName: "Alice's Notebook"
+  # Use the SAME image as the RayCluster head/workers so Ray + Python versions match exactly.
+  image: "public.ecr.aws/sagemaker/sagemaker-distribution:latest-cpu"
```
JGuinegagne: do you need to use SMD?
We generally try to stay away from aws-specific images in the OSS repo.

JGuinegagne (changes requested):
Primary concern is the `WorkspaceIntegration` as a separate resource. Hoping that won't be needed.

config/samples/workspace_integration_template_ray.yaml
HariniNarayanan: Yes - we assume they exist in the cluster

config/samples/workspace_integration_template_ray.yaml
HariniNarayanan: This is a sample template - I can change this to a more generic image

api/v1alpha1/workspaceintegrationtemplate_types.go
HariniNarayanan: Sure will do

api/v1alpha1/workspaceintegration_types.go
HariniNarayanan: I need a place to persist the resolved data from the resource so it is not reconciled at every deployment - WorkspaceIntegration is to solve that. Else I’ll have to persist it in the Workplace CR itself

Workspace reconciliation would end up pulling the resource data again if I don't freeze it

api/v1alpha1/workspace_types.go
HariniNarayanan: Can do - I'll update.

api/v1alpha1/workspace_types.go
```diff
@@ -232,6 +232,63 @@ type WorkspaceSpec struct {
...
+	// webhook then resolves each WorkspaceIntegration against its live referenced resources at the
+	// child's admission and freezes the literal result onto the child's own spec output fields
+	// (deploymentModifications/statusProbe/shareProcessNamespace). The workspace Deployment is built
+	// solely from those frozen children; the workspace controller never reads the
+	// WorkspaceIntegrationTemplate or the referenced resources at reconcile time.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=1
+	IntegrationRefs []IntegrationTemplateRef `json:"integrationRefs,omitempty"`
```
henrywa2: For my understanding, why did we name this field `IntegrationRefs` when it takes in an integration template ref and not an integration? At first glance users might assume this takes in a `WorkspaceIntegration` since that CRD also exists. It also seems different from the existing pattern of passing in a workspace template in the `TemplateRef` field.

config/samples/workspace_integration_template_ray.yaml
HariniNarayanan: Verified that we do want the flexibility across namespaces with Shreyas

config/samples/workspace_integration_template_ray.yaml
HariniNarayanan: We can't add this here Nilesh, the token review should happen on the ray cluster side - we need a role + role binding. I'll create add helm chart here for reference: https://github.com/aws/sagemaker-hyperpod-cli/tree/main/helm_chart/HyperPodHelmChart. Would look like https://raw.githubusercontent.com/ray-project/kuberay/refs/heads/master/ray-operator/config/samples/ray-cluster.kubernetes.auth.yaml

api/v1alpha1/workspace_types.go
HariniNarayanan: I've hard-coded this to 1 like we discussed. Limit - 1 integration with 1 resource, with max 10 parameters

api/v1alpha1/workspace_types.go
HariniNarayanan: Updated

henrywa2 (dismissed):

henrywa2 (approved):

api/v1alpha1/workspace_types.go
```diff
@@ -232,6 +232,62 @@ type WorkspaceSpec struct {
...
 	// +optional
 	InitContainers []corev1.Container `json:"initContainers,omitempty"`
+
+	// IntegrationTemplateRefs attaches one or more WorkspaceIntegrationTemplates that inject runtime
+	// capabilities (sidecars, volumes, env vars) into the workspace pod with template-based
+	// dynamic resolution. Each entry is the user's REQUEST only -- a reference to a
+	// WorkspaceIntegrationTemplate plus parameters -- never the resolved sidecars.
+	//
+	// The workspace controller resolves each integration against its referenced resources (e.g. a
+	// RayCluster) only when the input token -- hash(templateRef + parameters) -- changes. On an
```
JGuinegagne, line 242: non-blocking: let's keep this docstring generic please: remove the `e.g. a RayCluster`

api/v1alpha1/workspace_types.go
```diff
@@ -325,6 +381,80 @@ type WorkspaceStatus struct {
...
+	// +listMapKey=name
+	ResolvedIntegrations []ResolvedIntegration `json:"resolvedIntegrations,omitempty"`
+}
+
+// ResolvedIntegration is the frozen resolution record for a single attached integration. It is the
+// source of truth the controller replays from on unchanged-token reconciles.
+type ResolvedIntegration struct {
+	// Name is the WorkspaceIntegrationTemplate name this record resolves (matches the
+	// spec.integrationTemplateRefs[].name that produced it).
+	Name string `json:"name"`
```
JGuinegagne, line 414: sanity-check: do we envision a future where we can have:
```yaml
integrationRefs:
  - name: RayCluster
    namespace: A
  - name: RayCluster
    namespace: B
```
I _think_ this is disallowed because you use `listMapKey=name`, but could you confirm?

api/v1alpha1/workspace_types.go
```diff
@@ -325,6 +381,80 @@ type WorkspaceStatus struct {
...
+// KRO instance-status precedent (https://kro.run/docs/concepts/instances#understanding-status):
+// a name, a coarse state, and standard Kubernetes conditions carrying the machine-readable detail.
+type IntegrationStatus struct {
+	// Name is the name of the WorkspaceIntegrationTemplate this status reports on.
+	Name string `json:"name"`
+
+	// State is a coarse, human-facing rollup of the conditions below: "Ready" when the integration's
+	// last probe succeeded, "Degraded" when it failed. Empty means "not yet probed".
+	// +optional
+	State string `json:"state,omitempty"`
```
JGuinegagne, line 449: This is following KRO's [precedent](https://kro.run/docs/concepts/instances#understanding-status), right?
In any case I think it's fine, but just wanted to double confirm.

api/v1alpha1/workspaceintegrationtemplate_types.go
```diff
@@ -0,0 +1,172 @@
...
+	// command always runs in the workspace container (the operator fixes this; it is not
+	// author-selectable) so it sees the pod's real network and auth context, catching
+	// data-plane failures a control-plane status check would miss (e.g. a Ray GCS reconnect
+	// loop while .status still reads ready). This is the only transport implemented today.
+	// +optional
+	Exec *corev1.ExecAction `json:"exec,omitempty"`
+
+	// How often (in seconds) the operator performs the probe. Default: operator poll cadence.
+	// +optional
+	PeriodSeconds int32 `json:"periodSeconds,omitempty"`
```
JGuinegagne, line 137: non-blocking: consider removing this, and make the recheck status an operator parameter.

Otherwise we might get problems later: what if there are 10 `IntegrationTemplateRefs`, each w/ a different period. When do you requeue reconciliation, and do you risk doing it too often?

JGuinegagne: Here is the operator [precedent](https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/cmd/main.go#L193) for idleCheck. We expose it mostly to support E2E testing it!

JGuinegagne (approved):
LGTM, thank you for iterating Harini.

Few non-blocking comments.
Specifically consider parametrizing integration status checks at the operator level to avoid conflicting instructions.

api/v1alpha1/workspace_types.go
```diff
@@ -325,6 +381,80 @@ type WorkspaceStatus struct {
...
+// source of truth the controller replays from on unchanged-token reconciles.
+type ResolvedIntegration struct {
+	// Name is the WorkspaceIntegrationTemplate name this record resolves (matches the
+	// spec.integrationTemplateRefs[].name that produced it).
+	Name string `json:"name"`
+
+	// ParametersVersion is a hash of the integration ref's identity and user-supplied parameters
+	// (templateRef namespace+name + the sorted parameter map) captured when these values were
+	// resolved. A change means the user switched clusters or edited a parameter.
+	ParametersVersion string `json:"parametersVersion"`
```
psnilesh, line 419: nit: I don't know if we can call this a version. Should we call it `ParametersHash`?

HariniNarayanan: Fair, updated.

api/v1alpha1/workspace_types.go
HariniNarayanan: Yes - following the precedent.

api/v1alpha1/workspace_types.go
HariniNarayanan: Correct. integrationTemplateRefs is +listType=map + +listMapKey=name, so name is the unique key and the API server rejects two entries sharing a name regardless of namespace - and we cannot have two ray clusters with the same name in the same namespace.

api/v1alpha1/workspace_types.go
HariniNarayanan: Fixed.

api/v1alpha1/workspaceintegrationtemplate_types.go
HariniNarayanan: That's fair, removing it. We can add it if we see a need for it.

HariniNarayanan:
Will raise another small PR to handle nit comments.

PR #419, opened 2026-06-25
feat: add integration template expression resolver

HariniNarayanan:
## Summary

Adds `IntegrationTemplateResolver` — a small, **pure** (no Kubernetes I/O, no shared state) component that resolves the template expressions used by `WorkspaceIntegrationTemplate`, plus the `ResourceValueProvider` seam that lets the same render path run in two modes: **live capture** and **frozen replay**.

This is **branch 2 of the integration stack** (values-in-status / "freeze" design). It is intentionally self-contained: no API reads, no opinion about where resolved output is stored.

> **Stacked PR.** Built on top of #418 (`feat/workspace-integration-api`). Please review/merge that first; the diff here is just the 2 resolver files.

## The `ResourceValueProvider` seam

The `{{ resource "<id>" "{jsonpath}" }}` function is backed by a `ResourceValueProvider` interface with two implementations — this is what makes the freeze design work:

| provider | behavior | used when |
|---|---|---|
| **live (capture)** — `NewLiveResourceValueProvider(resources)` | reads the referenced resource via JSONPath **and records** every resolved value under `CaptureKey(id, jsonPath)` | first resolve / re-resolve — the recorded map is frozen into `status.resolvedIntegrations` |
| **frozen (replay)** — `NewFrozenResourceValueProvider(values)` | serves values from a previously captured map, **never reads** the referenced resource, **hard-errors** on a missing key | every subsequent reconcile — replays frozen values so external drift never re-resolves |

Both key on `(id, jsonPath)` via `CaptureKey`, so a value captured live replays under the identical expression. A hard error on a missing frozen key (rather than a silent empty) is what forces a re-resolve instead of silently rendering an incomplete pod spec.

## Expression forms

All expressed as ordinary Go `text/template` constructs, resolved by `ResolveTemplateExpression` in one pass:

| Expression | Resolves to |
|---|---|
| `{{ .Workspace.Name }}` / `{{ .Workspace.Namespace }}` | the owning Workspace's identity |
| `{{ .Parameters.<name> }}` | a value supplied via the `integrationTemplateRef`'s parameters |
| `{{ resource "<id>" "{jsonpath}" }}` | a field read from a referenced resource, served by the provider above |

- Parsed with `Option("missingkey=error")` so a typo'd parameter is a hard error, never a silent empty.
- The `resource` function is registered via `template.FuncMap` (same mechanism the access-strategy renderer uses).
- JSONPath is evaluated with `k8s.io/client-go/util/jsonpath` — the same engine behind `kubectl -o jsonpath`. The path is passed as a **quoted argument** (`{{ resource "rayCluster" "{.status.head.serviceName}" }}`) because raw `{.field}` is not valid Go template syntax; wrapping it lets dotted paths and `[N]` indexing pass through untouched.

`ResolveResourceRef` resolves a ref's templated name/namespace (workspace/parameter expressions only) fail-closed; `ResolvePodModifications` walks a template's pod modifications applying the resolver to every string field.

## Files

Both hand-authored:
- `internal/controller/integration_template_resolver.go`
- `internal/controller/integration_template_resolver_test.go`

## Testing

`go build` / `go vet` / `gofmt` clean. 16 unit tests (all pure, no envtest) cover:
- live capture (records values), frozen replay (same output, no resource read), fail-closed on unknown id / missing frozen key
- workspace + parameter rendering, `missingkey=error` on a typo'd parameter, `ResolveResourceRef` templated name/namespace + empty-result guard
- `ResolvePodModifications` end-to-end across every templatable field (container image/command/args/workingDir/env/volumeMounts, primary mergeEnv + volumeMounts)
- **deep-copy mutation-safety** (the original CR data is never modified), the **pod-level-volumes-not-resolved** contract, and fail-closed error propagation from a nested field
- JSONPath `[N]` array indexing, plus invalid-path and empty-result error paths


internal/controller/integration_template_resolver.go
```diff
@@ -0,0 +1,286 @@
...
+// expression names an id with no fetched resource, the JSONPath is invalid, or the
+// JSONPath evaluates to an empty string.
+func (r *IntegrationTemplateResolver) ResolveField(field string, data IntegrationTemplateData, resources map[string]*unstructured.Unstructured) (string, error) {
+	tmpl, err := template.New("field").
+		Option("missingkey=error").
+		Funcs(template.FuncMap{
+			"resource": func(id, jsonPath string) (string, error) {
+				res, ok := resources[id]
+				if !ok || res == nil {
+					return "", fmt.Errorf("resource template references id %q but no such resource is available (id not declared in resourceRefs or not fetched)", id)
```
psnilesh: Should we consider adding some template validations at the time of admission to minimize the chance of async failures?

In a way, these error messages are sort of an Admin concern because they are responsible (from user PoV) for the template. Can admins discover these failures easily, or do they have to go to the operator logs?

I'm hoping we'll wrap these messages with the workspace information at the caller site before they are logged.

internal/controller/integration_template_resolver.go
```diff
@@ -0,0 +1,286 @@
...
+//     <id> in resourceRefs, via JSONPath. This is registered as a template function (like
+//     b32encode in the access strategy renderer), so the JSONPath string — including [N]
+//     indexing — is passed through untouched as a quoted function argument.
+//
+// resources maps each resourceRef id to its fetched object. Returns an error if: the
+// template syntax is invalid, a referenced parameter key is missing, a {{ resource }}
+// expression names an id with no fetched resource, the JSONPath is invalid, or the
+// JSONPath evaluates to an empty string.
+func (r *IntegrationTemplateResolver) ResolveField(field string, data IntegrationTemplateData, resources map[string]*unstructured.Unstructured) (string, error) {
+	tmpl, err := template.New("field").
```
psnilesh: nit: why `field`?

HariniNarayanan: updated this

internal/controller/integration_template_resolver.go
```diff
@@ -0,0 +1,367 @@
...
+	// resolution (fail-closed): the caller must not apply a partially resolved overlay.
+	Value(id, jsonPath string) (string, error)
+}
+
+// CaptureKey is the stable key under which a single {{ resource id "jsonPath" }} resolution is
+// recorded and later replayed. Using both the id and the jsonPath makes each captured value
+// self-describing and collision-free across resourceRefs.
+func CaptureKey(id, jsonPath string) string {
+	return id + "|" + jsonPath
+}
```
JGuinegagne, line 55: sanity-check: pipe (`|`) is disallowed from `id`, correct?

Could you add a unit test to verify that this rule is in place (not necessarily in this PR)?

JGuinegagne (approved):

internal/controller/integration_template_resolver.go
HariniNarayanan: Template validations (for admin) + additional workspace validations(for data scientist integrating) are handled in https://github.com/jupyter-infra/jupyter-k8s/pull/421

internal/controller/integration_template_resolver.go
HariniNarayanan: yes, it's 

	// Name is the handle used to reference this resource in template expressions:
	// {{ resource "<name>" "<jsonpath>" }}. Must be unique within resourceRefs.
	// +kubebuilder:validation:Pattern=`^[a-z][a-zA-Z0-9-]*$`
	// +kubebuilder:validation:MaxLength=63
	
Will add a UT in the next PR

PR #420, opened 2026-06-25
feat: add WorkspaceIntegration builder (admission-time freeze)

HariniNarayanan:
## Summary

Adds `BuildWorkspaceIntegration` — the admission-time function that resolves a `WorkspaceIntegration` against its referenced resources and template parameters, then freezes the fully-resolved, literal output directly onto the object's own spec fields (`deploymentModifications`, `shareProcessNamespace`, `statusProbe`).

It mirrors the existing `WorkspaceTemplate → Workspace` precedent: just as a `Workspace` carries its resolved template values on its own spec fields, a `WorkspaceIntegration` carries its resolved values, in place, on its own spec. There is no separate "resolved" envelope type. Once the output is frozen here, the workspace controller reads only the frozen child and never re-resolves the template at reconcile time.

## What this function does

Given a `WorkspaceIntegration` and its `WorkspaceIntegrationTemplate`, the builder:

1. Validates the co-location invariant — a `workspaceRef.namespace` that differs from the object's own namespace is rejected (cross-namespace references are unsupported); empty defaults to the object's own namespace.
2. Fetches each `resourceRef` target (e.g. a `RayCluster`) via the client, after resolving any template expressions in the ref's name/namespace.
3. Resolves every template expression in the template's pod modifications into literal values.
4. Commits the result onto the object's spec, assigning **every** output field unconditionally so a re-build (e.g. after switching the referenced cluster) never leaves a stale value behind.

**Fail-closed:** any referenced resource that is missing, not yet ready, or resolves to an empty value returns an error so admission rejects the write and the prior good output is preserved. On error, the object's spec fields are left untouched.

## Dependencies

> ⚠️ **This branch does not compile on its own** and is expected to show a red build until its dependencies merge. It is published now for review of the resolution/freeze logic in isolation.

This commit builds on two earlier commits in the same series and references symbols introduced by each:

- **The API types commit** — provides the `WorkspaceIntegration` / `WorkspaceIntegrationTemplate` types and their `deploymentModifications` / `shareProcessNamespace` / `statusProbe` spec fields.
- **The template-resolver commit** — provides `IntegrationTemplateResolver` (`ResolveResourceRef`, `ResolvePodModifications`) and the `IntegrationTemplateData` context. This commit is purely the *consumer* of that resolver; it adds the fetch-and-freeze orchestration around it.

Once both of those land on `main`, this branch rebases onto `main` cleanly and compiles. The PR can be rebased in place at that point — same branch, same PR.

## Why this is a separate commit

The resolver (previous commit) is a **pure function**: given inputs, it substitutes template expressions into literals. It performs no Kubernetes I/O and holds no opinion about *where* the output is stored or *when* resolution happens.

This commit adds exactly the orchestration the resolver deliberately omits:

- **Kubernetes reads** — fetching the referenced resources the resolver needs as inputs.
- **A storage decision** — freezing the resolved output onto the object's own spec fields.
- **Admission-time policy** — fail-closed semantics, the namespace co-location check, and unconditional re-write on update.

Keeping these concerns in their own commit means the resolver stays small, side-effect-free, and trivially unit-testable, while the fetch/freeze/policy layer is reviewed and tested on its own terms. The next commit (the mutating webhook) wires this function into the admission path; logging and event recording live there, not here.

## What's in this PR

| File | What |
|---|---|
| `internal/controller/workspaceintegration_builder.go` | `BuildWorkspaceIntegration` + the `fetchReferencedResources` helper. |
| `internal/controller/workspaceintegration_builder_test.go` | Unit tests against a fake client. |

The GVK lookup uses [`schema.FromAPIVersionAndKind`](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#FromAPIVersionAndKind) from apimachinery rather than a hand-rolled parser.

## Testing

Unit tests against a fake client (no control plane required), covering:

- Resolve-to-literal-values (env/sidecar fields hold resolved literals, no `{{ }}` remaining).
- Freezing `statusProbe` (deep-copied, not aliased) and `shareProcessNamespace`.
- Workspace-namespace defaulting to the object's own namespace.
- Fail-closed on a missing referenced resource (named in the error, output spec left untouched).
- Re-build clearing stale output from a prior resolution.
- Nil-input guards and cross-namespace `workspaceRef` rejection.

```
go test ./internal/controller/ -run 'TestBuildWorkspaceIntegration' -count=1
```

## References

- Go `text/template` (the expression engine the resolver uses): https://pkg.go.dev/text/template
- `client-go` JSONPath (resource-field reads): https://pkg.go.dev/k8s.io/client-go/util/jsonpath
- KubeRay / `RayCluster` (an example referenced resource): https://docs.ray.io/en/latest/cluster/kubernetes/index.html

### DIFF
--- a/internal/controller/workspaceintegration_builder.go
+++ b/internal/controller/workspaceintegration_builder.go
@@ -1,0 +1,172 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// BuildWorkspaceIntegration resolves a WorkspaceIntegration against the live referenced resources
+// and its templateRef parameters, freezing the resolved output directly onto the WorkspaceIntegration's
+// own spec fields (DeploymentModifications, ShareProcessNamespace, StatusProbe) with all template
+// expressions substituted to literal values. This mirrors how a Workspace carries its resolved
+// WorkspaceTemplate values on its own spec fields -- no bespoke "resolved" envelope.
+//
+// It is the admission-time entry point for the WorkspaceIntegration mutating webhook: the webhook
+// calls it once at the WorkspaceIntegration's own admission (CREATE/UPDATE). The workspace controller
+// never calls this at reconcile time -- it only reads the already-resolved child.
+//
+// Resolution is fail-closed: any resourceRef that cannot be resolved or fetched, or any template
+// expression that resolves to empty (e.g. a not-yet-ready RayCluster whose .status.head.serviceName
+// is blank), returns an error so admission rejects the request and the prior good state is
+// preserved. On error the wi spec fields are left unchanged.
+//
+// The template's deploymentModifications keeps its native PodModifications shape (MergeEnv stays
+// []AccessEnvTemplate, its ValueTemplate now holding a literal); the statusProbe is frozen as-is
+// (it carries no template expressions today).
+func BuildWorkspaceIntegration(
+	ctx context.Context,
+	c client.Client,
+	wi *workspacev1alpha1.WorkspaceIntegration,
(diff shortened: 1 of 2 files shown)


internal/controller/workspaceintegration_builder.go
```diff
@@ -0,0 +1,172 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package controller
```
psnilesh, line 6: Should this be in `internal/webhook` instead?

HariniNarayanan:
Superseded by the 2b values-in-status freeze split (#418/#419/#421/#423/#424/#425).

PR #421, opened 2026-06-25
feat: validate WorkspaceIntegrationTemplate and workspace integration parameters

HariniNarayanan:
## Summary

Admission-time **validation** for the integration feature — catches authoring and usage mistakes at the write that causes them, rather than letting them surface later as a degraded/unresolved workspace.

> **Stacked PR.** Built on the integration API-types and resolver PRs — review/merge those first.
>
> Note: this repurposes an earlier *mutating* webhook (which resolved `WorkspaceIntegration` spec output in the old admission-child design). There is no `WorkspaceIntegration` object and no mutation in the values-in-status design — this is now a pair of **validating** checks.

The template's **declared `spec.parameters`** is the single source of truth — both checks validate against it, so the parameter contract is defined in one place (no regex-scanning of template bodies).

## What it adds

**1. `WorkspaceIntegrationTemplate` validating webhook** (`workspaceintegrationtemplate_webhook.go`)
Runs on template **create and update** (`failurePolicy=fail`). Statically validates **every templated field the resolver later renders** — image, command, args, `workingDir`, env values, and volume-mount paths (`mountPath`/`subPath`/`subPathExpr`) on both additional containers and the primary-container modifications, plus each resourceRef's name/namespace — in one render:
- Go-template syntax is parseable.
- every `{{ resource "<id>" ... }}` references an `id` declared in `resourceRefs`, with a valid JSONPath (and `{{ resource }}` is rejected in a resourceRef's own metadata, where it's nonsensical).
- every `{{ .Parameters.X }}` names a **declared** parameter, and every `{{ .Workspace.X }}` a real field — so an **author typo** is rejected at the admin's write instead of resolving to empty later.

The validated set is kept in lockstep with the resolver's rendered fields, so a field the resolver renders can't slip past admission.

**2. Workspace integration validator** (`integration_template_ref_validator.go`, wired into the existing Workspace webhook)
Enforces, at the **user's** write:
- **Namespace scope** — an `integrationTemplateRefs[].namespace` may only target the workspace's own namespace or the configured shared namespace, checked **before** any template read so a cross-namespace reference fails closed without the operator (which holds cluster-wide read on `WorkspaceIntegrationTemplate`) fetching another team's template. Mirrors the existing `WorkspaceTemplate` (`validateTemplateNamespace`) and `AccessStrategy` guards exactly.
- **Parameter completeness** — the workspace supplies every parameter the referenced template **declares** (`spec.parameters`).
- A referenced template that **does not exist is REJECTED** at admission (matching the `WorkspaceTemplate` precedent, where a missing template fails the write). Supplied-but-**undeclared** parameters produce a non-blocking **warning** (they're ignored at resolve time but usually signal a typo). `Validate` returns `(admission.Warnings, error)` accordingly.

## Files

- Hand-authored (all in `internal/webhook/v1alpha1/`): `integration_template_validation.go` (+ test), `integration_template_ref_validator.go` (+ test), `workspaceintegrationtemplate_webhook.go`, and the validator wiring in `workspace_webhook.go`.
- Registration: `cmd/main.go` (template webhook gated by `ENABLE_WORKSPACE_INTEGRATION_TEMPLATE_WEBHOOK`, defaults on) + the generated `config/webhook/manifests.yaml`.

## Testing

`go build` / `go vet` / `gofmt` clean; the webhook envtest suite passes, including specs for template static-validation (syntax, undeclared resourceRef, bad JSONPath, `{{ resource }}`-in-metadata, undeclared parameter, `.Workspace` field typo, and workingDir + volume-mount path fields), parameter completeness against the declared list, missing-template **rejection**, unused-parameter **warning**, and the namespace-scope guard (reject cross-ns, allow own-ns, allow shared-ns, allow empty-ns default, reject cross-ns with no shared configured).


internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -0,0 +1,188 @@
...
+	client           client.Client
+	defaultNamespace string
+}
+
+// NewIntegrationTemplateRefValidator creates the validator. defaultNamespace is the shared namespace an
+// integrationTemplateRef may additionally target (beyond the workspace's own namespace); it is also used
+// to resolve a ref that omits its own namespace (the workspace's namespace is used per-workspace). This
+// is the SAME value the WorkspaceTemplate guard uses (--default-template-namespace), so integration
+// templates and workspace templates share one shared-namespace setting.
+func NewIntegrationTemplateRefValidator(c client.Client, defaultNamespace string) *IntegrationTemplateRefValidator {
```
JGuinegagne: sanity-check: `defaultNamespace` refers to the [shared namespace](https://jupyter-k8s.readthedocs.io/en/latest/concepts/templates/shared-namespace.html), correct?

If yes, please rename `sharedNamespace` so that it's clearer.

internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -0,0 +1,188 @@
...
+	if ns == "" {
+		ns = workspace.Namespace
+	}
+	tmpl := &workspacev1alpha1.WorkspaceIntegrationTemplate{}
+	err := v.client.Get(ctx, client.ObjectKey{Name: ref.Name, Namespace: ns}, tmpl)
+	// Fall back to the shared namespace, mirroring the controller's template resolution order
+	// (ref-ns -> workspace-ns -> shared-ns). Without this the webhook would reject a by-name reference to
+	// an admin-installed shared template that the controller would happily resolve -- so a chart-shipped
+	// template in the shared namespace could never be referenced by name.
+	if errors.IsNotFound(err) && v.defaultNamespace != "" && ns != v.defaultNamespace {
```
JGuinegagne: hum, I'm not sure we should infer the `integrationRef.namespace` in this way.

1. Are we following a precedent from template?
2. Wouldn't it be cleaner to require the user to pass the shared ns explicitly in this case?

internal/webhook/v1alpha1/workspaceintegrationtemplate_webhook.go
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
```
JGuinegagne, line 3: Add unit tests for this webhook please

internal/webhook/v1alpha1/workspace_webhook.go
```diff
@@ -216,13 +216,15 @@ func SetupWorkspaceWebhookWithManager(mgr ctrl.Manager, defaultTemplateNamespace
 	serviceAccountValidator := NewServiceAccountValidator(mgr.GetClient())
 	serviceAccountDefaulter := NewServiceAccountDefaulter(mgr.GetClient())
 	volumeValidator := NewVolumeValidator(mgr.GetClient())
+	integrationTemplateRefValidator := NewIntegrationTemplateRefValidator(mgr.GetClient(), defaultTemplateNamespace)
```
JGuinegagne, line 219: update the unit tests for `workspace_webhook` please

internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	sort.Strings(unused)
+	var warnings admission.Warnings
+	for _, name := range unused {
+		warnings = append(warnings, fmt.Sprintf(
+			"integrationTemplateRefs[%q]: parameter %q is not declared by the template and will be ignored (typo?)",
+			ref.Name, name))
+	}
+	return warnings
+}
```
JGuinegagne, line 202: non-blocking: could you either create a GH issue to ensure that `integrationRefs` do not attempt to modify the _same_ underlying resource. For example if ws references `[integTemplate-1, integTemplate-2]`:
- that try to add the same named sidecar container --> reject
- that try to modify the same env var --> reject
- that declare the same parameter name --> reject
- that attempt to set inconsistent values for shareProcessNamespace --> reject

JGuinegagne (changes requested):
Please:
- add / update unit tests to webhooks
- add e2e test coverage
- cut GH issue to ensure consistency of `integrationRefs` in ws validation webhook when (if) we allow several integration refs

internal/webhook/v1alpha1/integration_template_ref_validator.go
HariniNarayanan: yes, I can update it, but you do set it through --default-template-namespace flag. That's why I picked this - but shared is clearer

internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -0,0 +1,188 @@
...
+// workspace on the first invalid one. It also returns non-blocking warnings (supplied parameters the
+// template does not declare) aggregated across all refs -- surfaced to the user without failing admission.
+func (v *IntegrationTemplateRefValidator) Validate(ctx context.Context, workspace *workspacev1alpha1.Workspace) (admission.Warnings, error) {
+	log := logf.FromContext(ctx).WithName("integration-template-ref-validator")
+	var warnings admission.Warnings
+	for i := range workspace.Spec.IntegrationTemplateRefs {
+		ref := &workspace.Spec.IntegrationTemplateRefs[i]
+		refWarnings, err := v.validateRef(ctx, workspace, ref)
+		if err != nil {
+			log.Info("Rejected Workspace: invalid integrationTemplateRef",
```
psnilesh: couple of nits
1. Warn / Error log lvl might be more appropriate. 
2. log says `invalid integrationTemplateRef`, but it may not be true if the error thrown is `failed to load integration template`.

internal/webhook/v1alpha1/integration_template_ref_validator.go
HariniNarayanan: 1. Yes
2. The shared namespace is set by the admin at the controller level (--default-template-namespace flag on the manager) — so a data scientist may not know it. If we move this to something the data scientist creating the workspace has to provide - it might make this harder to use compared to workspace templates

internal/webhook/v1alpha1/integration_template_ref_validator.go
HariniNarayanan: Ah we do this at resolution, but you think it would be better at admission. Fair. I'll create an issue.

internal/webhook/v1alpha1/integration_template_ref_validator.go
psnilesh: +1 Could you help me understand the use-case a bit better for Harini?

psnilesh: Just thinking out loud -  do we need to save the resolved namespace somewhere so the user knows which namespace we got the IntegrationTemplate? I think we should log it at least.

HariniNarayanan: https://jupyter-k8s.readthedocs.io/en/latest/concepts/templates/shared-namespace.html

Set as (--default-template-namespace on the [manager](https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/cmd/main.go#L179)) - see [here](https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/config/manager/manager.yaml#L68)

How workspace template validator handles this fallback logic: https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/internal/webhook/v1alpha1/template_validator.go#L40

internal/webhook/v1alpha1/integration_ref_defaulter.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if !errors.IsNotFound(err) {
+		return "", fmt.Errorf("failed to resolve integration template %q in namespace %q: %w", name, workspaceNamespace, err)
+	}
+
+	// Not in the workspace namespace: try the shared namespace if one is configured and different.
+	if d.sharedNamespace == "" || d.sharedNamespace == workspaceNamespace {
+		return "", nil
+	}
+	err = d.client.Get(ctx, client.ObjectKey{Name: name, Namespace: d.sharedNamespace}, tmpl)
```
JGuinegagne: can we follow template ref [precedent](https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/webhook/v1alpha1/template_validator.go#L40) here?

As in, don't allow arbitrary ns at all in the permission model. As in, any ref _must_ live in the workspace own ns or in the `shared` ns. Anything else -> reject

test/e2e/workspace_integration_admission_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		It("rejects a workspace that omits a declared parameter", func() {
+			createIntegrationTemplateForTest("valid-integration")
+			applyExpectRejected("ws-integration-missing-param-rejected",
+				"workspace", "ws-integration-missing-param-rejected",
+				"a workspace omitting a declared parameter should be rejected")
+		})
+	})
+
+	Context("Mutating defaulter namespace stamping", func() {
```
JGuinegagne: not sure about this behavior honestly.

Template have a "default referencing" mechanism, as in they may get applied forcibly even if a workspace user doesn't reference a `templateRef`. In that case, it makes sense to lookup and add the `templateRef.{name, namespace}`.

Extending this mechanism to `integrationRefs` is quite frankly a stretch. I would prefer to require that the user pass explicitly the correct `{name, ns}` in the `workspace.spec.integrationRef` entry.

That would also simplify the webhook logic.

internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -0,0 +1,205 @@
...
+
+// validateNamespaceScope rejects an integrationTemplateRefs[].namespace that targets a namespace other
+// than the workspace's own or the configured shared (default) namespace. This mirrors the workspace
+// templateRef guard (TemplateValidator.validateTemplateNamespace) and the AccessStrategy guard: without
+// it, a user could point an integrationTemplateRef at ANY namespace and have the operator -- which holds
+// cluster-wide read on WorkspaceIntegrationTemplate -- resolve a template from another team's namespace
+// (a confused-deputy cross-namespace read). Enforcing it at admission fails the write closed BEFORE any
+// such read happens; the admission webhook is the single enforcement point (the controller resolver
+// reads from the already-validated stored spec).
+func (v *IntegrationTemplateRefValidator) validateNamespaceScope(ref *workspacev1alpha1.IntegrationTemplateRef, workspaceNamespace string) error {
```
JGuinegagne, line 132: `[Security]` **Forward-looking (not blocking this PR — the resolver/fetch path has no caller on this branch):** this scopes the *integrationTemplateRef* namespace to own-ns/shared-ns, but the **resourceRef target namespace** (`resourceRef.metadata.namespace`, rendered from a user-supplied `{{ .Parameters.X }}`) is not yet scoped. When the resolver lands, the controller — which holds cluster-wide read — will fetch that object from whatever the user's parameter renders to, reading another namespace's resource into the pod (confused deputy). Please gate the **rendered** `metadata.namespace` to the **workspace's own namespace** at resolve time (fail-closed). Since `metadata.namespace` already rejects `{{ resource }}` and only references `.Workspace`/`.Parameters`, the value is knowable at workspace admission and could optionally also be enforced here.

internal/webhook/v1alpha1/integration_template_validation.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// validateTemplateExpression parses and executes one templated field with missingkey=error, so a single
+// pass catches: bad Go-template syntax; an undeclared {{ .Parameters.X }} or bad {{ .Workspace.X }}
+// (missingkey=error, since allowedParamNames seeds only the declared names); and, via the resource func,
+// a {{ resource }} that is disallowed here (allowResourceRefs=false), names an undeclared resourceRef,
+// or carries an invalid JSONPath. The rendered output is discarded -- only errors matter. Prefer the
+// validateMetadataExpression / validatePodFieldExpression wrappers over calling this directly.
+//
+// KNOWN LIMITATION of static validation: missingkey=error only fires for the field-access form
```
JGuinegagne: `[Hardening]` This `KNOWN LIMITATION` note should also mention that expressions inside an untaken `{{ if }}`/`{{ range }}` branch aren't validated — every parameter renders empty (falsy) here, so a parameter-guarded branch is skipped and its errors surface only at resolve time.

JGuinegagne (commented):
Security-focused review of the admission validation. Overall this is well-scoped and the security-critical guard (integrationTemplateRef namespace scope) is correctly placed on the `failurePolicy=Fail` Workspace webhook and mirrors the existing `validateTemplateNamespace` precedent exactly.

**Validated as safe:**
- **Namespace-scope guard** runs before any template read, so a cross-namespace ref fails closed without the operator fetching another team's template. ✓
- **Parameter-value injection:** the resolver renders single-pass, so a user-supplied parameter *value* containing `{{ ... }}` is treated as literal data, not re-executed. ✓
- **WIT template webhook `failurePolicy=Ignore`:** correct — it is authoring-convenience only (no security invariant depends on it; the resolver is the fail-closed backstop), and follows the `WorkspaceTemplate` precedent. ✓
- **`integrationRefsChanged` update gate:** correct — admission scope-enforcement is best-effort-at-write and the authoritative cross-namespace-read guard is the resolver; re-validating unchanged refs would reintroduce the reconcile wedge. ✓

**Two non-blocking findings inline** (one forward-looking for the resolver-landing PR, one doc-only).

internal/webhook/v1alpha1/integration_ref_defaulter.go
```diff
@@ -0,0 +1,106 @@
...
+	if err == nil {
+		return workspaceNamespace, nil
+	}
+	if !errors.IsNotFound(err) {
+		return "", fmt.Errorf("failed to resolve integration template %q in namespace %q: %w", name, workspaceNamespace, err)
+	}
+
+	// Not in the workspace namespace: try the shared namespace if one is configured and different.
+	if d.sharedNamespace == "" || d.sharedNamespace == workspaceNamespace {
+		return "", nil
```
psnilesh: What does `return "", nil` mean?  Since we already checked `workspaceNamespace` and concluded that `sharedNamespace` is not configured, should we return an error instead?

internal/webhook/v1alpha1/integration_ref_defaulter.go
psnilesh: +100 Even for templates, allowing access from any namespace would almost certainly not be what the administrators intended.

internal/webhook/v1alpha1/integration_ref_defaulter.go
HariniNarayanan: Resolved by deleting the defaulter — that confusing "found nowhere, return empty, no error" branch is gone. The validator now does the own→shared read directly and rejects with a clear "not found in namespace X or shared namespace Y" message when the template exists in neither.

internal/webhook/v1alpha1/integration_ref_defaulter.go
HariniNarayanan: To clarify the two pieces: the arbitrary-namespace rejection was never in the mutating webhook — it's validateNamespaceScope in the validating webhook, which has always rejected any ns that isn't the workspace's own or the shared namespace. So the permission model was already correct and is unchanged.

The mutating webhook (IntegrationRefDefaulter) was doing something narrower: for a bare-name ref (no namespace), it looked the template up — workspace ns, then shared ns — purely to stamp the resolved namespace onto the stored spec. A ref with an explicit namespace was skipped entirely (no read). That lookup-to-stamp is the "stretch" you flagged, so I removed the defaulter.

Current behavior for integrationTemplateRefs[].namespace:

    empty → workspace namespace
    non-empty → must equal the workspace or shared namespace, else rejected

No mutation, no discovery read — the only template Get now is the validating webhook's existence check. Happy to add a shared-ns-only defaulter back later if a use case needs it.

test/e2e/workspace_integration_admission_test.go
HariniNarayanan: Fair — went with explicit {name, ns}. The auto-lookup/stamp mechanism is removed

internal/webhook/v1alpha1/integration_template_validation.go
HariniNarayanan: Added. The note now covers both gaps: (1) the dynamic-key {{ index .Parameters "X" }} form bypasses missingkey=error; and (2) validation seeds params to empty, so a {{ if .Parameters.X }} / {{ range }} guard is falsy and its body isn't executed — expressions inside an untaken branch aren't validated here and only surface at resolve time.

internal/webhook/v1alpha1/integration_template_ref_validator.go
HariniNarayanan: For resourceRef, our PM was explicit to allow reads across namespaces. @psnilesh called out in the first PR whether this would lead to escalation of privilege - the answer to that is no and yes.

No escalation of privilege - in connecting to the Ray Cluster, this is set up at the Ray cluster creation
Escalation of privilege - in getting the ray cluster.

Solving the second has two options:
1. The solution recommended above where we restrict the namespace
2. SubjectAccessReview on the resource provided - PM was specific that connecting to ray clusters across namespaces is a requirement.

I'll take this discussion out of this PR.

JGuinegagne: It's a read escalation of privilege: a user can discover resources in a ns they don't have access to by probing via "create ws w/ integrationRef".

Overall, restricting integrations to resources in the same ns would be cleaner, but not a hard blocker.

internal/webhook/v1alpha1/integration_template_ref_validator.go
JGuinegagne: https://github.com/jupyter-infra/jupyter-k8s/issues/444

JGuinegagne (approved):

PR #422, opened 2026-06-25
feat: create and GC WorkspaceIntegration child shells from the workspace controller

HariniNarayanan:
## Summary

Wires the workspace controller to manage the `WorkspaceIntegration` child objects for a workspace. On each reconcile, the controller ensures one `WorkspaceIntegration` child exists per `workspace.spec.integrationRefs` entry, updates a child whose `templateRef`/`workspaceRef` has drifted, and deletes any child no longer backed by a ref. Children are owned by the workspace (controller `ownerReference`) so Kubernetes garbage-collects them when the workspace is deleted.

This is the operator half of the design: the controller owns each child's **identity and references**; it performs **no** resolution itself — it never reads the `WorkspaceIntegrationTemplate` or the referenced resources. The resolved output fields are produced separately at the child's own admission, which keeps the workspace reconcile loop free of reconcile-time resource reads.

## What this commit does

`WorkspaceIntegrationManager` (new) provides:

- **`EnsureForWorkspace`** — the reconcile entry point. For each `integrationRefs` entry it creates a child "shell" (`templateRef` + `workspaceRef` + identification label + ownerRef) if missing, updates it if the intent drifted, and deletes children whose ref was removed. It returns the reconciled set, assembled locally as it goes.
- **`ListForWorkspace`** — lists a workspace's children by the `workspace.jupyter.org/workspace-name` label (the canonical identification path), namespace-scoped.

The state machine calls `EnsureForWorkspace` on the running path before building the deployment; failing to ensure a child fails the reconcile so a workspace is never built missing an integration.

### Key properties

- **GC + re-enqueue:** `controllerutil.SetControllerReference(workspace, child, ...)` gives both garbage collection on workspace delete and re-enqueue of the workspace on child changes (via the `Owns(...)` watch).
- **No output clobbering:** create/update sets only the shell input fields (`templateRef`/`workspaceRef`); the resolved output fields are left untouched, so the controller never fights the component that fills them. The idempotency test verifies a re-reconcile preserves externally-resolved output.
- **Race-safe create:** `AlreadyExists` on create is handled by fetching the existing child, so a concurrent reconcile doesn't error the loop.

## Dependencies

> ⚠️ **This branch does not compile on its own** and is expected to show a red build until its dependency merges. It is published now so the child-lifecycle management can be reviewed in isolation.

This commit builds on the **API types commit**, which introduces the `WorkspaceIntegration` / `WorkspaceIntegrationList` types and the `workspace.spec.integrationRefs` field this manager reconciles against.

It does **not** depend on the resolver/builder/webhook commits — it only manages child identity and references and deliberately never touches resolved output. Once the API types land on `main`, this branch rebases onto `main` cleanly and compiles. The PR can be rebased in place at that point — same branch, same PR.

## Why this is a separate commit

The earlier commits in the series are the **resolution path** (turning a template into frozen output at the child's admission). This commit is the **lifecycle path** (creating, updating, GC-ing the right set of children from `integrationRefs`). They are orthogonal concerns with a clean boundary:

- This manager owns *which children exist and what they reference*.
- The admission path owns *what each child resolves to*.

Keeping them in separate commits means the controller's reconcile logic is reviewed and tested purely as set-reconciliation (create/update/delete/GC), with no resolution coupling — the manager's tests assert it never writes the output fields, proving the boundary holds.

## What's in this PR

| File | What |
|---|---|
| `internal/controller/workspaceintegration_manager.go` | `WorkspaceIntegrationManager` — `EnsureForWorkspace`, `ListForWorkspace`, shell builder. |
| `internal/controller/workspaceintegration_manager_test.go` | Unit tests against a fake client. |
| `internal/controller/state_machine.go` | Calls `EnsureForWorkspace` on the running path; wires the manager in. |
| `internal/controller/workspace_controller.go` | Constructs the manager in `SetupWorkspaceController`. |
| `internal/controller/constants.go` | `GenerateWorkspaceIntegrationName` (deterministic child name). |

## Testing

Unit tests against a fake client cover:

- Create a shell with the correct ownerRef, identification label, and references — and that the resolved output fields are left unset (owned elsewhere).
- Idempotency: a second reconcile creates nothing new and preserves externally-resolved output.
- Update on `templateRef` drift (e.g. switching the referenced cluster).
- Delete of a stale child when its `integrationRefs` entry is removed.
- Label-filtered listing (a workspace sees only its own children).

```
go test ./internal/controller/ -run 'TestEnsureForWorkspace|TestListForWorkspace' -count=1
```

## References

- controller-runtime owner references & GC: https://book.kubebuilder.io/reference/using-finalizers (and `controllerutil.SetControllerReference`)
- Kubernetes garbage collection (owner refs): https://kubernetes.io/docs/concepts/architecture/garbage-collection/

### DIFF
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -169,6 +169,14 @@ func GeneratePVCName(workspaceName string) string {
 	return fmt.Sprintf("%s-%s-pvc", ResourcePrefix, workspaceName)
 }
 
+// GenerateWorkspaceIntegrationName creates a deterministic WorkspaceIntegration child name for a
+// given workspace + integrationRef. The 1:1 <workspace>-<ref> mapping makes the child name
+// predictable; identification at deployment-build time uses the LabelWorkspaceName label selector
+// (option B), with this name only ensuring a stable, collision-free child per ref.
+func GenerateWorkspaceIntegrationName(workspaceName, integrationRefName string) string {
+	return fmt.Sprintf("%s-%s-%s", ResourcePrefix, workspaceName, integrationRefName)
+}
+
 // GenerateLabels creates consistent labels for resources
 func GenerateLabels(workspaceName string) map[string]string {
 	return map[string]string{
--- a/internal/controller/state_machine.go
+++ b/internal/controller/state_machine.go
@@ -33,6 +33,10 @@ type StateMachine struct {
 	recorder            record.EventRecorder
 	idleChecker         *WorkspaceIdleChecker
 	accessStartupProber AccessStartupProberInterface
+	// integrationManager is always wired by SetupWorkspaceController. It is a pointer (and the
+	// reconcile path nil-guards it) only so unit tests can construct a StateMachine without the
+	// integration machinery; production code must never leave it nil.
+	integrationManager *WorkspaceIntegrationManager
 }
 
 // NewStateMachine creates a new StateMachine
@@ -42,6 +46,7 @@ func NewStateMachine(
 	recorder record.EventRecorder,
 	idleChecker *WorkspaceIdleChecker,
 	accessStartupProber AccessStartupProberInterface,
+	integrationManager *WorkspaceIntegrationManager,
 ) *StateMachine {
 	return &StateMachine{
 		resourceManager:     resourceManager,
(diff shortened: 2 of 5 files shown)


HariniNarayanan:
Superseded by the 2b values-in-status freeze split (#418/#419/#421/#423/#424/#425).

PR #423, opened 2026-06-25
feat: build workspace Deployment from frozen integration values

HariniNarayanan:
## Summary

Implements the **values-in-status freeze** that shapes the workspace Deployment. Each integration is resolved against its referenced resource **only when its input changes** (a `parametersVersion` or `templateVersion` drift); the resolved substitution values are frozen into `status.resolvedIntegrations[]` and re-rendered from that frozen set on every other reconcile. This is the only PR that reads the referenced resource.

> **Stacked PR.** Built on the integration API-types (#418) and resolver (#419) PRs — review/merge those first.
>
> Note: there is no `WorkspaceIntegration` child object. The frozen values live on the workspace's own `status`, not a separate CRD.

## What it does

Two sides, in one file each:

- **Capture side — `resource_manager_integration.go`.** `reconcileIntegrations` runs before the build: for each `integrationTemplateRefs[]` entry it computes the re-resolve gate from two versions —
  - `parametersVersion` — hash of the templateRef identity (namespace+name) + sorted parameters (changes when the user switches clusters or edits a parameter);
  - `observedIntegrationTemplateVersion` — the template's `UID.Generation` (changes when the admin edits/replaces the template).

  On a drift (`hasIntegrationChanged`), `captureIntegrationValues` reads the referenced resource (`fetchReferencedResources`) and records the `{{ resource }}` → literal value map — rendering **both** the pod modifications **and the `statusProbe` command** so every `{{ resource }}` either side references is frozen — which `persistResolvedIntegrationsToStatus` writes in one status patch. On no drift, the frozen record is carried forward — no resource read.
- **Build side — `deployment_builder_integration.go`.** `applyIntegrationsToDeployment` (called from `BuildWorkspaceDeployment`, the single desired-state build path used by create/update/`NeedsUpdate`) renders each integration's pod modifications via `buildIntegrationPodModifications` from the **frozen** values only — never reading the referenced resource — and OR-reduces the template's `shareProcessNamespace` onto the pod. Same frozen values → byte-identical pod template → the pod never rolls on external drift.
- **Status-probe resolution.** `getIntegrationStatusProbeSpec` returns the template's `statusProbe` with its exec command **resolved from the frozen values** (via the shared `resolveStatusProbeCommand` helper). The probe command carries the same `{{ resource }}` / `{{ .Parameters }}` expressions as the sidecar, so it must be rendered from the frozen record before the prober (#424) execs it — otherwise the probe runs against a literal `{{ resource … }}` string. Capture (live provider, harvests the keys) and this path (frozen provider, replays them) call the same helper, so the identical expression set is frozen and replayed.
- **Shared helpers — `integration_helpers.go`.** `findResolvedIntegration`, `buildIntegrationTemplateData`, `getIntegrationTemplate`, `resolveStatusProbeCommand` (used by both sides).

**Fail-closed + non-fatal** (contract documented in `integration_helpers.go`): a capture error preserves the existing frozen record; a render error aborts the build so the running Deployment (and its sidecar) is left untouched rather than stripped; a desired-build error in `NeedsUpdate` is non-fatal for integration workspaces.

## RBAC

Grants the operator **get-only** access to `ray.io/rayclusters` (the referenced resource kind), synced to `config/rbac/role.yaml` and the chart `manager-role.yaml`. Narrow by design — a new referenced kind needs its own marker.

## What's in this PR

| File | What |
|---|---|
| `internal/controller/resource_manager_integration.go` | Capture: `reconcileIntegrations`, `captureIntegrationValues` (renders pod-mods + probe), `getIntegrationStatusProbeSpec` (resolves probe from frozen values), `fetchReferencedResources`, `persistResolvedIntegrationsToStatus`, `hasIntegrationChanged`, version helpers. |
| `internal/controller/deployment_builder_integration.go` | Build: `applyIntegrationsToDeployment`, `buildIntegrationPodModifications`, pod-mod merge. |
| `internal/controller/integration_helpers.go` | Shared helpers used by both sides, incl. `resolveStatusProbeCommand`. |
| `internal/controller/deployment_builder.go` | `BuildWorkspaceDeployment` (base + access + integration), `NeedsUpdate` includes the overlay. |
| `internal/controller/resource_manager.go` | Wires the freeze into `EnsureDeploymentExists`. |
| `config/rbac/role.yaml`, `dist/chart/.../manager-role.yaml`, `dist/install.yaml` | `ray.io/rayclusters` get RBAC (config + chart). |

## Testing

Unit + envtest cover: first resolve (freeze), external-drift-no-roll (unchanged input → byte-identical pod), parameter-switch re-resolve, template-edit re-resolve, fail-closed on a missing referenced resource, and fail-closed on a deleted template (running sidecar preserved). `resolveStatusProbeCommand` has its own unit test (`TestResolveStatusProbeCommand_CaptureThenReplay`) proving the capture→replay round-trip renders the identical command from the live and frozen providers.

```
go test ./internal/controller/ -run 'Integration|Freeze|ResolveStatusProbeCommand' -count=1
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -79,6 +79,12 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ray.io
+  resources:
+  - rayclusters
+  verbs:
+  - get
+- apiGroups:
   - traefik.io
   resources:
   - ingressroutes
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -79,12 +79,6 @@ rules:
   - update
   - watch
 - apiGroups:
-  - ray.io
-  resources:
-  - rayclusters
-  verbs:
-  - get
-- apiGroups:
   - traefik.io
   resources:
   - ingressroutes
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
@@ -26,19 +25,14 @@ type DeploymentBuilder struct {
 	scheme        *runtime.Scheme
 	options       WorkspaceControllerOptions
 	imageResolver *ImageResolver
-	// client is used to load a WorkspaceIntegrationTemplate when replaying an integration's frozen
-	// values into the pod template (see deployment_builder_integration.go). It never reads a referenced
-	// resource -- that read happens during capture, before the build.
-	client client.Client
 }
 
 // NewDeploymentBuilder creates a new DeploymentBuilder
-func NewDeploymentBuilder(scheme *runtime.Scheme, options WorkspaceControllerOptions, k8sClient client.Client) *DeploymentBuilder {
+func NewDeploymentBuilder(scheme *runtime.Scheme, options WorkspaceControllerOptions) *DeploymentBuilder {
 	return &DeploymentBuilder{
 		scheme:        scheme,
 		options:       options,
 		imageResolver: NewImageResolver(options.ApplicationImagesRegistry),
-		client:        k8sClient,
 	}
 }
 
@@ -58,15 +52,18 @@ func (db *DeploymentBuilder) BuildDeployment(ctx context.Context, workspace *wor
 	return deployment, nil
 }
 
-// BuildWorkspaceDeployment builds the full desired Deployment for a workspace: the base pod template,
-// then the access-strategy overlay, then the integration overlay (frozen values replayed from
-// status.resolvedIntegrations). It is the single desired-state entry point -- create, update, and
-// NeedsUpdate all call it, so the compared desired state is identical in every path and an unchanged
-// input produces no diff and no pod roll.
+// BuildWorkspaceDeployment builds the full desired Deployment for a workspace by layering, in order:
+// the base pod template, then the integration overlays (frozen values replayed from
+// status.resolvedIntegrations), then the access-strategy overlay. Access strategy is applied LAST so it
+// wins every conflict (a primary-container env var it sets overwrites the same key from an integration):
+// access to the workspace is fundamental to its usability, so an integration must never be able to break
+// it. It is the single desired-state entry point -- create, update, and NeedsUpdate all call it, so the
+// compared desired state is identical in every path and an unchanged input produces no diff and no pod roll.
 func (db *DeploymentBuilder) BuildWorkspaceDeployment(
 	ctx context.Context,
 	workspace *workspacev1alpha1.Workspace,
 	accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy,
+	integrationTemplates map[string]*workspacev1alpha1.WorkspaceIntegrationTemplate,
 ) (*appsv1.Deployment, error) {
 	// Build the base deployment
 	deployment, err := db.BuildDeployment(ctx, workspace)
@@ -74,15 +71,17 @@ func (db *DeploymentBuilder) BuildWorkspaceDeployment(
 		return nil, err
 	}
 
+	// Apply the workspace's integrations by replaying their frozen values (status.resolvedIntegrations)
+	// onto the pre-fetched templates the caller supplies.
+	if err := db.applyIntegrationsToDeployment(ctx, deployment, workspace, integrationTemplates); err != nil {
+		return nil, err
+	}
+
+	// Access strategy is applied after integrations so its settings take precedence.
 	if accessStrategy != nil {
 		if err := db.ApplyAccessStrategyToDeployment(deployment, workspace, accessStrategy); err != nil {
 			return nil, fmt.Errorf("failed to apply access strategy to deployment: %w", err)
 		}
-	}
-
-	// Apply the workspace's integrations by replaying their frozen values (status.resolvedIntegrations).
-	if err := db.applyIntegrationsToDeployment(ctx, deployment, workspace); err != nil {
-		return nil, err
 	}
 
 	return deployment, nil
@@ -332,9 +331,10 @@ func (db *DeploymentBuilder) NeedsUpdate(
 	existingDeployment *appsv1.Deployment,
 	workspace *workspacev1alpha1.Workspace,
 	accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy,
+	integrationTemplates map[string]*workspacev1alpha1.WorkspaceIntegrationTemplate,
 ) (bool, error) {
 	// Build the desired deployment spec with access strategy applied
-	desiredDeployment, err := db.BuildWorkspaceDeployment(ctx, workspace, accessStrategy)
+	desiredDeployment, err := db.BuildWorkspaceDeployment(ctx, workspace, accessStrategy, integrationTemplates)
 	if err != nil {
 		return false, fmt.Errorf("failed to build desired deployment: %w", err)
 	}
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -26,6 +26,10 @@ type DeploymentBuilder struct {
 	scheme        *runtime.Scheme
 	options       WorkspaceControllerOptions
 	imageResolver *ImageResolver
+	// client is used to load a WorkspaceIntegrationTemplate when replaying an integration's frozen
+	// values into the pod template (see deployment_builder_integration.go). It never reads a referenced
+	// resource -- that read happens during capture, before the build.
+	client client.Client
 }
 
 // NewDeploymentBuilder creates a new DeploymentBuilder
@@ -34,6 +38,7 @@ func NewDeploymentBuilder(scheme *runtime.Scheme, options WorkspaceControllerOpt
 		scheme:        scheme,
 		options:       options,
 		imageResolver: NewImageResolver(options.ApplicationImagesRegistry),
+		client:        k8sClient,
 	}
 }
 
@@ -53,9 +58,12 @@ func (db *DeploymentBuilder) BuildDeployment(ctx context.Context, workspace *wor
 	return deployment, nil
 }
 
-// BuildDeploymentWithAccessStrategy creates a Deployment resource with access strategy applied
-// This is a helper function that should be called from ResourceManager
-func (db *DeploymentBuilder) BuildDeploymentWithAccessStrategy(
+// BuildWorkspaceDeployment builds the full desired Deployment for a workspace: the base pod template,
+// then the access-strategy overlay, then the integration overlay (frozen values replayed from
+// status.resolvedIntegrations). It is the single desired-state entry point -- create, update, and
+// NeedsUpdate all call it, so the compared desired state is identical in every path and an unchanged
+// input produces no diff and no pod roll.
+func (db *DeploymentBuilder) BuildWorkspaceDeployment(
 	ctx context.Context,
 	workspace *workspacev1alpha1.Workspace,
 	accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy,
@@ -70,6 +78,11 @@ func (db *DeploymentBuilder) BuildDeploymentWithAccessStrategy(
 		if err := db.ApplyAccessStrategyToDeployment(deployment, workspace, accessStrategy); err != nil {
 			return nil, fmt.Errorf("failed to apply access strategy to deployment: %w", err)
 		}
+	}
+
+	// Apply the workspace's integrations by replaying their frozen values (status.resolvedIntegrations).
+	if err := db.applyIntegrationsToDeployment(ctx, deployment, workspace); err != nil {
+		return nil, err
 	}
 
 	return deployment, nil
@@ -236,7 +249,7 @@ func (db *DeploymentBuilder) buildPrimaryContainer(workspace *workspacev1alpha1.
 	}
 
 	container := corev1.Container{
-		Name:            ResourcePrefix,
+		Name:            PrimaryContainerName,
 		Image:           image,
 		ImagePullPolicy: db.options.ApplicationImagesPullPolicy,
 		SecurityContext: workspace.Spec.ContainerSecurityContext,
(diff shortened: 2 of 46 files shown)


config/rbac/role.yaml
```diff
@@ -78,6 +78,12 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ray.io
+  resources:
+  - rayclusters
+  verbs:
+  - get
```
JGuinegagne: Hum, that shouldn't be on by default.
If you want a specific operator flag, use [this pattern](https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/cmd/main.go#L171) please. But let's not turn this flag on in the operator chart by default. For the generic case, we have [this pattern](https://github.com/jupyter-infra/jupyter-k8s/blob/b52e003a9e535c00a2a18729b4818992f79f710e/cmd/main.go#L175).

BTW, i don't think the `traefik` permission should be there either, would you be able to cut an issue to look into this?

internal/controller/deployment_builder.go
```diff
@@ -26,6 +26,10 @@ type DeploymentBuilder struct {
 	scheme        *runtime.Scheme
 	options       WorkspaceControllerOptions
 	imageResolver *ImageResolver
+	// client is used to load a WorkspaceIntegrationTemplate when replaying an integration's frozen
+	// values into the pod template (see deployment_builder_integration.go). It never reads a referenced
+	// resource -- that read happens during capture, before the build.
+	client client.Client
```
JGuinegagne: anti-pattern: why does `DeploymentBuilder` need a client?

context: `_builder` should not make k8s calls. Please fetch the integration refs in `resource_manager` (once) and pass to the builder(s) all the references it needs.

internal/controller/deployment_builder_integration.go
```diff
@@ -0,0 +1,188 @@
...
+			return &ps.Containers[i]
+		}
+	}
+	return nil
+}
+
+// setEnv sets (overwriting an existing same-named entry) a literal-valued env var on a container.
+// It assigns a FRESH EnvVar so that if the prior entry carried a ValueFrom, that source is cleared --
+// leaving both Value and ValueFrom set is rejected by the API server.
+func setEnv(c *corev1.Container, name, value string) {
```
JGuinegagne, line 180: here we need to be careful, so what's the priority order?

My mental model:
Access-strategy.MergeEnv > IntegrationRef.MergeEnv > workspace.Env

Is that what you have in mind?

internal/controller/integration_helpers.go
```diff
@@ -0,0 +1,98 @@
...
+		resolved.Exec.Command[i] = rendered
+	}
+	return resolved, nil
+}
+
+// getIntegrationTemplate loads the referenced template from the ref's namespace, defaulting to the
+// workspace's own namespace when the ref omits one. The mutating webhook (IntegrationRefDefaulter)
+// stamps the resolved namespace onto the ref at admission, so a stored workspace's ref is already
+// fully-qualified. Missing is a fail-closed error.
+func getIntegrationTemplate(
```
JGuinegagne, line 79: `resource_manager` needs to call this, not within the `deployment_builder` please.

internal/controller/resource_manager.go
```diff
@@ -189,6 +190,17 @@ func (rm *ResourceManager) EnsureDeploymentExists(
...
+	// Integration workspaces: refresh the frozen resolution (status.resolvedIntegrations) BEFORE
+	// building the Deployment, so create/update below replay the current frozen values. Re-resolution
+	// happens only on an input-token change; a capture failure is non-fatal here (fail-closed --
+	// preserve the running pod) and is surfaced via the workspace's integration status elsewhere.
+	if rm.hasIntegrationTemplateRefs(workspace) {
+		if err := rm.reconcileIntegrations(ctx, workspace); err != nil {
+			logf.FromContext(ctx).Error(err, "integration freeze reconcile reported an error; proceeding with preserved frozen values",
+				"workspace", workspace.Name)
+		}
+	}
```
JGuinegagne: It's a bit odd that this is in `EnsureDeploymentExists`. I would say look at where `AccessStrategy` get fetches, and fetch these resources next to there when necessary.

Careful w/ `workspace.status` as well; if fetching an `integrationRef` fails, that ought to be reported in `workspace.status`.

However, we _must_ call update status _at most_ once in a single reconciliation pass. So we copy the `status`, and update it throughout the reconciliation loop, then update it once. Follow this pattern please.

internal/controller/resource_manager_integration.go
```diff
@@ -0,0 +1,260 @@
...
+// concurrent condition/probe writes) and updates the in-memory workspace so the same reconcile's build
+// renders from the just-frozen values.
+func (rm *ResourceManager) persistResolvedIntegrationsToStatus(
+	ctx context.Context,
+	workspace *workspacev1alpha1.Workspace,
+	resolved []workspacev1alpha1.ResolvedIntegration,
+) error {
+	base := workspace.DeepCopy()
+	workspace.Status.ResolvedIntegrations = resolved
+	if err := rm.client.Status().Patch(ctx, workspace, client.MergeFrom(base)); err != nil {
```
JGuinegagne: nope; `workspace.status` updates / patch may only happen once per reconciliation. It _must_ be handled by `status_manager`

internal/controller/resource_manager_integration.go
```diff
@@ -0,0 +1,260 @@
...
+// fetchReferencedResources reads each resourceRef's live object (unstructured). Fail-closed: a
+// missing/forbidden ref errors so no partial overlay is applied.
+func (rm *ResourceManager) fetchReferencedResources(
+	ctx context.Context,
+	template *workspacev1alpha1.WorkspaceIntegrationTemplate,
+	data IntegrationTemplateData,
+	refResolver *IntegrationTemplateResolver,
+) (map[string]*unstructured.Unstructured, error) {
+	resources := make(map[string]*unstructured.Unstructured, len(template.Spec.ResourceRefs))
+	for i := range template.Spec.ResourceRefs {
```
JGuinegagne, line 185: non-blocking: we should probably fetch these in parallel (w/ channels). Okay for now since we are only limiting the size to 1.

internal/controller/deployment_builder_integration.go
```diff
@@ -0,0 +1,188 @@
...
+
+// buildIntegrationPodModifications loads the template for pod SHAPE but resolves {{ resource }} values
+// from the frozen set (a missing frozen value is a hard error). shareProcessNamespace is carried through.
+func (db *DeploymentBuilder) buildIntegrationPodModifications(
+	ctx context.Context,
+	workspace *workspacev1alpha1.Workspace,
+	ref *workspacev1alpha1.IntegrationTemplateRef,
+	frozen *workspacev1alpha1.ResolvedIntegration,
+) (*workspacev1alpha1.PodModifications, *bool, error) {
+	template, err := getIntegrationTemplate(ctx, db.client, workspace, ref)
```
JGuinegagne: Anti-pattern: External refs should not be fetched in a builder object.

internal/controller/deployment_builder.go
```diff
@@ -72,6 +80,11 @@ func (db *DeploymentBuilder) BuildDeploymentWithAccessStrategy(
 		}
 	}
 
+	// Apply the workspace's integrations by replaying their frozen values (status.resolvedIntegrations).
+	if err := db.applyIntegrationsToDeployment(ctx, deployment, workspace); err != nil {
```
JGuinegagne: applying integration modifications _after_ access strategy means that `IntegrationRefs` induced modifications take precedence over access-strategy induced modifications, doesn't it?

If yes, that's wrong, `accessStrategy` settings should always win: it's fundamental to the workspace usability, we can't allow it to be broken by an integration. Said another way, we would rather break an integration than the access to the workspace.

JGuinegagne: This is subtle, so please add unit-tests for this behavior to ensure we don't regress later.

JGuinegagne (changes requested):
Watch out for:
- order of precedence between access-strategy, references, and `workspace.spec`
- status update
- where resources are fetched and how error are reported

internal/controller/deployment_builder.go
HariniNarayanan: Yes, moved.

internal/controller/integration_helpers.go
HariniNarayanan: Fixed

internal/controller/deployment_builder_integration.go
HariniNarayanan: Fixed

internal/controller/deployment_builder.go
HariniNarayanan: Done, added UTs to verify

internal/controller/resource_manager_integration.go
HariniNarayanan: Fixed

internal/controller/resource_manager.go
HariniNarayanan: Fixed

internal/controller/deployment_builder_integration.go
HariniNarayanan: Yes, added a test to verify this

HariniNarayanan:
> ## roborev: Combined Review (`083dc01`)
> 
> **Verdict:** One Medium security finding worth addressing (potential cross-namespace confused-deputy read); no High or Critical issues.
> ## Medium
> ### Confused-deputy cross-namespace read of referenced resources
> 
> **File:** `internal/controller/resource_manager_integration.go:172-192` (`fetchReferencedResources`), reachable from the capture path.
> 
> `fetchReferencedResources` resolves each `resourceRef`'s namespace from the template's `Metadata.Namespace`, which the shipped template pattern templatizes from a workspace-author-supplied parameter (e.g. `Namespace: "{{ .Parameters.rayClusterNamespace }}"`; see `ResolveResourceRef` at `integration_template_resolver.go:191`). The controller then reads that object with its own (potentially cluster-wide) privileges:
> 
> ```go
> obj.SetGroupVersionKind(gvk)
> if err := rm.client.Get(ctx, client.ObjectKey{Name: resolvedName, Namespace: resolvedNamespace}, obj); err != nil {
> ```
> 
> A workspace author who can create a `Workspace` in namespace A and reference a namespace-parameterized template can point the ref at namespace B and have the declared JSONPath field(s) of that object resolved and injected into their own pod (sidecar args / primary-container env via `setEnv`), where they can read it. Admission (`template_validator`) confines _which template_ a workspace may reference, but does not constrain the namespace a `resourceRef` inside the template resolves to. The `== ""` default only applies when the template omits the namespace.
> 
> The exposure is bounded by what the template author declares (Kind + JSONPaths), so the shipped Ray case (`{.status.head.serviceName}`) leaks innocuous data — but a template referencing a sensitive kind plus a namespace parameter silently becomes a cross-tenant read primitive.
> 
> **Exploit path:** author creates `Workspace` in ns A referencing a shared-namespace integration template whose `resourceRef.metadata.namespace` is `{{ .Parameters.someNs }}`; sets `someNs: victim-ns`; the operator reads the object in `victim-ns` and freezes its declared field into the author's pod env.
> 
> **Remediation:** confine the resolved `resourceRef` namespace to the workspace's namespace (reject/ignore a resolved namespace `!= workspace.Namespace`), or gate cross-namespace refs behind an explicit, admin-set allow-list on the template that is _not_ parameter-derived. If cross-namespace reads are intentional and admin-trusted, document that a namespace-parameterized `resourceRef` grants workspace authors the operator's read scope for that kind.
> 
> _Note (not a finding):_ resolved values are injected as discrete `[]string` exec args and env `Value` fields, not shell strings, so there is no command-injection path; and `setEnv` correctly clears a prior `ValueFrom` when overwriting, avoiding an invalid dual-source EnvVar.
> 
> _Synthesized from 2 reviews (agents: claude-code | types: default, security)_

Thanks — this is a real concern and a correct read of the current code. It's the same cross-namespace resourceRef scoping issue JGuinegagne flagged on #421 (integration_template_ref_validator.go:135), and it's inherent to the capture path rather than introduced by this PR.

Deferring to a follow-up, by design. #423 is scoped to the build/freeze path — it replays already-captured values into the pod and does not itself gate where a resourceRef resolves. The fix lands in the follow-on resource-access PR, which is where the controller actually reads the referenced object and is the right place to enforce it. That PR adds two guards on the resolved resourceRef:

    Namespace confinement — the resolved resourceRef namespace is constrained (fail-closed to the workspace's own namespace / the allowed shared scope), so a workspace author can't point it at an arbitrary victim namespace.
    SubjectAccessReview — the referenced read is authorized against the requesting user's identity, not the operator's cluster-wide privileges, which closes the confused-deputy path directly.

Tracking it so it isn't lost, and will link the follow-up PR here once it's up. Keeping #423 focused on the build/freeze path.

internal/controller/deployment_builder_integration.go
```diff
@@ -65,23 +75,14 @@ func (db *DeploymentBuilder) applyIntegrationsToDeployment(
...
-// buildIntegrationPodModifications loads the template for pod SHAPE but resolves {{ resource }} values
-// from the frozen set (a missing frozen value is a hard error). shareProcessNamespace is carried through.
+// buildIntegrationPodModifications resolves {{ resource }} values from the frozen set (a missing frozen
+// value is a hard error) against the pre-fetched template's pod shape. shareProcessNamespace is carried through.
 func (db *DeploymentBuilder) buildIntegrationPodModifications(
-	ctx context.Context,
 	workspace *workspacev1alpha1.Workspace,
 	ref *workspacev1alpha1.IntegrationTemplateRef,
 	frozen *workspacev1alpha1.ResolvedIntegration,
+	template *workspacev1alpha1.WorkspaceIntegrationTemplate,
```
JGuinegagne, line 84: non-blocking: `template` in this repo naturally maps to `WorkspaceTemplate`; could we use `integrationTemplate` ?

internal/controller/resource_manager_integration.go
```diff
@@ -31,36 +31,47 @@ import (
...
+// status_manager) persists it -- so status is written at most once per reconcile. Fail-closed: a template
+// that can't be loaded, or a capture that fails, preserves the existing frozen record and is reported as an
+// error; a template absent from the returned map makes the build preserve the running pod for that ref.
+func (rm *ResourceManager) reconcileIntegrations(
+	ctx context.Context,
+	workspace *workspacev1alpha1.Workspace,
+) (map[string]*workspacev1alpha1.WorkspaceIntegrationTemplate, error) {
 	logger := logf.FromContext(ctx)
 
+	templates := make(map[string]*workspacev1alpha1.WorkspaceIntegrationTemplate, len(workspace.Spec.IntegrationTemplateRefs))
```
JGuinegagne, line 64: same non-blocking nit here

JGuinegagne (approved):

PR #424, opened 2026-06-25
feat: probe integration status from the workspace pod

HariniNarayanan:
## Summary

Adds a non-gating, **report-only status probe** for integrations. On each reconcile of a Running workspace, the operator runs each frozen integration's `statusProbe` (an exec in the workspace pod) and writes one verdict per integration to `workspace.status.integrationStatuses[]`. The probe never blocks the workspace from becoming Available and never restarts the pod — it only reports health, mirroring the existing idle-checker pattern.

> **Stacked PR.** Built on the API-types (#418), resolver (#419), and freeze (#423) PRs — review/merge those first.
>
> Note: the `statusProbe` is read from the **template** (a static, admin-authored field), not from the referenced resource — so reading it does not re-read the RayCluster and the freeze contract from #423 is preserved. There is no `WorkspaceIntegration` child object.

## What it does

- **`IntegrationProber`** — two operations kept separate so the driver controls cost:
  - `FindRunningPod` locates the workspace pod by label, **skipping a terminating pod** (a `Running` pod with a `DeletionTimestamp` during a Recreate rollout would give a stale verdict for the incoming pod).
  - `Probe` execs the integration's `statusProbe.Exec` command in the workspace container and returns one `IntegrationStatus` (Ready + machine-readable Reason + optional Message carrying stderr). Only the Exec handler is implemented today. The command it receives is already **resolved from the frozen values** by `getIntegrationStatusProbeSpec` (#423) — the prober itself does no template resolution, so it never execs a literal `{{ resource … }}` string.
- **`probeIntegrationStatus` (state machine)** — the multi-integration driver. It resolves the pod **once** for the whole pass (N integrations cost one pod lookup, not N) and lands each ref in exactly one branch: a probe verdict for a resolved integration that declares a `statusProbe`; `NotResolved` (Degraded) for one that failed to freeze; `ProbeError` (Degraded) when its template has gone missing after freeze; `PodNotFound` (Degraded) when no running pod exists yet (no exec attempted); and no entry at all for a resolved integration that declares no probe. The slice is rebuilt each pass, so a removed integration leaves no stale entry. The whole pass is best-effort — a lookup/probe failure never fails the reconcile.
- **Re-probe cadence + backoff** — there is no watch on the referenced resource, so a Running workspace requeues to re-probe. The base period defaults to **5 minutes** (health is reported, not gating, so it need not be tight; author-overridable via `statusProbe.periodSeconds`). A continuously-failing integration backs off exponentially from the base, **capped at 30 minutes**, derived statelessly from how long the verdict has been failing (`now − LastTransitionTime`) — no failure counter is stored.
- **Requeue composability** — `getShorterInterval` combines the idle-shutdown requeue and the (per-integration shortest) probe cadence, treating a 0 delay as "no timer" so it never cancels the other.
- **`ExecInPodWithStderr`** — `ExecInPod` now delegates to it so the prober can surface stderr in failure messages; the existing `ExecInPod` signature/behavior is unchanged.

### Expected first-cycle behavior

A workspace reaches Running as soon as its Deployment is rollout-ready, which can be before an integration-injected sidecar finishes starting, so the first probe may briefly report `ProbeFailed`. This is by design (report-only) and self-corrects on the next cycle — deliberately **not** suppressed with a startup grace window, which would also hide genuine early failures.

### Status churn

An unchanged verdict preserves its prior `LastTransitionTime` (conditions compared on **Status + Reason**, deliberately excluding Message so a run-to-run varying stderr does not churn the timestamp). Holding the timestamp stable across a failing streak is also what the backoff clock counts from.

## What's in this PR

| File | What |
|---|---|
| `internal/controller/integration_prober.go` | `IntegrationProber` (`FindRunningPod` + `Probe`), probe reasons, period/timeout resolution, `integrationProbeRequeueAfter` backoff, `preserveConditionTimestamp`. |
| `internal/controller/state_machine.go` | `probeIntegrationStatus` driver, `getShorterInterval`, Running-path wiring. |
| `internal/controller/pod_exec_util.go` | `ExecInPodWithStderr` + `ExecInPod` delegating to it. |
| `internal/controller/conditions.go`, `constants.go` | `IntegrationConditionTypeReady`, probe reasons/state constants, 5-min default period, 30-min backoff cap. |
| `internal/controller/workspace_controller.go` | Wires the prober into the controller. |
| `internal/controller/integration_prober_test.go` | Prober + driver + pure-function unit tests. |

## Testing

Unit tests cover:
- **Prober** — exec verdicts (exit-zero → Ready, nonzero → Degraded with stderr in the message); `FindRunningPod` (running found / none → error / terminating skipped); `integrationProbeRequeueAfter` backoff (ready → base, just-failed → base, one period → > base, long → 30-min cap); default/override period resolution.
- **Driver (`probeIntegrationStatus`)** via a mock prober — the branch matrix (Ready / NotResolved / no-probe-skip / ProbeError), the **single pod lookup shared across N integrations**, `PodNotFound` reported with no exec attempted, empty-refs clearing stale status, and unchanged-verdict timestamp preservation end-to-end.
- **Pure functions** — `getShorterInterval` (0 = no-timer, commutativity) and `preserveConditionTimestamp` (Status+Reason compared, Message excluded).

```
go test ./internal/controller/ -run 'Probe|Integration|ShorterInterval|PreserveConditionTimestamp' -count=1
```


dist/chart/values.yaml
```diff
@@ -179,6 +179,14 @@ idleShutdown:
   # clamped up to 1s to avoid saturating the controller at scale.
   checkInterval: "5m"
 
+# [INTEGRATION PROBE]: Report-only integration status probe configuration
+integrationProbe:
+  # Base cadence for report-only integration status probes on running workspaces. Must be a valid Go
+  # duration string with a unit (e.g. "5m", "30s"); a bare number like "5" is invalid.
+  # Leave empty to fall back to the controller default (5m). Values below 1s are
+  # clamped up to 1s to avoid saturating the controller at scale.
+  period: "5m"
```
JGuinegagne: non-blocking: could we use `checkInterval` for consistency?

internal/controller/integration_prober.go
```diff
@@ -0,0 +1,306 @@
...
+	// MinIntegrationProbePeriod is the floor a configured --integration-probe-period is clamped up to, so
+	// a misconfigured tiny value cannot hot-loop the reconciler. Mirrors MinIdleCheckInterval.
+	MinIntegrationProbePeriod = 1 * time.Second
+
+	// maxProbeMessageBytes bounds the probe verdict message (built from the command's stderr/stdout)
+	// before it is written to a condition. The CRD caps conditions[].message at 32768 bytes, so a
+	// chatty probe that dumps kilobytes to stderr would otherwise fail the status write outright; we
+	// keep it small (a few KB is plenty to diagnose a failure) and well under the CRD ceiling.
+	maxProbeMessageBytes = 2048
+)
```
JGuinegagne: consider refactoring to [conditions](https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/controller/conditions.go) and [constants](https://github.com/jupyter-infra/jupyter-k8s/blob/main/internal/controller/constants.go).

internal/controller/pod_exec_util.go
```diff
@@ -51,18 +51,18 @@ func NewPodExecUtil() (*PodExecUtil, error) {
 	}, nil
 }
 
-// ExecInPod executes a command in a specific container of a pod with optional stdin input
-func (p *PodExecUtil) ExecInPod(ctx context.Context, pod *corev1.Pod, containerName string, cmd []string, stdin string) (string, error) {
+// ExecInPodWithStderr executes a command in a container and returns stdout, stderr, and error
+// separately. Used by the integration status probe so a failing probe can surface the command's
+// stderr in workspace.status.integrationStatuses[].message.
+func (p *PodExecUtil) ExecInPodWithStderr(ctx context.Context, pod *corev1.Pod, containerName string, cmd []string, stdin string) (stdoutStr, stderrStr string, err error) {
```
JGuinegagne, line 57: sanity-check: why do you need to update pod exec?

HariniNarayanan: ExecInPod's existing callers (idle detector, SSM adapter) only use a failure internally — the idle probe branches on the exit code in the error, and the SSM calls just check pass/fail — so stdout + the exec error was enough, and stderr was logged at V(1) for developers. The integration status probe is different: it writes a user-facing verdict into status.integrationStatuses[].message, so it needs the command's actual stderr (e.g. "cannot connect to GCS") — a bare "exited non-zero" isn't diagnostic for the reader. I added ExecInPodWithStderr (returns stdout, stderr, err separately); ExecInPod now delegates to it and returns just (stdout, err), so existing callers are unchanged.

internal/controller/integration_prober.go
HariniNarayanan: Done

dist/chart/values.yaml
HariniNarayanan: Done

internal/controller/state_machine.go
```diff
@@ -341,6 +360,107 @@ func (sm *StateMachine) reconcileDesiredRunningStatus(
...
+		statuses = append(statuses, s)
+	}
+	for i := range workspace.Spec.IntegrationTemplateRefs {
+		ref := &workspace.Spec.IntegrationTemplateRefs[i]
+		// An integration with no frozen record has not resolved yet -- its first-attach capture failed
+		// (e.g. the referenced resource does not exist, or the template is broken). The freeze reconcile
+		// (which ran earlier this reconcile) already logged the detailed cause; surface a Degraded entry
+		// so an admin sees the unresolved integration on the Workspace status rather than only in logs.
+		if findResolvedIntegration(&workspace.Status, ref.Name) == nil {
+			appendStatus(buildIntegrationStatus(ref.Name, false, IntegrationReasonNotResolved,
```
henrywa2, line 415: should we record an event as well?

henrywa2 (approved):

psnilesh (approved):

internal/controller/state_machine.go
HariniNarayanan: Yes can do, will do it as a follow up as this PR is already approved

PR #425, opened 2026-06-25
test: add e2e coverage for the workspace integration feature

HariniNarayanan:
## Summary

End-to-end coverage for the **values-in-status integration** on a real cluster (Kind in CI): the workspace controller resolves the referenced `RayCluster`, freezes the substitutions into `status.resolvedIntegrations[]`, builds the Deployment from the frozen values (sidecar + merged env + `shareProcessNamespace`), probes the frozen `statusProbe` into `status.integrationStatuses[]`, and garbage-collects the workspace-owned Deployment on delete.

> **Stacked PR.** Depends on the whole series (#418 API, #419 resolver, #421 validation, #423 freeze, #424 status probe) — review/merge those first.
>
> Note: there is no `WorkspaceIntegration` child object — the frozen values live on the workspace's own `status`.

## Scenarios

| Spec | What it proves |
|---|---|
| **Resolvable integration** | Freeze resolves the RayCluster head service into `status.resolvedIntegrations[]`; the Deployment gets the resolved `ray-sidecar` (image + args) and merged primary-container env; `shareProcessNamespace` is OR-reduced onto the pod; `status.integrationStatuses[]` reports ready. |
| **External drift replays** | Mutating the referenced RayCluster (unchanged token) replays the frozen values without rolling the pod — the Deployment's `metadata.generation` stays stable (the roll signal; `pod-template-hash` is a ReplicaSet/Pod label, not on `deployment.spec.template`, so it is not used). |
| **Failing statusProbe (report-only)** | A `statusProbe` that fails flips `status.integrationStatuses[].state` to `Degraded`/`ProbeFailed` while the workspace's own `Available` condition stays `True` — the non-gating contract. |
| **Switch cluster** | Re-applying the workspace pointed at a second RayCluster changes `parametersVersion`, re-resolves, and flips the Deployment env to the new cluster's head service. |
| **Missing resource** | A referenced RayCluster that doesn't exist is fail-closed and non-fatal: the workspace becomes Available with a base-only pod (no sidecar), and the unresolved integration surfaces a `Degraded`/`NotResolved` entry on `status.integrationStatuses[]` (visible to an admin, not just logs). |
| **Garbage collection** | Deleting the workspace cascades deletion of its owned Deployment via the ownerReference. |

### Design notes

- **Templated probe (regression guard).** The `statusProbe` command is a `sh -c '[ "<resolved head svc>" = demo-cluster-head-svc ]'` that compares the **resolved** `{{ resource }}` value against the known head service name — it exits 0 (Ready) only if the operator actually resolved the probe command from the frozen values. A bare `command: ["true"]` would pass regardless and mask a probe-resolution regression; this fixture fails the spec if the operator ever execs a literal `{{ resource … }}` string. Uses `sh`+`test` (no Ray binary needed), so it stays deterministic in Kind.
- **RayCluster fixture.** A minimal `RayCluster` CRD + instances give `resourceRefs` a real target. The CRD uses `x-kubernetes-preserve-unknown-fields: true` and intentionally does **not** declare a status subresource, so the fixture's inline `.status` (the head service name the resolver reads) survives `kubectl apply` — a comment guards this trap.

## What's in this PR

| File | What |
|---|---|
| `test/e2e/workspace_integration_test.go` | The Ordered specs + fixture helpers. |
| `test/e2e/static/integration/raycluster-crd.yaml` | Minimal RayCluster CRD (no status subresource — see comment). |
| `test/e2e/static/integration/raycluster.yaml`, `raycluster-2.yaml`, `raycluster-drifted.yaml` | RayCluster instances (switch-cluster + drift). |
| `test/e2e/static/integration/ray-integration.yaml`, `ray-integration-failprobe.yaml` | The integration templates (sidecar + env + statusProbe + `parameters`); failprobe variant for the report-only spec. |
| `test/e2e/static/integration/workspace-*.yaml` | Workspace fixtures (with-integration / switch-cluster / missing-resource / failprobe). |

## Running

```
make test-e2e            # provisions Kind, deploys the operator, runs the suite
# or, against an existing cluster with assets restored:
go test -tags=e2e ./test/e2e/ -run 'Workspace Integration'
```


test/e2e/static/integration/service-cache-2.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: other-cache
+  namespace: default
+spec:
+  clusterIP: None
+  ports:
+    - name: redis
```
henrywa2: curious, why are there redis ports?

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,551 @@
...
+			applyIntegrationFixture("service-integration")
+
+			By("creating a workspace referencing the integration template")
+			createWorkspaceForTest("workspace-with-integration", groupDir, "")
+
+			By("waiting for the workspace to become available")
+			WaitForWorkspaceToReachCondition(
+				workspaceName, workspaceNamespace, controller.ConditionTypeAvailable, ConditionTrue)
+
+			By("verifying the resolved values were FROZEN into status.resolvedIntegrations")
```
henrywa2: can you get each item once (workspace, deployment, etc) and get the entire spec and validate against that, instead of making separate calls?

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,551 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/jupyter-infra/jupyter-k8s/internal/controller"
+	"github.com/jupyter-infra/jupyter-k8s/test/utils"
+)
+
+// Workspace Integration (Approach 2b: values-in-status freeze).
```
psnilesh: nit: can remove `Approach 2b:` ...

test/e2e/static/integration/service-cache-2.yaml
HariniNarayanan: it's just arbitrary name for this port for testing - I just called it cache now. Need to test that the port is frozen

test/e2e/workspace_integration_test.go
HariniNarayanan: Fixed.

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,551 @@
...
+
+			By("waiting for the workspace to become available")
+			WaitForWorkspaceToReachCondition(
+				workspaceName, workspaceNamespace, controller.ConditionTypeAvailable, ConditionTrue)
+
+			By("verifying the resolved values were FROZEN into status.resolvedIntegrations")
+			frozenName, err := kubectlGet("workspace", workspaceName, workspaceNamespace,
+				"{.status.resolvedIntegrations[?(@.name=='service-integration')].name}")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(frozenName).To(Equal("service-integration"),
```
psnilesh: nit: previous like is  filtering resolvedIntegrations by the same name so this will always be true. Perhaps a more intuitive assertion would `len({.status.resolvedIntegrations[?(@.name=='service-integration')}) == 1`?

internal/controller/resource_manager_integration.go
```diff
@@ -129,7 +129,15 @@ func (rm *ResourceManager) reconcileIntegrations(
 
 	// Record the reconciled set on the in-memory workspace so the build renders from it and the reconcile's
 	// single status write persists it. Pruned records (integration removed from spec) fall out naturally.
-	workspace.Status.ResolvedIntegrations = resolvedIntegrations
+	// Normalize an empty result to nil (not a zero-length slice): resolvedIntegrations is omitempty, so a
+	// freshly-read workspace has it nil; assigning []T{} would make the status-write diff (reflect.DeepEqual)
+	// see a phantom change and churn the status subresource on every reconcile of a no-integration workspace
+	// (this runs unconditionally now). nil round-trips cleanly and still prunes a removed integration.
```
JGuinegagne: non-blocking: try editing the AI-generated inline comment before committing please, this comment could fit in 1 line.

test/e2e/static/integration/service-integration-edited.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: workspace.jupyter.org/v1alpha1
+kind: WorkspaceIntegrationTemplate
+metadata:
+  name: service-integration
+  namespace: default
```
JGuinegagne: nit: we usually place the template / admin-resources in the shared ns

test/e2e/static/integration/workspace-no-integration.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: workspace.jupyter.org/v1alpha1
+kind: Workspace
+metadata:
+  name: workspace-with-integration
+  namespace: default
+spec:
+  # Same metadata.name as workspace-with-integration.yaml but with NO integrationTemplateRefs, so applying
+  # this UPDATES the existing workspace to drop its only integration. The controller must prune the frozen
+  # status entry and re-render the pod WITHOUT the sidecar (base-only), while the workspace stays Available.
```
JGuinegagne: we usually use json patch for updates, example [here](https://github.com/jupyter-infra/jupyter-k8s/blob/main/test/e2e/static/access-probe/patch-workspace-switch-to-failure-strategy.json)

test/e2e/static/integration/workspace-switch-cluster.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: workspace.jupyter.org/v1alpha1
```
JGuinegagne: same comment, use a patch please

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+
+// containerByName returns the named container from a pod spec, or nil.
+func containerByName(spec corev1.PodSpec, name string) *corev1.Container {
+	for i := range spec.Containers {
+		if spec.Containers[i].Name == name {
+			return &spec.Containers[i]
+		}
+	}
+	return nil
+}
```
JGuinegagne: move these utils to `helpers_integration` please

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+
+			By("editing the template in place (adds a resolved env var; parameters unchanged)")
+			applyIntegrationFixture("service-integration-edited")
+
+			By("nudging the workspace to reconcile so it observes the new template version")
+			// The controller watches Workspace (and owns Deployment/Service/PVC) but does NOT watch
+			// WorkspaceIntegrationTemplate, so a template edit alone does not enqueue the referencing
+			// workspace -- the re-resolve otherwise waits for the next periodic reconcile (the integration
+			// probe cadence, 5m in the chart). A real admin gets the update on that timer; here we annotate
+			// the workspace to trigger an immediate reconcile so the spec is deterministic without a 5m wait.
```
JGuinegagne, line 328: interesting - should the workspace controller watch integrationTemplates then?

Fine to keep this way as this would increase scope/complexity, just would like to hear your thoughts.

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+
+			By("verifying the pod was re-rendered with the edit's resolved env var")
+			Eventually(func(g Gomega) {
+				marker, err := kubectlGet("deployment", deploymentName, workspaceNamespace,
+					"{.spec.template.spec.containers[0].env[?(@.name=='CACHE_EDITED_MARKER')].value}")
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(marker).To(Equal("shared-cache-edited"),
+					"the edited template's new env var must be resolved from the Service and rendered onto the pod")
+			}).WithTimeout(2 * time.Minute).WithPolling(5 * time.Second).Should(Succeed())
+		})
```
JGuinegagne, line 351: I think what we functionally want to test here is that an `integrationTemplate` drift that would cause the workspace to be restarted (ie require a rollout of the ws deployment) does NOT do that thanks to the frozen spec mechanism. Would it be possible to write the test this way?

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+// the deployment's pod template -- used for the fail-closed base-only path.
+func verifyNoSidecarInDeployment(deploymentName, namespace string) {
+	GinkgoHelper()
+	Consistently(func(g Gomega) {
+		names, err := kubectlGet("deployment", deploymentName, namespace,
+			"{.spec.template.spec.containers[?(@.name=='cache-proxy')].name}")
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(names).To(BeEmpty(), "no cache-proxy overlay must be applied when capture failed")
+	}, 15*time.Second, 5*time.Second).Should(Succeed())
+}
```
JGuinegagne: move to `helpers_integration` please

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,583 @@
...
+			By("verifying the frozen value is REPLAYED and the deployment stays byte-stable over time")
+			// The env must remain the pre-drift frozen value, and the deployment generation must not
+			// change -- a bump would mean the operator re-rendered the pod template and rolled the pod.
+			// One deployment read per tick covers both checks.
+			Consistently(func(g Gomega) {
+				var deploy appsv1.Deployment
+				g.Expect(kubectlGetInto("deployment", deploymentName, workspaceNamespace, &deploy)).To(Succeed())
+				g.Expect(envValue(containerByName(deploy.Spec.Template.Spec, controller.PrimaryContainerName), "CACHE_PORT")).
+					To(Equal("6379"), "drift must be ignored: the frozen port must be replayed")
+				g.Expect(deploy.Generation).To(Equal(preGen), "deployment generation must not change (no roll)")
```
JGuinegagne: **[Reliability / test-efficacy] Does the drift spec actually observe a reconcile within the 30s window?**

This spec drifts `shared-cache`'s port, then `Consistently(30s, 5s)` asserts `CACHE_PORT == "6379"` and `deploy.Generation == preGen`.

My reading of the reconcile cadence suggests nothing enqueues a reconcile during that window:
- `shared-cache` is a standalone fixture Service — not owned or watched by the controller (`Owns` covers only the workspace-owned Deployment/Service/PVC).
- The workspace's last requeue after becoming Available is the integration probe cadence, which defaults to `integrationProbe.checkInterval: 5m` and isn't overridden by the e2e helm install; idle-shutdown is off, so `getShorterInterval(0, 5m) = 5m`.

If that's right, the controller never re-renders the pod in 30s, so `Generation`/`CACHE_PORT` can't change **regardless of whether frozen-replay works** — the spec would pass even if replay were broken. Contrast with the edit spec, which correctly calls `touchWorkspaceToForceReconcile` before asserting.

**Could you verify whether a reconcile actually fires inside the 30s window?** If not, consider forcing one (`touchWorkspaceToForceReconcile`) *after* drifting the Service and *then* asserting the port is still `6379` and generation is unchanged — that exercises the replay-over-drift path rather than the absence of a trigger. If there *is* a sub-30s trigger I've missed (e.g. from `workspacePodWatching.enable=true`), disregard.

internal/controller/resource_manager_integration.go
HariniNarayanan: done

test/e2e/static/integration/service-integration-edited.yaml
HariniNarayanan: done

test/e2e/static/integration/workspace-no-integration.yaml
HariniNarayanan: done

test/e2e/static/integration/workspace-switch-cluster.yaml
HariniNarayanan: done

test/e2e/workspace_integration_test.go
HariniNarayanan: I actively considered making this change a while ago - but we use labels to track and we need a reverse index on a list + number of integrations to track. I think the wait is acceptable. I can add it add it as an issue that we can choose to address

test/e2e/workspace_integration_test.go
HariniNarayanan: Done

test/e2e/workspace_integration_test.go
HariniNarayanan: done

test/e2e/workspace_integration_test.go
HariniNarayanan: You're right. This was a miss. I made sure the pod did roll when the template was updated using touchWorkspaceToForceReconcile but missed updating this because it was always passing. Fixed it

test/e2e/workspace_integration_test.go
```diff
@@ -0,0 +1,471 @@
...
+			Expect(markerBefore).To(BeEmpty(), "the edited-marker env must not exist before the template edit")
+
+			By("editing the template in place (adds a resolved env var; parameters unchanged)")
+			applyIntegrationFixture("service-integration-edited")
+
+			By("nudging the workspace to reconcile so it observes the new template version")
+			// The controller watches Workspace (and owns Deployment/Service/PVC) but does NOT watch
+			// WorkspaceIntegrationTemplate, so a template edit alone does not enqueue the referencing
+			// workspace -- the re-resolve otherwise waits for the next periodic reconcile (the integration
+			// probe cadence, 5m in the chart). A real admin gets the update on that timer; here we annotate
```
henrywa2, line 327: How does an admin get this update?

henrywa2 (dismissed):

JGuinegagne (approved):

PR #430, opened 2026-07-08
chore: upgrade to go 1.26

JGuinegagne:
This PR:
- upgrades `go` version: `1.24.5` > `1.26.5`
- upgrades linter version `2.4.0` > `2.12.2` (latest)
- pins `jupyter-k8s-plugin` dependency to the newly released version `v0.1.0`, following [PR4](https://github.com/jupyter-infra/jupyter-k8s-plugin/pull/4)

### Testing
- will rely on E2E test CI


henrywa2 (dismissed):

.github/workflows/lint.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
       - name: Install golangci-lint
         run: |
-          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.4.0
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v2.12.2/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.2
```
earaghbidikashani, line 24: this approach pipes a remote script into sh without checksum verification. 
Claude suggested: Consider using the official golangci-lint [GitHub Action](https://github.com/golangci/golangci-lint-action) instead, which handles caching and version pinning

.golangci.yml
```diff
@@ -45,6 +45,14 @@ linters:
           - dupl
           - lll
         path: internal/*
+      # Upstream deprecations — kubebuilder scaffolding (scheme.Builder) and
+      # old events API (GetEventRecorderFor). Tracked for a future migration.
+      - linters:
```
earaghbidikashani, line 50: why are we suppressing these deprecations

go.mod
```diff
@@ -1,133 +1,137 @@
 module github.com/jupyter-infra/jupyter-k8s
 
-go 1.24.5
+go 1.26.0
```
earaghbidikashani: why not updating to 1.26.5? this is an application I believe

description also says 1.26.5

go.mod
```diff
@@ -1,133 +1,137 @@
...
+	github.com/onsi/gomega v1.40.0
 	github.com/stretchr/testify v1.11.1
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.34.0
-	k8s.io/apimachinery v0.34.0
-	k8s.io/apiserver v0.34.0
-	k8s.io/client-go v0.34.0
-	sigs.k8s.io/controller-runtime v0.21.0
-	sigs.k8s.io/controller-tools v0.19.0
+	k8s.io/api v0.36.2
```
earaghbidikashani, line 14: do we need to verify hyperpod compatibility verification with these version jumps? (we will verify it anyway in every addon release but we should be super diligent in the next release I think)

go.mod
JGuinegagne: Nice catch, this should be 1.26.5, thank you

.github/workflows/lint.yml
JGuinegagne: Yeah, it also provided this as an option.
How about we create a GH issue to upgrade across our repos? For now, i would prefer to keep consistency.

go.mod
JGuinegagne: Yeah definitely. I asked it to "upgrade all dependencies", but we should test against the latest K8s version supported by HP and against the older K8s version we want to support.

.golangci.yml
JGuinegagne: we need to upgrade `kubebuilder`. Let's track separately.

.github/workflows/lint.yml
JGuinegagne: filed as #431

bhanderig (approved):

earaghbidikashani (approved):

PR #434, opened 2026-07-10
fix: use ClusterRole for extension-apiserver-authentication access

aakashmandavilli96:
**Description**

Replace the `kube-system` RoleBinding with a ClusterRole/ClusterRoleBinding for `extension-apiserver-authentication` configmap access.

**Motivation**

The controller reads the extension-apiserver-authentication configmap from kube-system at startup — both the aggregated API server `(RecommendedOptions.ApplyTo())` and the secure metrics server require it. Without access, the controller-manager crashes on startup with:
- configmaps "extension-apiserver-authentication" is forbidden: User
- "system:serviceaccount:jupyter-k8s-system:jupyter-k8s-controller-manager"
- cannot get resource "configmaps" in API group "" in the namespace "kube-system"

The current fix uses a RoleBinding in kube-system, which requires the installer to have cross-namespace permissions. This fails for any namespace-scoped deployment tool — EKS addons, ArgoCD with restricted RBAC, OLM, fleet managers, etc.
A ClusterRole/ClusterRoleBinding with resourceNames: ["extension-apiserver-authentication"] grants the same least-privilege access using cluster-scoped resources, without assuming the installer can write to kube-system. This is compatible with all deployment scenarios.

**Testing Done**
- Added helm test verifying no kube-system references and ClusterRole/ ClusterRoleBinding presence
- Confirmed on two fresh HyperPod quick-create clusters that the RoleBinding approach fails and ClusterRole approach succeeds
- helm lint passes

**Backwards Compatibility Criteria**
Same effective permissions (read configmaps/extension-apiserver-authentication). Existing clusters with the old RoleBinding will continue working; new installs will use the ClusterRole approach. The old RoleBinding becomes orphaned on upgrade but is harmless (read-only permission).

### DIFF
--- a/config/rbac/extension_api_auth_binding.yaml
+++ b/config/rbac/extension_api_auth_binding.yaml
@@ -1,16 +1,27 @@
-## This RoleBinding is applied separately from kustomize because it must live
-## in kube-system, and kustomize's global namespace transform would override that.
-## See Makefile deploy/undeploy targets.
+## ClusterRole/ClusterRoleBinding for reading extension-apiserver-authentication
+## configmap. Uses cluster-scoped resources instead of a kube-system RoleBinding
+## for compatibility with environments that lack cross-namespace permissions
+## (e.g., EKS addon framework).
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRole
 metadata:
-  name: jupyter-k8s-extension-api-auth
-  namespace: kube-system
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: extension-apiserver-authentication-reader
+  name: jupyter-k8s-read-auth-configmap
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["extension-apiserver-authentication"]
+    verbs: ["get", "list", "watch"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: jupyter-k8s-read-auth-configmap
 subjects:
-- kind: ServiceAccount
-  name: jupyter-k8s-controller-manager
-  namespace: jupyter-k8s-system
+  - kind: ServiceAccount
+    name: jupyter-k8s-controller-manager
+    namespace: jupyter-k8s-system
+roleRef:
+  kind: ClusterRole
+  name: jupyter-k8s-read-auth-configmap
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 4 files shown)


JGuinegagne (approved):

config/rbac/extension_api_auth_binding.yaml
```diff
@@ -1,16 +1,27 @@
-## This RoleBinding is applied separately from kustomize because it must live
-## in kube-system, and kustomize's global namespace transform would override that.
-## See Makefile deploy/undeploy targets.
+## ClusterRole/ClusterRoleBinding for reading extension-apiserver-authentication
+## configmap. Uses cluster-scoped resources instead of a kube-system RoleBinding
+## for compatibility with environments that lack cross-namespace permissions
+## (e.g., EKS addon framework).
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRole
```
henrywa2, line 6: Why do we need ClusterRole here? configmap is not cluster scoped resource

config/rbac/extension_api_auth_binding.yaml
```diff
@@ -1,16 +1,27 @@
...
 metadata:
-  name: jupyter-k8s-extension-api-auth
-  namespace: kube-system
+  name: jupyter-k8s-read-auth-configmap
```
henrywa2, line 8: why are we changing the ns?

config/rbac/extension_api_auth_binding.yaml
joshuatowner: this is an EKS-specific issue, the EKS addons cannot create resources in other namespaces. 

@aakashmandavilli96  agree with the recommendation of doing this in our helm chart tho

config/rbac/extension_api_auth_binding.yaml
aakashmandavilli96: This resource is now a `ClusterRoleBinding` (cluster-scoped), so it doesn't belong to any namespace. The old `RoleBinding` lived in `kube-system`, but the EKS addon framework can't create resources there, which caused the controller to crash. `ClusterRoleBinding` gives the same read access without requiring cross-namespace install permissions.

henrywa2 (approved):

config/rbac/extension_api_auth_binding.yaml
aakashmandavilli96: @jjtowner Just to clarify — this isn't EKS-specific. Any installer that's namespace-scoped (EKS addons, ArgoCD, OLM, etc.) would fail on the cross-namespace RoleBinding in kube-system. The ClusterRole approach works universally without widening permissions.

- The AWS repo (charts/aws-hyperpod/) deploys the sidecar, auth middleware, and workspace templates. It has no RBAC resources and no dependency on the upstream operator chart.
- The extension-apiserver-authentication access is needed by the controller at startup, which is deployed by the upstream jupyter-k8s chart — so the RBAC must ship with it.

joshuatowner (approved):

PR #438, opened 2026-07-20
fix: idle shutdown enforcement

### LINKED ISSUE #437: WorkspaceTemplate idleShutdownOverrides policy is not enforced on Workspaces
### Problem

A `WorkspaceTemplate` can declare an `idleShutdownOverrides` policy (`allow`, `minIdleTimeoutInMinutes`, `maxIdleTimeoutInMinutes`), but this policy is **never applied to Workspaces**. The type is defined and the template webhook locks the fields against edits (`idleShutdownAllowOverrideChanged` / `idleShutdownTimeoutBoundsChanged` in `template_webhook.go`), but no code path evaluates a Workspace against it:

- The workspace validating webhook (`template_validator.go`) has no idle-shutdown check.
- The mutating defaulter (`lifecycle_defaulter.go`) only copies `defaultIdleShutdown` when the field is nil — it never clamps or forces a user-supplied value.
- The controller reads `idleTimeoutInMinutes` verbatim; no clamping.
- The scaffolded violation constants `ViolationTypeIdleShutdownOverrideNotAllowed` and `ViolationTypeIdleShutdownTimeoutOutOfBounds` (`validation_types.go`) are dead code.

As a result a user can `kubectl apply` a Workspace that disables idle shutdown (`enabled: false` or omitted) or sets a timeout outside the template's declared bounds, and admission accepts it — defeating cost-control policies.

### Desired behavior

When a Workspace references a template with an `idleShutdownOverrides` policy:
- if `allow: false`, the Workspace must not disable idle shutdown;
- `idleTimeoutInMinutes` must fall within `[min, max]`.

Violations should be surfaced at admission time.

JGuinegagne:
This PR wires up the `template.spec.idleShutdownOverrides` field to enforce bounds that users select for a `workspace.spec.idleShutdown` when referencing the template.

Closes: #437 

### User experience
- on create-workspace w/ `templateRef`
    - if `template.spec.idleShutdownOverrides.allow = False`:
        - `workspace.spec.idleShutdown` _must_ match exactly `template.spec.defaultIdleShutdown`
        - except that if `template.spec.idleShutdownOverrides.minIdleTimeoutInMinutes` or/and `.maxIdleTimeoutInMinutes` are set, then allow to override `workspace.spec.idleShutdown.idleTimeoutInMinutes`
        - if either of `minIdleTimeoutInMinutes` or `maxIdleTimeoutInMinutes` is _not_ set, uses default for that bound
    - if `template.spec.idleShutdownOverrides.allow = True`:
        - enforce only bounds on `idleTimeoutInMinutes` if `minIdleTimeoutInMinutes` and / or `maxIdleTimeoutInMinutes` are set AND workspace.spec enabled idleShutdown
- same logic on update workspace
- fail fast if template defines `template.spec.idleShutdownOverrides.allow = False` but omits `template.spec.defaultIdleShutdown`

### Implementation
1. add `idle_shutdown_validator` and wire-up in workspace validation webhook
2. add idleShutdown overrides <-> default consistency in template validation webhook
3. update documentation: `concepts/templates/bounds` and `dive-deeper/webhooks/workspace-validation`

### Testing
- added E2E tests for `idleShutdownOverrides` enforcement (+ve and -ve) on workspace admission
- added E2E tests for `idleShutdownOverrides` <-> `defaultIdleShutdown` on template admission

### DIFF
--- a/internal/webhook/v1alpha1/idle_shutdown_validator.go
+++ b/internal/webhook/v1alpha1/idle_shutdown_validator.go
@@ -1,0 +1,143 @@
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package v1alpha1
+
+import (
+	"fmt"
+	"reflect"
+
+	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+)
+
+// validateIdleShutdownOverrides enforces the template's idleShutdownOverrides policy.
+//
+// The policy has two orthogonal knobs:
+//
+//   - Allow governs the STRUCTURE. When Allow=false, the workspace may not deviate from the
+//     template's defaultIdleShutdown: every field must match the default EXCEPT
+//     idleTimeoutInMinutes. When Allow is unset or true (the kubebuilder default), the
+//     workspace may freely choose its idle shutdown config, including disabling it.
+//
+//   - MinIdleTimeoutInMinutes/MaxIdleTimeoutInMinutes bound the timeout VALUE. Declared bounds
+//     are enforced whenever the workspace has idle shutdown enabled, regardless of Allow — so
+//     an admin can permit overrides yet still cap the timeout.
+//
+// Missing-bound fallback differs by intent:
+//
+//   - Allow=false (locked window): an unset bound falls back to the default timeout, pinning
+//     the unspecified side to the default ("the only wiggle room is what I explicitly granted").
+//   - Allow=true/unset (permissive): an unset bound means unbounded on that side.
+//
+// The structural lock requires a defaultIdleShutdown to compare against; without one there is
+// no baseline to match, so only the timeout bounds apply.
+func validateIdleShutdownOverrides(workspace *workspacev1alpha1.Workspace, template *workspacev1alpha1.WorkspaceTemplate) []TemplateViolation {
+	policy := template.Spec.IdleShutdownOverrides
+	if policy == nil {
+		return nil
+	}
+
+	allowOverride := policy.Allow == nil || *policy.Allow
+	def := template.Spec.DefaultIdleShutdown
+	ws := workspace.Spec.IdleShutdown
+
+	var violations []TemplateViolation
+
+	// Structural lock: when overrides are not allowed, the workspace must match the template
+	// default on every field except the timeout. Requires a default baseline to compare against.
+	if !allowOverride && def != nil {
+		if ws == nil {
+			// Overrides are not allowed but the workspace carries no idle shutdown config,
+			// which drops the template's policy entirely.
+			return []TemplateViolation{{
+				Type:    ViolationTypeIdleShutdownOverrideNotAllowed,
+				Field:   "spec.idleShutdown",
+				Message: fmt.Sprintf("Template '%s' does not allow overriding idle shutdown; idleShutdown must match the template default", template.Name),
+				Allowed: "match template defaultIdleShutdown",
+				Actual:  "unset",
+			}}
+		}
+
+		// Compare copies with the timeout zeroed so a permitted timeout difference doesn't trip
+		// the equality check.
+		wsCopy := ws.DeepCopy()
+		defCopy := def.DeepCopy()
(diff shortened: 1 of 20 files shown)


henrywa2:
What is the use of idleShutdownOverrides allow true?

JGuinegagne:
> What is the use of idleShutdownOverrides allow true?

if you set `allow: true`, you allow workspace users to a/ disable `idleShutdown` altogether, b/ change the `detection` settings (in practice, can be used to de-fang idle shutdown and make it _not_ work at all).

So why do we have `allow: true` at all?

That flag allows to set `{min|max}TimeoutMinutes`, and enforce limits on `timeoutMinutes`, while allowing to users to opt out of idleShutdown. Practically, it'll let us build UIs that allow toggle on/off idle shutdown AND allow a slider for the timeoutMinutes within bounds defined by template, while still giving the flexibility for expert users to override the `idleShutdown` config.

henrywa2 (approved):

internal/webhook/v1alpha1/idle_shutdown_validator.go
```diff
@@ -0,0 +1,143 @@
...
+			return []TemplateViolation{{
+				Type:    ViolationTypeIdleShutdownOverrideNotAllowed,
+				Field:   "spec.idleShutdown",
+				Message: fmt.Sprintf("Template '%s' does not allow overriding idle shutdown; idleShutdown must match the template default", template.Name),
+				Allowed: "match template defaultIdleShutdown",
+				Actual:  "unset",
+			}}
+		}
+
+		// Compare copies with the timeout zeroed so a permitted timeout difference doesn't trip
```
earaghbidikashani, line 63: what if timeouts are different?

internal/webhook/v1alpha1/idle_shutdown_validator.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	return violations
+}
+
+// validateIdleTimeoutBounds checks the workspace idle timeout against the policy's declared
+// bounds, applying the missing-bound fallback appropriate to the Allow setting. Returns nil when
+// the timeout is within bounds (or no bound applies to the relevant side).
+func validateIdleTimeoutBounds(
```
earaghbidikashani, line 94: No validation that minIdleTimeoutInMinutes <= maxIdleTimeoutInMinutes

test/e2e/template_enforce_idle_shutdown_test.go
```diff
@@ -0,0 +1,202 @@
...
+		"--ignore-not-found", "--wait=true", "--timeout=120s")
+	_, _ = utils.Run(cmd)
+
+	By("cleaning up templates")
+	cmd = exec.Command("kubectl", "delete", "workspacetemplate", "--all", "-n", SharedNamespace,
+		"--ignore-not-found", "--wait=true", "--timeout=60s")
+	_, _ = utils.Run(cmd)
+
+	By("waiting an arbitrary fixed time for resources to be fully deleted")
+	time.Sleep(1 * time.Second)
```
earaghbidikashani, line 201: static sleep, might cause extra flakiness

internal/webhook/v1alpha1/idle_shutdown_validator.go
JGuinegagne: it's compared later - there are e2e tests and unit tests for this.

internal/webhook/v1alpha1/idle_shutdown_validator.go
JGuinegagne: That would need to go on the template admission webhook, not the workspace's. I'm planning to raise another PR to address #440 and #439 which will tighten template admission, I can add this rule there.

test/e2e/template_enforce_idle_shutdown_test.go
JGuinegagne: We have this pattern in at least 8 other files, the tests are stable. That's just the agent following precedent.

bhanderig (approved):

PR #442, opened 2026-07-21
feat: support Authorization Bearer header in handleBearerAuth

earaghbidikashani:
Adds support for extracting JWT tokens from the `Authorization: Bearer` header in the bearer auth middleware, alongside the existing URL query parameter (`?token=`) method.

   ## Motivation

   WebSocket connections via websocat pass the JWT token as an HTTP header (`-H "Authorization: Bearer <token>"`) rather than a URL query parameter. This avoids tokens appearing in Traefik/ALB access logs. 
   ## Changes

   - `serverroute_bearer_auth.go`: Check `Authorization: Bearer <token>` header first, fall back to `?token=` query param if header not present
   - `serverroute_bearer_auth_test.go`: 3 new tests (valid header, invalid header, empty bearer token)

   ## Backward Compatibility

   Fully backward compatible. Existing web UI flows using `?token=` in the URL continue working unchanged. The Authorization header is only used when present.

   **E2E testing for the Authorization header flow will be covered as part of the full WebSocket connection E2E test, which exercises the complete path: Extension API → websocat with -H "Authorization: Bearer" → Traefik ForwardAuth → proxy → remote access server.**

### DIFF
--- a/internal/authmiddleware/serverroute_bearer_auth.go
+++ b/internal/authmiddleware/serverroute_bearer_auth.go
@@ -13,7 +13,7 @@ import (
 )
 
 // handleBearerAuth handles bearer token authentication requests
-// Takes short lived JWT tokens from URL parameter and exchanges them for 6-hour session cookies
+// Takes short lived JWT tokens from Authorization header or URL parameter and exchanges them for session cookies
 func (s *Server) handleBearerAuth(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet {
 		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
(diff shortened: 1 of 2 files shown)


internal/authmiddleware/serverroute_bearer_auth.go
```diff
@@ -36,12 +36,23 @@ func (s *Server) handleBearerAuth(w http.ResponseWriter, r *http.Request) {
...
+	if authHeader := r.Header.Get(HeaderAuthorization); authHeader != "" {
+		extracted, err := ExtractBearerToken(authHeader)
+		if err != nil {
+			s.logger.Error("Invalid Authorization header", "error", err)
+			http.Error(w, "Invalid Authorization header", http.StatusBadRequest)
+			return
+		}
+		token = extracted
+	} else {
+		token = parsedURL.Query().Get("token")
```
JGuinegagne, line 50: sanity-check: do we need to worry about backward-compatibility here?

Because headers is for sure the best practice, and I would prefer that we support only this path unless URL parameter is absolutely needed.

earaghbidikashani: we should worry backward compatibility, and this is backwards compatible. 

The URL parameter path `?token=` is still needed for web UI because browsers can't set custom headers on navigation requests. The new Authorization header path is for WebSocket connections as we discussed in the design to use headers instead of parameter

JGuinegagne: makes sense, got confused about the package.

This PR targets the operator, not the sidecar repo. Hence I agree we definitely need to worry about backward compatibility.

JGuinegagne (approved):

bhanderig (approved):
LGTM

PR #446, opened 2026-07-23
feat: authorize integration resourceRefs against the requesting user

HariniNarayanan:
## What

Adds an admission-time authorization check so a user cannot reference — via a `WorkspaceIntegrationTemplate`'s `resourceRefs` — a resource they themselves cannot read, and have the operator read it on their behalf. This closes a **confused-deputy privilege escalation**: the operator holds cluster-wide read on integration-referenced resources (e.g. a `RayCluster`) and reads them at reconcile time to inject their values into the workspace pod; without this check, that read happens with the operator's identity regardless of the requesting user's own access.

## How

The `Workspace` integration validator issues a **`get` SubjectAccessReview** for the requesting user against each resolved `resourceRef`:

- **Interleaved into the existing per-ref loop.** For each `integrationTemplateRef` the validator runs the correctness checks (namespace scope → template existence → parameter completeness) and *then*, for non-exempt callers, authorizes the referenced resources — reusing the template already fetched for correctness, so the referenced template is read exactly once per ref.
- **Same resolver as reconcile.** Each `resourceRef`'s name/namespace is resolved with the SAME resolver the controller uses at reconcile, so admission authorizes exactly the object reconcile will fetch (including the resolver's empty-value guards — a blank resolved name is rejected without issuing a review).
- **Fails closed.** The first denied or unevaluable review rejects the workspace.
- **Exempt callers bypass it.** The controller service account and cluster admins (`system:masters` / the configured admin group) skip the permission check — determined locally from the admission request's `UserInfo`, no API call — the same bypass the other permission checks already use. They still get the correctness checks.
- **Create vs update.** On create it runs whenever the caller isn't exempt. On update it runs only when the integration refs actually change (`integrationRefsChanged`), and after the admin/controller bypass — mirroring the other update-path checks and avoiding the wedge where an admin's out-of-band template edit re-validates a live workspace.

Grants the operator `create` on `authorization.k8s.io/subjectaccessreviews`, synced to `config/rbac`, `dist/install.yaml`, and the chart manager role (backed by a `+kubebuilder:rbac` marker, so the generated role is reproducible).

## Testing

- **Unit** (`integration_template_ref_validator_test.go`): a fake API server with a Create-interceptor plays the authorizer — allow / deny / multi-ref / blank-name / exempt-bypass paths.
- **E2E** (`workspace_integration_admission_test.go`, run as impersonated users via `kubectl --as` against the real cluster authorizer and the operator's real SAR-create grant — coverage a fake-authorizer unit test can't provide):
  - **denied** — a user with workspace-create but no `get` on the referenced Service is rejected with the `may not get` verdict (asserted, so the rejection is provably the resource-authorization verdict, not a workspace RBAC denial).
  - **allowed** — a user additionally granted `get` on the Service is admitted.
  - **exempt** — an admin (`system:masters`) is admitted without any Service-get binding, proving the bypass.
- Verified locally: `make verify-generated` clean, controller + webhook suites green, and the three E2E specs green on a fresh Kind cluster.

## Notes

Follow-up to #421 (integration validating webhook) / #423 / #424 in the WorkspaceIntegration stack.

### DIFF
--- a/internal/webhook/v1alpha1/integration_template_ref_validator.go
+++ b/internal/webhook/v1alpha1/integration_template_ref_validator.go
@@ -13,12 +13,16 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	workspacev1alpha1 "github.com/jupyter-infra/jupyter-k8s/api/v1alpha1"
+	"github.com/jupyter-infra/jupyter-k8s/internal/controller"
 )
 
 // IntegrationTemplateRefValidator validates a Workspace's integrationTemplateRefs at admission on three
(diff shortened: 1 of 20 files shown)


internal/webhook/v1alpha1/integration_template_ref_validator.go
```diff
@@ -118,9 +139,116 @@ func (v *IntegrationTemplateRefValidator) validateRef(
...
+		// (mapping.Scope == meta.RESTScopeNamespace) -- a cluster-scoped ref must omit Namespace or the
+		// SubjectAccessReview will not match the cluster-scoped resource.
+		review := &authorizationv1.SubjectAccessReview{
+			Spec: authorizationv1.SubjectAccessReviewSpec{
+				User:   user.Username,
+				Groups: user.Groups,
+				UID:    user.UID,
+				Extra:  convertUserExtraToAuthorization(user.Extra),
+				ResourceAttributes: &authorizationv1.ResourceAttributes{
+					Verb:      "get",
```
psnilesh, line 208: This is tricky. Just having read permissions is not enough to justify giving access for interactive development.  Normally, scientists have broad read access but only limited write access. 

Let me think about this a little bit more.

HariniNarayanan:
Closing this PR.

We've concluded that 
1. we're not actually "getting" we're referencing/creating a connection - the action we're checking for is not accurate
2. we need to introduce a integration reference and do SAR on that. 
3. this means that we cannot support cross namespace resource reference today
4. we'll stick to resource present only in the workspace namespace.

PR #447, opened 2026-07-23
feat: emit edge-triggered Events on integration status transitions

HariniNarayanan:
## What

Emit a Kubernetes **Event** on the `Workspace` when an integration's report-only status probe verdict changes, so a degraded or recovered integration is visible via `kubectl describe workspace` / `kubectl get events` — not only in `status.integrationStatuses[]`.

Follow-up to #424, answering @henrywa2's review question ("should we record an event as well?").

## How

The decision lives in a pure, unit-tested helper `getIntegrationStatusEvent(name, prior, next)` that is **edge-triggered** — it fires only on a verdict **transition**, never on an unchanged re-probe.

Edge-triggering is required here **by necessity**: `probeIntegrationStatus` runs every reconcile (the flat probe cadence, ~5m in the chart, plus every event-driven reconcile), so a naive `recorder.Event` in the not-ready branch would emit a duplicate event **every cycle** while an integration stays degraded and flood the event stream. This differs from the existing state-machine events (`WorkspaceStopped` / `IdleShutdown`), which sit on one-shot terminal transitions and so are naturally single-fire without any transition guard — the report-only probe has no such natural guard, so the helper provides one explicitly.

| Transition | Event |
|---|---|
| newly-seen not-ready · Ready→not-ready · reason change while not-ready | **Warning** `IntegrationDegraded` |
| not-ready → Ready | **Normal** `IntegrationRecovered` |
| unchanged verdict · healthy first attach | **silent** (no event) |

- Message is excluded from the change test (mirrors `preserveConditionTimestamp`), so a failing probe's run-to-run stderr does not retrigger an event.
- Wired at the single `appendStatus` choke point in `probeIntegrationStatus`, guarded on a non-nil recorder.
- **No RBAC/chart change** — the manager `ClusterRole` already grants `events` create.

## Testing

- `TestIntegrationStatusEvent` — 8 transition cases exercising the anti-spam decision directly (silent first-attach, silent steady-state, degrade, recover, reason-change re-warn, nil-next).
- `TestProbeIntegrationStatus_EmitsEdgeTriggeredEvents` — drives the full state-machine probe path with a real `FakeRecorder` over 4 cycles: degraded→**1 Warning**, unchanged→**0**, recovered→**1 Normal**, steady-state→**0**.
- `TestProbeIntegrationStatus_HealthyFirstAttachIsSilent` — a healthy first probe emits nothing.
- Verified locally: `golangci-lint run` and `golangci-lint run ./test/e2e/...` both 0 issues, controller unit suite green, `make verify-generated` clean.

### DIFF
--- a/internal/controller/integration_prober.go
+++ b/internal/controller/integration_prober.go
@@ -210,6 +210,70 @@ func preserveConditionTimestamp(prior, next *workspacev1alpha1.IntegrationStatus
 	}
 }
 
+// integrationEvent describes a Kubernetes Event to record for an integration status transition. A zero
+// value (Type == "") means "no event": the verdict did not transition in a way worth recording.
+type integrationEvent struct {
+	Type    string // corev1.EventTypeNormal or corev1.EventTypeWarning
+	Reason  string // IntegrationEventDegraded / IntegrationEventRecovered
+	Message string
+}
+
+// getIntegrationStatusEvent decides whether an integration's probe-verdict change warrants a Kubernetes
+// Event on the Workspace, and of what kind. It is EDGE-TRIGGERED to avoid event spam on the flat
+// report-only probe cadence: an event fires only on a verdict TRANSITION, never for an unchanged re-probe.
+//
+//   - newly-seen not-ready (no prior), Ready->not-ready, or a reason change while still not-ready
+//     (e.g. PodNotFound->ProbeFailed) -> Warning IntegrationDegraded.
+//   - not-ready -> Ready -> Normal IntegrationRecovered.
+//   - an unchanged verdict (same condition Status+Reason), or a healthy first attach (no prior, Ready)
+//     -> no event: a steady-state healthy integration is silent.
+//
+// Message is deliberately excluded from the change test (mirrors preserveConditionTimestamp): a failing
+// probe's stderr can vary run-to-run without the verdict changing, and must not trigger a fresh event.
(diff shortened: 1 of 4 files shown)


internal/controller/integration_prober.go
```diff
@@ -210,6 +210,70 @@ func preserveConditionTimestamp(prior, next *workspacev1alpha1.IntegrationStatus
...
+	var hadPrior, priorReady bool
+	var priorReason string
+	if prior != nil && len(prior.Conditions) > 0 {
+		hadPrior = true
+		priorReady = prior.Conditions[0].Status == metav1.ConditionTrue
+		priorReason = prior.Conditions[0].Reason
+	}
+
+	if nextReady {
+		// Announce a RECOVERY only (was not-ready -> now Ready). A healthy first attach (no prior) or a
```
henrywa2, line 249: steady-state Ready re-probe stays silent makes sense to me, but should a healthy first attach emit an event?

HariniNarayanan: Intentionally silent on healthy first attach. The happy path is already covered — WorkspaceRunning fires and integrationStatuses[].Ready=True records when it went ready — so a per-integration "ready" event here is redundant. The property I wanted: a healthy fleet emits zero integration events, so anything in the stream is real signal. We do warn on an unhealthy first attach (that needs attention); Normal is only for clearing a prior problem.

henrywa2 (approved):

PR #450, opened 2026-07-28
feat: expose access-strategy sidecar ports on the workspace Service

earaghbidikashani:
The workspace Service only declares the primary container's port (8888). When an access strategy injects a sidecar that listens on its own port, the Service becomes an incomplete description of the pod — and because a Service only forwards ports it explicitly declares, there is nothing to route traffic to.
 
This derives a `ServicePort` for every `containerPort` declared by the access strategy's `additionalContainers`. The rule is generic: no knowledge of any particular sidecar, no feature flag, no hardcoded port.

Goal: let an access strategy expose a sidecar without a code change per sidecar. The immediate consumer is WebSocket-based remote IDE access as a cloud-agnostic alternative to AWS SSM, but nothing here is WebSocket-specific.

  #### Also fixes a pre-existing bug in the Service update path

  `NeedsUpdate` compared the whole `Spec`. `existingService` comes from the API server with defaulted fields (`ClusterIP`, `IPFamilies`, `SessionAffinity`, …); `desiredService` is built in memory with those empty, and `equality.Semantic` does not treat `""` as equal to a defaulted value. So it returned `true` on every reconcile of every available workspace, and `UpdateServiceSpec` then overwrote the whole spec — issuing an `Update` with an empty `clusterIP`, an immutable field.

Now only the operator-owned fields (`Type`, `Selector`, `Ports`) are compared and assigned. The existing test missed it because it built the "existing" Service with `BuildService`, so both sides had an empty `ClusterIP`; the new tests apply API-server defaults first and were confirmed to fail against the pre-fix code.

  ## Notes for reviewers

  - **Port names.** Required, not cosmetic: the API server rejects an unnamed port once a Service has more than one. `ContainerPort.Name` is optional, so it falls back to the container name; both are already valid `ServicePort` names, so they're used verbatim. A `-2`/`-3` suffix resolves collisions, which are reachable since `ContainerPort` names are unique only within a container.
  - **Duplicates are skipped and logged.** Nothing upstream rejects duplicate `containerPort` numbers, so such a strategy can exist; skipping keeps the emitted Service valid.
  - **No behavior change for existing workspaces.** With no access strategy, or one declaring no ports, the Service is byte-identical to before — asserted, not assumed. SSM workspaces are unaffected: the SSM sidecar declares no ports, since the agent dials outbound.
  - **Inert until the charts open the port.** NetworkPolicy is a second gate enforced on the pod, and the policies in `jupyter-k8s-aws` pin ingress to 8888. That's separate chart work; this is safe to merge ahead of it because it's a no-op on its own.
  - **Security.** Grants only in-cluster reachability on a `ClusterIP` Service. External access still requires an IngressRoute behind ForwardAuth *and* an open NetworkPolicy port


internal/controller/service_builder.go
```diff
@@ -31,11 +32,19 @@ func NewServiceBuilder(scheme *runtime.Scheme) *ServiceBuilder {
 	}
 }
 
-// BuildService creates a Service resource for the given Workspace
-func (sb *ServiceBuilder) BuildService(workspace *workspacev1alpha1.Workspace) (*corev1.Service, error) {
+// BuildService creates a Service resource for the given Workspace.
+//
+// The Service describes the ports the workspace pod actually serves. Sidecars injected by the
+// access strategy may listen on their own ports, so those are exposed alongside the Jupyter port;
```
JGuinegagne: `the Jupyter port;` > `the application port;`

internal/controller/service_builder.go
```diff
@@ -56,43 +65,172 @@ func (sb *ServiceBuilder) buildObjectMeta(workspace *workspacev1alpha1.Workspace
...
-		},
+		Ports:    ports,
+	}
+}
+
+// sidecarServicePorts derives the Service ports for containers the access strategy injects
+// into the workspace pod. Each container port a sidecar declares becomes routable; sidecars
+// that declare no ports (for example an agent that only dials outbound) contribute nothing.
+//
+// Ports colliding with the Jupyter port, or duplicated across sidecars, are skipped and logged.
```
JGuinegagne: same comment here.

internal/controller/service_builder.go
```diff
@@ -56,43 +65,172 @@ func (sb *ServiceBuilder) buildObjectMeta(workspace *workspacev1alpha1.Workspace
...
+//
+// Both possible sources are already valid ServicePort names — a ContainerPort name is validated
+// as a port name (at most 15 chars, [-a-z0-9]) and a container name as a DNS-1123 label (at most
+// 63) — while ServicePort.Name allows a full 63-char DNS-1123 label. Neither can therefore be too
+// long or contain an illegal character, so the name is used as-is.
+//
+// Uniqueness does need handling: ContainerPort names are only unique within a single container,
+// so two sidecars may each legally declare the same port name (or default to the same container
+// name). A colliding name is qualified with its port number, which the caller has already made
+// unique across the Service.
```
JGuinegagne: Thin this out please, it reads too much like AI slop.

This whole context could and shout fit in 4-5 lines.

internal/controller/service_builder_test.go
```diff
@@ -51,24 +52,351 @@ var _ = Describe("ServiceBuilder", func() {
...
+	wsProxyContainer := func() corev1.Container {
+		return corev1.Container{
+			Name:  nameWSProxy,
+			Image: "ghcr.io/jupyter-infra/workspace-websocket-proxy:v0.1.0-rc.1",
+			Ports: []corev1.ContainerPort{{Name: nameWSProxy, ContainerPort: 8080}},
+		}
+	}
+
+	Context("when no sidecar ports are declared", func() {
+		It("exposes only the Jupyter port with a nil access strategy", func() {
```
JGuinegagne: replace `Jupyter port` > `application port`

internal/controller/state_machine.go
```diff
@@ -237,7 +237,7 @@ func (sm *StateMachine) reconcileDesiredRunningStatus(
 
 	// Ensure service exists
 	// EnsureServiceExists internally fetches the service and returns it with current status
-	service, err := sm.resourceManager.EnsureServiceExists(ctx, workspace)
+	service, err := sm.resourceManager.EnsureServiceExists(ctx, workspace, accessStrategy)
```
JGuinegagne, line 240: Add state machine UT for access-strategy controlled service changes please.

internal/controller/service_builder.go
```diff
@@ -56,43 +65,172 @@ func (sb *ServiceBuilder) buildObjectMeta(workspace *workspacev1alpha1.Workspace
 }
 
 // buildServiceSpec creates the service specification
-func (sb *ServiceBuilder) buildServiceSpec(workspace *workspacev1alpha1.Workspace) corev1.ServiceSpec {
+func (sb *ServiceBuilder) buildServiceSpec(
+	workspace *workspacev1alpha1.Workspace,
+	accessStrategy *workspacev1alpha1.WorkspaceAccessStrategy,
+) corev1.ServiceSpec {
+	sidecarPorts := sidecarServicePorts(accessStrategy)
+	ports := make([]corev1.ServicePort, 0, 1+len(sidecarPorts))
```
JGuinegagne, line 81: Implicitly, all sidecars that i/ the access-strategy defines and ii/ that expose ports, get _all_ its port exposed in the workspace service.

1. is this a reasonable assumption that _all_ ports of _all_ sidecars should be exposed?
2. should we use the same service and why?
3. have we considered new fields in `AccessStrategy.spec`, and why did we decide against it?

JGuinegagne (changes requested):
Let's talk about the mechanism of exposing sidecar ports.

Also add:
- state machine unit tests
- E2E tests

internal/controller/service_builder.go
earaghbidikashani: will do

internal/controller/service_builder.go
earaghbidikashani: will do

internal/controller/service_builder.go
earaghbidikashani: Will fix for every instance

internal/controller/service_builder_test.go
earaghbidikashani: will do

internal/controller/state_machine.go
earaghbidikashani: will do

internal/controller/service_builder.go
earaghbidikashani: > is this a reasonable assumption that all ports of all sidecars should be exposed?

No. `containerPort` is informational in Kubernetes + a container listing a debug or admin port isn't asking for cluster-wide reachability. Deriving from `additionalContainers` conflates the two.

> should we use the same service and why?

I intentionally chose to use the same service as I didn't see a point in creating a separate service. Also separate service will cause extra work in reconciliation logic and lifecycle management. 

> have we considered new fields in `AccessStrategy.spec`, and why did we decide against it?

fair, I think a new field in access strategy is better. Deriving from `additionalContainers` means exposure from an informational field, so you have to read every sidecar to know what's reachable. `exposedPorts` field on `podModifications` (referencing container ports by name) is definitely cleaner.

Will implement that.

test/e2e/workspace_sidecar_ports_test.go
```diff
@@ -0,0 +1,223 @@
...
+	})
+
+	// servicePortNames returns the names of every port on the workspace Service, space separated.
+	servicePortNames := func(serviceName string) (string, error) {
+		return kubectlGet("service", serviceName, workspaceNamespace,
+			"{range .spec.ports[*]}{.name}{\" \"}{end}")
+	}
+
+	// serviceNameFor waits until the workspace publishes the Service it owns and returns its name.
+	serviceNameFor := func(workspaceName string) string {
```
JGuinegagne: how about moving `servicePortNames` and `serviceNameFor` to `helpers_workspace.go`?

test/e2e/workspace_sidecar_ports_test.go
```diff
@@ -0,0 +1,223 @@
...
+				controller.ConditionTypeAvailable,
+				ConditionTrue,
+			)
+
+			serviceName := serviceNameFor(workspaceName)
+
+			By("exposing the second declared port on the live access strategy")
+			patch := fmt.Sprintf(
+				`{"spec":{"deploymentModifications":{"podModifications":{"exposedPorts":[%q,%q]}}}}`,
+				exposedPortName, unexposedPortName)
```
JGuinegagne: nit: for consistency, could you use a patch json file, see [precedent](https://github.com/jupyter-infra/jupyter-k8s/blob/main/test/e2e/static/access-strategy-updates/access-strategy-update-priorities-patch.json)

test/e2e/workspace_sidecar_ports_test.go
```diff
@@ -0,0 +1,223 @@
...
+			cmd := exec.Command("kubectl", "patch", "workspaceaccessstrategy", accessStrategyName,
```
JGuinegagne: verify if there are no common utility for this please.

api/v1alpha1/workspaceaccessstrategy_types.go
```diff
@@ -51,6 +51,18 @@ type PodModifications struct {
...
 
+	// ExposedPorts lists the container port names to publish on the workspace Service.
+	// Each entry must match the name of a port declared by exactly one additional container.
+	// Ports not listed here stay reachable inside the pod but are not routed by the Service.
+	// A containerPort is informational and does not by itself grant reachability, so exposure
+	// is opt-in: this makes what the Service publishes explicit rather than inferred.
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=15
```
JGuinegagne, line 58: optional: consider starting with a lower number (say 5)

test/e2e/workspace_sidecar_ports_test.go
earaghbidikashani: Will do

test/e2e/workspace_sidecar_ports_test.go
earaghbidikashani: Will do

test/e2e/workspace_sidecar_ports_test.go
earaghbidikashani: there is in `helpers_access_strategy.go` will use

api/v1alpha1/workspaceaccessstrategy_types.go
earaghbidikashani: this is k8 valid port name limit: https://github.com/kubernetes/apimachinery/blob/release-1.34/pkg/util/validation/validation.go#L351-L369 (according to claude) 
Keeping at 15 if that's ok, lmk if that is a blocker

JGuinegagne: Generally, I'm more in favor of tightening till we get asked to loosen, but that's okay.

JGuinegagne (dismissed):

internal/controller/service_builder.go
```diff
@@ -56,43 +69,170 @@ func (sb *ServiceBuilder) buildObjectMeta(workspace *workspacev1alpha1.Workspace
 }
 
 // buildServiceSpec creates the service specification
-func (sb *ServiceBuilder) buildServiceSpec(workspace *workspacev1alpha1.Workspace) corev1.ServiceSpec {
+func buildServiceSpec(
```
henrywa2, line 72: I think as part of direct SSH implementation, we made a new service. I don't recall the reason why, but can you check and see if it's applicable here, as websocket would also be a new access pattern

api/v1alpha1/workspaceaccessstrategy_types.go
```diff
@@ -51,6 +51,18 @@ type PodModifications struct {
 	// +optional
 	AdditionalContainers []corev1.Container `json:"additionalContainers,omitempty"`
 
+	// ExposedPorts lists the container port names to publish on the workspace Service.
+	// Each entry must match the name of a port declared by exactly one additional container.
+	// Ports not listed here stay reachable inside the pod but are not routed by the Service.
+	// A containerPort is informational and does not by itself grant reachability, so exposure
+	// is opt-in: this makes what the Service publishes explicit rather than inferred.
```
henrywa2: nit: does it make sense to describe what a containerPort is in this description? We don't define other concepts like Service or container

internal/controller/service_builder.go
```diff
@@ -56,43 +69,170 @@ func (sb *ServiceBuilder) buildObjectMeta(workspace *workspacev1alpha1.Workspace
...
+		return nil, nil
+	}
+
+	podMods := accessStrategy.Spec.DeploymentModifications.PodModifications
+	if len(podMods.ExposedPorts) == 0 {
+		return nil, nil
+	}
+
+	ports := make([]corev1.ServicePort, 0, len(podMods.ExposedPorts))
+	usedNumbers := map[int32]string{JupyterPort: httpScheme}
```
henrywa2: Do you need to add JupyterPort to the map? In line 131 anything that uses JupyterPort will already throw an error

internal/controller/service_builder.go
earaghbidikashani: the SSH separate-Service for DNS/ExternalDNS-specific and doesn't transfer. Keeping ws on the shared Service is intended

api/v1alpha1/workspaceaccessstrategy_types.go
earaghbidikashani: will trim

internal/controller/service_builder.go
earaghbidikashani: will drop
that line already covers that case

henrywa2 (approved):

JGuinegagne (approved):

PR #460, opened 2026-08-14
fix: authenticate bearer tokens scoped to a workspace sub-path

earaghbidikashani:
The bearer-auth handler normalizes the request path through `ExtractAppPath` but compares it against the token's raw `path` claim. The comparison is asymmetric, so a token minted for any sub-path of a workspace is rejected with `403 Token path mismatch` 

  This surfaced building WebSocket remote access (jupyter-k8s-aws#85): the proxy needs its own route under `/workspaces/<ns>/<name>/ssh-ws`, and the bootstrap token minted for that path fails the check because `ExtractAppPath` collapses the request path to `/workspaces/<ns>/<name>` while the claim keeps the sub-path.

  ## Fix

  Normalize the token's claim through the same `ExtractAppPath` before comparing, so both sides are compared at workspace-app-path granularity:

  - No behavior change for the web UI. Its bootstrap token claim already equals the app path, so ExtractAppPath of it is itself and the comparison is unchanged.
  - No cross-workspace weakening. A token for a different workspace still normalizes to a different app path and is rejected (existing PathMismatch test still passes). ExtractWorkspaceInfo independently validates the namespace.
  - The token stays narrowly scoped to its sub-path; the middleware just authorizes at workspace granularity, which is the granularity that matters.

 ## Testing

  - New TestHandleBearerAuth_BearerTokenReview_SubPathClaim: a token claiming /workspaces/default/myworkspace/ssh-ws authenticates on that request. Verified it fails (403) against the pre-fix code and passes with it.
  - Existing cross-workspace PathMismatch and Success tests unchanged and passing.
  - make build test lint passed


internal/authmiddleware/serverroute_bearer_auth.go
```diff
@@ -93,8 +93,10 @@ func (s *Server) handleBearerAuth(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if reviewStatus.Path != appPath {
-		s.logger.Error("Token path mismatch", "token_path", reviewStatus.Path, "request_path", appPath)
+	tokenAppPath := ExtractAppPath(reviewStatus.Path, s.config.PathRegexPattern)
```
andrii-i, line 96: The docs page for this endpoint (`docs/source/dive-deeper/authmiddleware/routes.md:33`) still says step 5 verifies that the token's path matches the request path. After this change both sides are cut down to `/workspaces/<ns>/<name>` before comparing. We should update this to keep the documentation current, fine to track and do in a follow-up if you'd prefer that.

internal/authmiddleware/serverroute_bearer_auth_test.go
```diff
@@ -297,6 +297,59 @@ func TestHandleBearerAuth_BearerTokenReview_PathMismatch(t *testing.T) {
...
 }
 
+func TestHandleBearerAuth_BearerTokenReview_SubPathClaim(t *testing.T) {
+	mockServer := NewMockK8sServer(t)
+	defer mockServer.Close()
+
+	response := CreateBearerTokenReviewResponse(
+		TestDefaultNamespace,
+		true,
+		"/workspaces/default/myworkspace/ssh-ws", // token scoped to a sub-path of the workspace
```
andrii-i, line 307: This covers the accept case, please add a test for a reject case as well, fine to track and do in a follow-up if you'd prefer that.

andrii-i (approved):
Thanks @earaghbidikashani, this looks good: the request side already uses `ExtractAppPath` so it makes sense to apply it to the token's claim as well; sub-paths being cut are not used in the assertions anywhere.

Note that the PR description currently says that `ExtractWorkspaceInfo` independently validates the namespace while it does not: `ExtractWorkspaceInfo` only parses the namespace and name out of the request path, and the `BearerTokenReview` only checks the token signature and token type. Other than correctness, I think this matters because the path comparison this PR changes is essentially the only mechanism that ensures workspaces only receive requests with tokens minted for them. There is no second check, which makes the change more impactful.

henrywa2:
when do tokens get minted for sub path?

henrywa2 (approved):

internal/authmiddleware/serverroute_bearer_auth.go
earaghbidikashani: follow up: #463

internal/authmiddleware/serverroute_bearer_auth_test.go
earaghbidikashani: follow up: #463

earaghbidikashani:
> when do tokens get minted for sub path?

The Extension API mints them for WebSocket/remote-IDE connections: `generateWebSocketConnectionURL` derives the token's path claim from the access strategy's `bearerAuthURLTemplate`. When that template points at a sub-path the claim becomes `/workspaces/<ns>/<name>/ssh-ws`. The web-UI flow uses a base-path template, so it never produced a sub-path claim.

PR #463, opened 2026-08-21
feat: bearer-auth docs/tests + decouple websocket URL template

earaghbidikashani:
Follow-up to #460, addressing the two non-blocking review comments.

## 1. Docs + test follow-ups (from #460 review)
  - **Docs:** `authmiddleware/routes.md` step 5 for `/bearer-auth` said the token path must match the request path. Since #460 both sides are normalized via `ExtractAppPath` to `/workspaces/<ns>/<name>` before comparing — updated to reflect that (sub-path token authenticates for its workspace; different workspace rejected).
  - **Test:** added a reject-case test — a sub-path token scoped to a *different* workspace returns 403, guarding the cross-workspace boundary alongside the existing accept case.

  ## 2. Decouple the WebSocket connection URL from `bearerAuthURLTemplate`
  The extension API derived the WebSocket URL from `bearerAuthURLTemplate` (stripping `/bearer-auth`), so the same field had to serve both the browser bootstrap path and the WebSocket endpoint. Once the WebSocket proxy moved to a distinct sub-path (`/ssh-ws`), no single value could satisfy both — the browser bootstrap would dead-end at the proxy, or the WebSocket URL would point where no route matches. (Surfaced in jupyter-k8s-aws#85 review.)

  - New optional `webSocketURLTemplate` field on `WorkspaceAccessStrategy`.
  - `generateWebSocketConnectionURL` prefers it (used as-is, scheme → `wss://`); falls back to the previous behavior (`bearerAuthURLTemplate` minus `/bearer-auth`) when unset, so existing strategies are unchanged.
  - **Auth is unchanged** — same bootstrap JWT, same `authmiddleware-bearer-auth` handler, same path normalization. Only the endpoint URL is decoupled.

  ## Testing
  - New unit test asserting `webSocketURLTemplate` takes precedence and is used verbatim (no `/bearer-auth` stripping); existing WebSocket URL tests still pass.
  - `make build test lint verify-generated` all pass. CRD, chart CRD, `install.yaml`, and reference docs regenerated for the new field.


api/v1alpha1/workspaceaccessstrategy_types.go
```diff
@@ -158,6 +158,10 @@ type WorkspaceAccessStrategySpec struct {
 	// +optional
 	BearerAuthURLTemplate string `json:"bearerAuthURLTemplate,omitempty"`
 
+	// WebSocketURLTemplate constructs the WebSocket (ssh-over-websocket) connection URL; if unset, it falls back to BearerAuthURLTemplate with /bearer-auth stripped. Template vars: .Workspace, .AccessStrategy.
```
henrywa2: why do we have to maintain backwards compatibility here?

earaghbidikashani: apologies this was related to my testing I did earlier which is irrelevant. Will remove

henrywa2:
image
👀

PR #464, opened 2026-08-21
feat: bearer-auth docs/tests + decouple websocket URL template

earaghbidikashani:
Follow-up to #460, addressing the two non-blocking review comments.

## 1. Docs + test follow-ups (from #460 review)
  - **Docs:** `authmiddleware/routes.md` step 5 for `/bearer-auth` said the token path must match the request path. Since #460 both sides are normalized via `ExtractAppPath` to `/workspaces/<ns>/<name>` before comparing — updated to reflect that (sub-path token authenticates for its workspace; different workspace rejected).
  - **Test:** added a reject-case test — a sub-path token scoped to a *different* workspace returns 403, guarding the cross-workspace boundary alongside the existing accept case.

  ## 2. Decouple the WebSocket connection URL from `bearerAuthURLTemplate`
  The extension API derived the WebSocket URL from `bearerAuthURLTemplate` (stripping `/bearer-auth`), so the same field had to serve both the browser bootstrap path and the WebSocket endpoint. Once the WebSocket proxy moved to a distinct sub-path (`/ssh-ws`), no single value could satisfy both — the browser bootstrap would dead-end at the proxy, or the WebSocket URL would point where no route matches. (Surfaced in jupyter-k8s-aws#85 review.)

  - New optional `webSocketURLTemplate` field on `WorkspaceAccessStrategy`.
  - `generateWebSocketConnectionURL` prefers it (used as-is, scheme → `wss://`); falls back to the previous behavior (`bearerAuthURLTemplate` minus `/bearer-auth`) when unset, so existing strategies are unchanged.
  - **Auth is unchanged** — same bootstrap JWT, same `authmiddleware-bearer-auth` handler, same path normalization. Only the endpoint URL is decoupled.

  ## Testing
  - New unit test asserting `webSocketURLTemplate` takes precedence and is used verbatim (no `/bearer-auth` stripping); existing WebSocket URL tests still pass.
  - `make build test lint verify-generated` all pass. CRD, chart CRD, `install.yaml`, and reference docs regenerated for the new field.


henrywa2 (approved):

api/v1alpha1/workspaceaccessstrategy_types.go
```diff
@@ -158,6 +158,10 @@ type WorkspaceAccessStrategySpec struct {
 	// +optional
 	BearerAuthURLTemplate string `json:"bearerAuthURLTemplate,omitempty"`
 
+	// WebSocketURLTemplate constructs the WebSocket (ssh-over-websocket) connection URL; required to enable WebSocket connections. Template vars: .Workspace, .AccessStrategy.
```
andrii-i, line 161: `generateWebSocketConnectionURL` replaces the rendered URL's scheme with `wss`, so a template author or admin writing `ws://` for a non-TLS dev setup would ideally learn that here rather than from the connection failing. An explanation of this behavior should be in the docstring.

internal/extensionapi/serverroute_connection.go
```diff
@@ -181,44 +181,23 @@ func (s *ExtensionServer) generateWebSocketConnectionURL(r *http.Request, ws *wo
 		return "", fmt.Errorf("failed to create signer: %w", err)
 	}
 
-	// Generate URL from template (reuses the same BearerAuthURLTemplate)
-	wsURL, err := s.renderBearerAuthURL(accessStrategy.Spec.BearerAuthURLTemplate, ws, accessStrategy)
+	renderedURL, err := s.renderBearerAuthURL(accessStrategy.Spec.WebSocketURLTemplate, ws, accessStrategy)
```
andrii-i, line 184: `renderBearerAuthURL` now renders `webSocketURLTemplate` as well; should it be renamed to something like `renderAccessURLTemplate` to match? It's package-private with both call sites in this file, so the rename has no API impact.

andrii-i (approved):
Thanks @earaghbidikashani, looks good and I think a template field per client-facing entry point is a good architecture.

Note that the PR description says `generateWebSocketConnectionURL` "falls back to the previous behavior (`bearerAuthURLTemplate` minus `/bearer-auth`) when unset, so existing strategies are unchanged", while this doesn't happen. At the same time I think the actual logic is good, up to you if you want to update the PR description, nonblocking.

Also please find a couple of non-blocking comments worth noting below.

PR #465, opened 2026-08-24
feat: add emptyDir workspace volumes

### LINKED ISSUE #466: Support `emptyDir-backed` workspace volumes
## Summary

- Workspace users cannot configure native Kubernetes `emptyDir` volumes through the Workspace CRD. This blocks pod-scoped mounts such as memory-backed `/dev/shm`, which are needed by workloads using POSIX shared memory.
- Today `Workspace.spec.volumes[]` supports additional PVC-backed mounts only. This proposal extends the existing volume API so each volume can be backed by either `persistentVolumeClaimName` or `emptyDir`.

  ## Evidence
  Kubernetes supports memory-backed shared-memory mounts with:
  ```yaml
  emptyDir:
    medium: Memory
    sizeLimit: 1Gi

This is the standard Kubernetes pattern for pod-lifetime tmpfs storage. It is useful for /dev/shm consumers such as Python multiprocessing, PyTorch dataloaders, Chromium-based tooling, and other workloads that need larger shared memory than the container runtime default. The current Workspace API cannot express this through spec.volumes[].

## Impact
  - Users cannot configure `/dev/shm` through the Workspace CRD.
  - Workloads that need pod-lifetime shared memory must rely on image/runtime defaults or platform-specific workarounds.
  - Admins cannot provide a default memory-backed `/dev/shm` mount through `WorkspaceTemplate.spec.defaultVolumes[]`.

  ## Recommended fix

  Extend VolumeSpec with an emptyDir source while preserving existing PVC behavior:

```
type VolumeSpec struct {
       Name      string `json:"name"`
       MountPath string `json:"mountPath"`

       PersistentVolumeClaimName string                       `json:"persistentVolumeClaimName,omitempty"`
       EmptyDir                  *corev1.EmptyDirVolumeSource `json:"emptyDir,omitempty"`
  }
```
Add CRD validation requiring exactly one volume source:

```
x-kubernetes-validations:
    - message: exactly one of persistentVolumeClaimName or emptyDir must be specified
      rule: (has(self.persistentVolumeClaimName) && !has(self.emptyDir)) ||
            (!has(self.persistentVolumeClaimName) && has(self.emptyDir))
```

Render PVC-backed entries as PersistentVolumeClaimVolumeSource and emptyDir entries as EmptyDirVolumeSource.

Template behavior should remain unchanged in this change:
- WorkspaceTemplate.spec.defaultVolumes[] can include emptyDir because it reuses VolumeSpec.
- defaultVolumes remains all-or-nothing and applies only when the workspace does not specify any spec.volumes[].
- allowSecondaryStorages: false continues to reject all secondary volumes, whether PVC-backed or emptyDir-backed.

Future template policy can add source-specific controls such as allowed volume types, required sizeLimit, mount path rules, or emptyDir.medium restrictions.

aakashmandavilli96:
## Summary

  Adds support for `emptyDir`-backed volumes in `Workspace.spec.volumes[]`, enabling memory-backed mounts such as `/dev/shm` while preserving the existing PVC-backed
  volume behavior.

  This keeps the change narrow:
  - `spec.storage` remains unchanged as the operator-managed persistent home volume.
  - `spec.volumes[]` now supports either `persistentVolumeClaimName` or `emptyDir`.
  - CRD validation enforces exactly one backing source per volume, matching Kubernetes volume semantics.
  - The controller renders either a PVC volume source or an `emptyDir` volume source.
  - PVC ownership validation skips non-PVC volumes.

  ## Testing

  - `go test ./api/v1alpha1 ./internal/controller ./internal/webhook/v1alpha1`
  - `make test`
  - `make build`
  - `make build-e2e`
  - `make verify-generated`
  - `make helm-lint`
  - `make helm-test`

  Full Kind e2e was not run locally because this environment does not have host Go available for the full e2e Makefile target.


api/v1alpha1/workspace_types.go
```diff
@@ -14,16 +14,22 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-// VolumeSpec defines a volume to mount from an existing PVC
+// VolumeSpec defines a volume to mount from an existing PVC or emptyDir source.
+// +kubebuilder:validation:XValidation:rule="(has(self.persistentVolumeClaimName) && !has(self.emptyDir)) || (!has(self.persistentVolumeClaimName) && has(self.emptyDir))",message="exactly one of persistentVolumeClaimName or emptyDir must be specified"
```
earaghbidikashani, line 18: `self.persistentVolumeClaimName` is true for an empty string, so a volume with persistentVolumeClaimName: "" and no emptyDir passes. 

consider adding minLength 1 on persistentVolumeClaimName.

api/v1alpha1/workspace_types.go
```diff
@@ -14,16 +14,22 @@ import (
...
 	// MountPath is the path where the volume should be mounted (Unix-style path, e.g. /data)
 	MountPath string `json:"mountPath"`
+
+	// PersistentVolumeClaimName is the name of the existing PVC to mount
+	// +optional
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName,omitempty"`
+
+	// EmptyDir configures a pod-scoped ephemeral volume
+	// +optional
+	EmptyDir *corev1.EmptyDirVolumeSource `json:"emptyDir,omitempty"`
```
earaghbidikashani, line 33: No bound on `EmptyDir`
 
Consider adding a template level max sizeLimit and require sizeLimit when medium is Memory, matching `StorageConfig` and `ResourceBounds` (not sure if this absolutely necessary, please take a look for emptyDir limiting conventions)

api/v1alpha1/workspacetemplate_types.go
```diff
@@ -77,8 +77,6 @@ type WorkspaceTemplateSpec struct {
 	AllowSecondaryStorages *bool `json:"allowSecondaryStorages,omitempty"`
 
 	// DefaultVolumes specifies default additional volumes for workspaces using this template
-	// Volumes are applied during defaulting only if the workspace does not specify any volumes
-	// Each volume references a pre-existing PVC by name in the workspace's namespace
```
earaghbidikashani, line 81: why remove these?

internal/controller/deployment_builder.go
```diff
@@ -195,13 +195,17 @@ func (db *DeploymentBuilder) buildPodSpec(workspace *workspacev1alpha1.Workspace
 			// Skip if name conflicts with primary storage
 			continue
 		}
-		podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
-			Name: vol.Name,
-			VolumeSource: corev1.VolumeSource{
-				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-					ClaimName: vol.PersistentVolumeClaimName,
-				},
+		volumeSource := corev1.VolumeSource{
```
earaghbidikashani, line 198: User `emptyDir` volumes can collide with volumes injected by the access strategy overlay (the HyperPod shm case injects an emptyDir at /dev/shm). 

The integration overlay checks name collisions and errors, but nothing checks mountPath. A user shm volume at `/dev/shm` then produces a duplicate name or mountPath and the deployment update is rejected, leaving the workspace stuck. 

We should add collision handling or reserve the shm path.

internal/webhook/v1alpha1/volume_validator.go
```diff
@@ -60,6 +60,10 @@ func validateSecondaryStorages(volumes []workspacev1alpha1.VolumeSpec, template
 // validateVolumeOwnership checks that volumes don't reference PVCs owned by other workspaces
 func validateVolumeOwnership(ctx context.Context, k8sClient client.Client, workspace *workspacev1alpha1.Workspace) *TemplateViolation {
```
earaghbidikashani, line 61: `emptyDir` is still blocked by `allowSecondaryStorages` so shm cannot be allowed when PVC mounts are blocked. 

Consider decoupling

earaghbidikashani:
Thank you @aakashmandavilli96 overall LGTM, couple of general comments: 

1. Nothing bounds emptyDir sizeLimit anywhere. primaryStorage bounds only spec.storage and resourceBounds only spec.resources, so a memory backed emptyDir is an unbounded node memory vector
2. `volume_defaulter.go` still uses a shallow copy(). Now that VolumeSpec holds `EmptyDir *corev1.EmptyDirVolumeSource`, and `EmptyDirVolumeSource` holds `SizeLimit *resource.Quantity`, the workspace and the resolved template share the same EmptyDir pointer. Every other defaulter uses DeepCopy. Switch to per element DeepCopyInto.
3. Add unit coverage for a PVC and emptyDir volume together.
4. Add negative e2e or envtest cases: both persistentVolumeClaimName and emptyDir set, and neither set

api/v1alpha1/workspacetemplate_types.go
aakashmandavilli96: That was unintentional cleanup noise. I restored the comments and updated them to reflect both supported sources areexisting `PVC` and `emptyDir`.

api/v1alpha1/workspace_types.go
aakashmandavilli96: I added `+kubebuilder:validation:MinLength=1` for `persistentVolumeClaimName` and regenerated the `CRD/chart/reference ` docs so `persistentVolumeClaimName: ""` is rejected.

api/v1alpha1/workspace_types.go
aakashmandavilli96: I looked into this. `emptyDir.sizeLimit` is optional in Kubernetes, including for `medium: Memory`, so I don’t think the base `Workspace` CRD should require it or hard-code a max in this PR.

This PR keeps the `emptyDir` source aligned with the Kubernetes API shape. Platform-specific bounds, such as requiring `sizeLimit` for memory-backed volumes or enforcing a max size, should follow the existing template/admission policy model. 
Admins can already set bounded defaults through  `WorkspaceTemplate.spec.defaultVolumes[]`;

api/v1alpha1/workspace_types.go
aakashmandavilli96: I added `MinLength=1` for `persistentVolumeClaimName` and updated the generated CRD/chart/reference docs so an empty PVC name is rejected

api/v1alpha1/workspace_types.go
aakashmandavilli96: I looked into this. `emptyDir.sizeLimit` is optional in Kubernetes, including for `medium: Memory`, so I don’t think the base `Workspace` CRD should require it or hardcode a max in this PR.

This PR keeps the `emptyDir` source aligned with the Kubernetes API shape. Platform-specific bounds such as requiring `sizeLimit` for memory-backed volumes or enforcing a max size should follow the existing template/admission policy model as a follow-up.

api/v1alpha1/workspacetemplate_types.go
aakashmandavilli96: That was unintentional cleanup noise. I restored the comments and updated them to reflect both supported sources `existing PVC` and `emptyDir`.

internal/controller/deployment_builder.go
aakashmandavilli96: I agree duplicate mount paths can be an issue if an access strategy also injects `/dev/shm`. I don’t think reserving `/dev/shm` in the base `Workspace` API is the right fix because `/dev/shm` is the main customer use case for user-managed `emptyDir`.

If we want collision handling, I think it belongs in the overlay `merge/validation path` as a separate follow-up because it affects all access-strategy injected mounts, not just this `emptyDir` source addition.

internal/webhook/v1alpha1/volume_validator.go
aakashmandavilli96: This PR preserves the existing meaning of `allowSecondaryStorages` it gates `spec.volumes[]` as a whole. Decoupling by source type would be a new template policy API and I’d prefer to handle that separately instead of changing template semantics in this PR

aakashmandavilli96:
> Thank you @aakashmandavilli96 overall LGTM, couple of general comments:
> 
>     1. Nothing bounds emptyDir sizeLimit anywhere. primaryStorage bounds only spec.storage and resourceBounds only spec.resources, so a memory backed emptyDir is an unbounded node memory vector
> 
>     2. `volume_defaulter.go` still uses a shallow copy(). Now that VolumeSpec holds `EmptyDir *corev1.EmptyDirVolumeSource`, and `EmptyDirVolumeSource` holds `SizeLimit *resource.Quantity`, the workspace and the resolved template share the same EmptyDir pointer. Every other defaulter uses DeepCopy. Switch to per element DeepCopyInto.
> 
>     3. Add unit coverage for a PVC and emptyDir volume together.
> 
>     4. Add negative e2e or envtest cases: both persistentVolumeClaimName and emptyDir set, and neither set

Fixed these in the new commit

api/v1alpha1/workspace_types.go
JGuinegagne: Note that these validations may become more complex if we add other types of volumes. This would call for a `volumeType` enum field responsible for the switch, and the `emptyDir` could have a simpler validation that `volumeType==emptyDir`.

This is somewhat of a one-way door: adding `volumeType` later would break the pre-existing `emptyDir` behavior if we merge this CRD change as is.

However:
1. reading the [documentation](https://kubernetes.io/docs/concepts/storage/volumes/), there doesn't seem to be any other likely volume that wouldn't use a PV; perhaps `local` but I'm not sure?
2. given `pod.spec.volumes` itself doesn't rely on such an enum, we are following the upstream model

Based on 1/ & 2/, you are likely making the right call here, but just flagging it for your consideration re extensibility. Asked another way: if you get asked to support say `local` type, how would you extend the CRD? As long as you have a reasonable plan we can merge the PR.

api/v1alpha1/workspace_types.go
JGuinegagne: We may need some form of limits Sri; the [documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#memory-backed-emptydir) flags that `emptyDir` can grow to cause OOM errors. However, this would require more thoughts.

Could you track as a separate GH issue and reference it here?

internal/controller/deployment_builder.go
JGuinegagne: sanity-check: does the CRD validates that `ws.spec.volumes.mountPath` are distinct?

if not, could you cut an issue to track it please?

It gets more complex w/ [access-strategy-injected volumes](https://github.com/jupyter-infra/jupyter-k8s/blob/6664d49ff23ef760573d83d83f03cee037a0135b/api/v1alpha1/workspaceaccessstrategy_types.go#L64) and [pod-modifications](https://github.com/jupyter-infra/jupyter-k8s/blob/6664d49ff23ef760573d83d83f03cee037a0135b/api/v1alpha1/workspaceaccessstrategy_types.go#L79). Ideally we would have 2 levels of validations there:
1. within access-strategy resource, via `AccessStrategy` CRD validations
2. at workspace admission time when using an access strategy. Problem: access-strategy may change while a workspace is using it, the edge cases are hard to handle.

internal/webhook/v1alpha1/volume_validator.go
```diff
@@ -60,6 +60,10 @@ func validateSecondaryStorages(volumes []workspacev1alpha1.VolumeSpec, template
...
 	for _, volume := range workspace.Spec.Volumes {
+		if volume.PersistentVolumeClaimName == "" {
+			continue
```
JGuinegagne, line 65: add an inline comment why we `continue` here please.

test/e2e/static/storage-node/workspace-with-emptydir.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: workspace.jupyter.org/v1alpha1
```
JGuinegagne, line 1: non-blocking: technically an `emptyDir` isn't a `storage-external`...

Could you move this `.yaml` to a new `<>/static/storage-node/` dir ?

test/e2e/workspace_storage_test.go
```diff
@@ -293,6 +293,43 @@ var _ = Describe("Workspace Storage", Ordered, func() {
 			VerifyPodCanAccessExternalVolumes(workspaceName, workspaceNamespace, externalPvc2Name, "/home/jovyan/shared")
 		})
 
+		It("should mount memory-backed emptyDir volumes specified in volumes field", func() {
```
JGuinegagne, line 350: move to its own `Context` please.

Also, how about:
1. adding a test case where we add both a PV/PVC _and_ an `emptyDir` ?
2. adding a test case where two workspaces attempt to use `emptyDir` (which `kind` will place on the same node) ?

JGuinegagne (commented):
Few non-blocking comments

api/v1alpha1/workspace_types.go
```diff
@@ -14,16 +14,23 @@ import (
...
 	MountPath string `json:"mountPath"`
+
+	// PersistentVolumeClaimName is the name of the existing PVC to mount
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName,omitempty"`
+
+	// EmptyDir configures a pod-scoped ephemeral volume
+	// +optional
+	EmptyDir *corev1.EmptyDirVolumeSource `json:"emptyDir,omitempty"`
```
andrii-i, line 33: Note that currently, due to the kubelet's size selection logic for memory-backed emptyDir volumes (https://github.com/kubernetes/kubernetes/blob/v1.33.1/pkg/volume/emptydir/empty_dir.go#L114-L145), when a workspace's memory limit is below `sizeLimit`, the user silently gets a smaller `/dev/shm` than they asked for. Then if the container's memory limit is exhausted, the kernel kills the container instead of returning a disk-full error. Such conflicts should fail loudly at admission, the same way the webhook already rejects resource requests outside template bounds:

https://github.com/jupyter-infra/jupyter-k8s/blob/6664d49ff23ef760573d83d83f03cee037a0135b/internal/webhook/v1alpha1/resource_validator.go#L68-L92

Fine to address now or in a follow-up.

andrii-i (dismissed):
Thanks @aakashmandavilli96, looks good. Please find one non-blocking comment below; I also agree with Jonathan's and Eilia's comments on mountPath / injected-volume collision handling.

Also I already did this but it's good practice to link a PR to the issue it addresses / will close; a couple of ways to do that on GitHub: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

earaghbidikashani (dismissed):
LGTM

api/v1alpha1/workspace_types.go
aakashmandavilli96: - Thanks, that makes sense. My plan would be to keep the Kubernetes-style union here rather than introduce a separate `volumeType` discriminator now. `persistentVolumeClaimName` and `emptyDir` act as the source discriminators, matching the upstream `pod.spec.volumes` model.

- If we add another pod-scoped source later, we can add another optional source field and update the CEL rule to require exactly one supported source. Existing `emptyDir` specs would remain valid. For `local`, my understanding is that Kubernetes exposes local storage through a PV/PVC, so the existing `persistentVolumeClaimName` path should already cover that case

api/v1alpha1/workspace_types.go
aakashmandavilli96: Tracking issue for limits https://github.com/jupyter-infra/jupyter-k8s/issues/467

internal/controller/deployment_builder.go
aakashmandavilli96: - The current CRD validates the reserved volume name and the PVC/emptyDir union, but it does not validate uniqueness of `spec.volumes[].mountPath`.
- I opened a follow-up to track mountPath collision validation here: https://github.com/jupyter-infra/jupyter-k8s/issues/468
- I kept it separate because the full solution needs to consider workspace-defined volumes, access-strategy injected volume mounts, and pod modifications.

internal/webhook/v1alpha1/volume_validator.go
aakashmandavilli96: Will fix this in a revision

test/e2e/static/storage-node/workspace-with-emptydir.yaml
aakashmandavilli96: Agree moving this to a different path in the next revison

test/e2e/workspace_storage_test.go
aakashmandavilli96: Added and moved to its own context

earaghbidikashani (approved):

andrii-i (approved):
Thanks for creating issues to track follow-ups.

JGuinegagne (approved):

PR #473, opened 2026-09-03
test: gpu and lifecycle e2e coverage

andrii-i:
GPU contract e2e that runs on plain Kind, plus lifecycle coverage for data persistence and out-of-band drift. Scheduling and resource propagation are API-server and scheduler behavior, so a node that merely advertises fake GPU capacity exercises them without hardware; only CUDA itself needs a real GPU, and that lives in jupyter-deploy's EKS e2e.

### Implementation

1. `test/e2e/helpers_gpu.go`: `setupFakeGPUNode` labels the cluster's first node and patches `nvidia.com/gpu` into its status capacity and allocatable together. Patching both matters: the kubelet only recomputes allocatable from capacity on its periodic status update, so patching capacity alone leaves the two inconsistent for the scheduler. Teardown restores the node.
2. `test/e2e/workspace_gpu_test.go` pins the GPU contracts against a bounded GPU template fixture: template GPU defaults and placement (nodeSelector, toleration) propagate to the pod and it schedules onto the advertising node; a workspace GPU request within template bounds is honored; a request above the bound is rejected by validation and the error names the GPU resource; a CPU-only resource override gains no injected GPU defaults; and a workspace without a template stays free of GPU resources and placement.
3. `test/e2e/workspace_storage_test.go` adds the stop/start data-persistence spec: write a file, stop (pod deleted, PVC retained and Bound), start on a fresh pod, and the file is still there.
4. `test/e2e/workspace_drift_test.go`: deleting the workspace's Service out of band gets it recreated as a new object, and an out-of-band mutation of the deployment's pod template is reverted, both driven by the controller's watch on the resources it owns.
5. Fixtures under `test/e2e/static/gpu/` and `test/e2e/static/drift/`.

### Testing

Adds eight e2e specs: the five GPU contract specs, the stop/start persistence spec, and the two drift-repair specs. They compile into the existing Kind e2e lane and run on every PR; no workflow changes.

### DIFF
--- a/test/e2e/helpers_gpu.go
+++ b/test/e2e/helpers_gpu.go
@@ -1,0 +1,152 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright (c) Amazon Web Services
+Distributed under the terms of the MIT license
+*/
+
+package e2e
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/jupyter-infra/jupyter-k8s/internal/controller"
+	"github.com/jupyter-infra/jupyter-k8s/test/utils"
+)
+
+// Helpers for the Workspace GPU e2e suite (workspace_gpu_test.go). Kind nodes have no GPUs, so the
+// suite advertises nvidia.com/gpu on a node through the status subresource — the documented way to
+// advertise an extended resource without a device plugin. The scheduler fits pods against the
+// patched allocatable and the kubelet runs them; only CUDA itself would need real hardware.
+
+const (
+	// fakeGPUResourceName is the extended resource advertised on the fake GPU node.
+	fakeGPUResourceName = "nvidia.com/gpu"
+	// fakeGPUNodeLabel marks the advertising node; the GPU template's defaultNodeSelector targets it.
+	fakeGPUNodeLabel = "e2e.jupyter.org/fake-gpu"
+	// fakeGPUCapacity leaves headroom over the largest single request (2) so a pod still
+	// terminating from the previous spec cannot make the next one unschedulable.
+	fakeGPUCapacity = "4"
+)
+
+// fakeGPUResource is fakeGPUResourceName as a typed key into ResourceList maps.
+var fakeGPUResource = corev1.ResourceName(fakeGPUResourceName)
+
+// setupFakeGPUNode advertises fake GPU capacity on the cluster's first node and labels it for
+// nodeSelector-based placement. Returns the node name. Capacity and allocatable are patched
+// together: the kubelet only recomputes allocatable from capacity on its periodic status sync,
+// too slow for a test to wait on.
+func setupFakeGPUNode() string {
+	ginkgo.GinkgoHelper()
+
+	nodeName, err := kubectlGet("nodes", "", "", "{.items[0].metadata.name}")
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	gomega.Expect(nodeName).NotTo(gomega.BeEmpty(), "expected at least one node in the cluster")
+
+	ginkgo.By(fmt.Sprintf("advertising %s=%s on node %s", fakeGPUResourceName, fakeGPUCapacity, nodeName))
+	patch := fmt.Sprintf(
+		`[{"op":"add","path":"/status/capacity/%[1]s","value":"%[2]s"},`+
+			`{"op":"add","path":"/status/allocatable/%[1]s","value":"%[2]s"}]`,
+		jsonPatchEscapedGPUResource(), fakeGPUCapacity)
+	cmd := exec.Command("kubectl", "patch", "node", nodeName,
+		"--subresource=status", "--type=json", "-p", patch)
+	_, err = utils.Run(cmd)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	ginkgo.By("verifying the node reports the fake GPU allocatable")
+	var node corev1.Node
+	gomega.Expect(kubectlGetInto("node", nodeName, "", &node)).To(gomega.Succeed())
+	gomega.Expect(node.Status.Allocatable).To(gomega.HaveKey(fakeGPUResource))
+
+	ginkgo.By(fmt.Sprintf("labeling node %s with %s=true", nodeName, fakeGPUNodeLabel))
+	cmd = exec.Command("kubectl", "label", "--overwrite", "node", nodeName, fakeGPUNodeLabel+"=true")
+	_, err = utils.Run(cmd)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	return nodeName
+}
+
+// teardownFakeGPUNode removes the fake GPU advertisement and label. Best-effort: CI deletes the
+// Kind cluster after the suite; this matters for reruns against a kept dev cluster.
+func teardownFakeGPUNode(nodeName string) {
+	ginkgo.GinkgoHelper()
+	if nodeName == "" {
+		return
+	}
+	patch := fmt.Sprintf(
+		`[{"op":"remove","path":"/status/capacity/%[1]s"},`+
+			`{"op":"remove","path":"/status/allocatable/%[1]s"}]`,
+		jsonPatchEscapedGPUResource())
+	cmd := exec.Command("kubectl", "patch", "node", nodeName,
+		"--subresource=status", "--type=json", "-p", patch)
+	_, _ = utils.Run(cmd)
+	cmd = exec.Command("kubectl", "label", "node", nodeName, fakeGPUNodeLabel+"-")
+	_, _ = utils.Run(cmd)
+}
+
+// jsonPatchEscapedGPUResource returns fakeGPUResourceName with '/' escaped as '~1' (RFC 6901).
+func jsonPatchEscapedGPUResource() string {
+	return strings.ReplaceAll(fakeGPUResourceName, "/", "~1")
+}
+
+// gpuQuantity returns the nvidia.com/gpu value in a resource list and whether the key is present.
+func gpuQuantity(list corev1.ResourceList) (int64, bool) {
+	q, ok := list[fakeGPUResource]
+	if !ok {
+		return 0, false
+	}
+	return q.Value(), true
+}
+
+// hasToleration reports whether tolerations contains an exact key/operator/effect match.
+func hasToleration(
+	tolerations []corev1.Toleration,
+	key string,
+	op corev1.TolerationOperator,
+	effect corev1.TaintEffect,
+) bool {
+	for _, t := range tolerations {
+		if t.Key == key && t.Operator == op && t.Effect == effect {
+			return true
+		}
+	}
+	return false
+}
+
+// workspacePod returns the workspace's pod, decoded.
+func workspacePod(workspaceName, namespace string) *corev1.Pod {
+	ginkgo.GinkgoHelper()
+	podName, err := kubectlGetByLabels("pod",
+		fmt.Sprintf("%s=%s", controller.LabelWorkspaceName, workspaceName),
+		namespace, "{.items[0].metadata.name}")
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	gomega.Expect(podName).NotTo(gomega.BeEmpty())
+	var pod corev1.Pod
+	gomega.Expect(kubectlGetInto("pod", podName, namespace, &pod)).To(gomega.Succeed())
+	return &pod
+}
+
+// deleteResourcesForGPUTest removes only the objects this Ordered suite creates, by explicit name,
+// so it can never nuke unrelated objects sharing the "default" namespace.
+func deleteResourcesForGPUTest(workspaceNamespace string) {
+	ginkgo.GinkgoHelper()
+
+	ginkgo.By("cleaning up GPU test workspaces")
+	cmd := exec.Command("kubectl", "delete", "workspace",
+		"gpu-default-workspace", "gpu-override-workspace", "gpu-exceed-workspace",
+		"gpu-cpu-only-workspace", "gpu-no-template-workspace",
+		"-n", workspaceNamespace, "--ignore-not-found", "--wait=true", "--timeout=120s")
+	_, _ = utils.Run(cmd)
+
+	ginkgo.By("cleaning up the GPU template")
+	cmd = exec.Command("kubectl", "delete", "workspacetemplate", "gpu-template",
+		"-n", SharedNamespace, "--ignore-not-found", "--wait=true", "--timeout=60s")
+	_, _ = utils.Run(cmd)
+}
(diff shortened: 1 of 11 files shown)


test/e2e/static/gpu/gpu-default-workspace.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: workspace.jupyter.org/v1alpha1
+kind: Workspace
+metadata:
+  name: gpu-default-workspace
+spec:
+  displayName: "GPU Defaults Test"
+  templateRef:
+    name: gpu-template
+    namespace: jupyter-k8s-shared
+  ownershipType: Public
```
JGuinegagne: sanity-check: no `nvidia.com/gpu` resource request?

test/e2e/workspace_storage_test.go
```diff
@@ -258,6 +259,83 @@ var _ = Describe("Workspace Storage", Ordered, func() {
 			By("verifying the data was persisted")
 			VerifyHomeVolumeDataPersisted(workspaceName, workspaceNamespace)
 		})
+
+		It("should retain the PVC and persist data across Stopped and back to Running", func() {
```
JGuinegagne: sanity-check: we really had no e2e tests that verified volume persistence?

non-blocking: consider writing to volume, stop, restart, verify it has persisted.

JGuinegagne (dismissed):

test/e2e/static/gpu/gpu-default-workspace.yaml
andrii-i: This is deliberate: the test covers the basic user flow where the user does not set resources and the template's defaults apply. A workspace created from a GPU template in the web UI with no explicit GPU choice arrives exactly like this, with no resources in the request. The explicit GPU request has [its own test](https://github.com/jupyter-infra/jupyter-k8s/blob/8d67f2b/test/e2e/workspace_gpu_test.go#L107). Added a comment on the fixture clarifying this.

test/e2e/workspace_storage_test.go
andrii-i: Before this PR there was one test that verified volume persistence: [`should persist data across pod restart`](https://github.com/jupyter-infra/jupyter-k8s/blob/4e0750158ffe3acf94ce8265b0129698efa9c61d/test/e2e/workspace_storage_test.go#L237) deletes the pod while the workspace stays Running and checks the data is still there. The new [`should keep the PVC and its data across Stopped and back to Running`](https://github.com/jupyter-infra/jupyter-k8s/blob/8d67f2b/test/e2e/workspace_storage_test.go#L266) covers writing to the volume, stop (Deployment and Service deleted, PVC kept and Bound), start, and verifying the file persisted.

test/e2e/workspace_gpu_test.go
```diff
@@ -0,0 +1,272 @@
...
+var _ = Describe("Workspace GPU", Ordered, func() {
+	const (
+		workspaceNamespace = "default"
+		groupDir           = "gpu"
+		gpuTemplateName    = "gpu-template"
+	)
+
+	var gpuNodeName string
+
+	BeforeAll(func() {
```
JGuinegagne: `[Efficiency]` **This suite spends ~5 of the ~6.5 minutes of timeout headroom the e2e suite had left** — see the blocking note in the review summary for the `Makefile` side. Two reductions here, both cheap:

`gpuTemplateName` is identical in all five specs that create it (`:54`, `:111`, `:141`, `:191`, `:215`), and `deleteResourcesForGPUTest` tears it down with `--wait=true` after *every* spec — six create/finalizer-delete cycles for one fixture. It already lives in a `BeforeAll`-bearing container:

```go
BeforeAll(func() {
    gpuNodeName = setupFakeGPUNode()
    createTemplateForTest(gpuTemplateName, groupDir, "")
})
```

…with the template delete moved to `AfterAll` alongside `teardownFakeGPUNode`, and dropped from `deleteResourcesForGPUTest`.

Second, `deleteResourcesForGPUTest` (`helpers_gpu.go:170`) deletes **all five** workspace names with `--wait=true --timeout=120s` after every spec, though each spec creates exactly one. Four of those five are `--ignore-not-found` no-ops, but the fifth blocks on finalizer + PVC teardown while the others are resolved serially. Taking the name as a parameter keeps the "delete only what this suite created" guarantee the comment promises, at 1/5 the wait.

test/e2e/helpers_gpu.go
```diff
@@ -0,0 +1,184 @@
...
+var fakeGPUResource = corev1.ResourceName(fakeGPUResourceName)
+
+// setupFakeGPUNode advertises fake GPU capacity on the cluster's first node and labels it for
+// nodeSelector-based placement. Returns the node name. Capacity and allocatable are patched
+// together: the kubelet only recomputes allocatable from capacity on its periodic status sync,
+// too slow for a test to wait on.
+func setupFakeGPUNode() string {
+	ginkgo.GinkgoHelper()
+
+	nodeName, err := kubectlGet("nodes", "", "", "{.items[0].metadata.name}")
```
JGuinegagne: `[Reliability]` Two sharp edges in picking `{.items[0]}` unconditionally.

`add` on an existing JSON Pointer path is an **overwrite**, not a conflict. The suite only pins the kubectl context when `KIND_CLUSTER` is set (`e2e_suite_test.go:91-100`) — every `make` path sets it, but a bare `go test -tags=e2e ./test/e2e/` does not. Run that way against a cluster with real GPU nodes, this silently rewrites a real node's advertised `nvidia.com/gpu` to `4` and pods get scheduled onto hardware that isn't there.

Separately, `items[0]` assumes the single-node kind cluster that `setup-test-e2e` builds today. Add a worker to that config and `items[0]` is likely `…-control-plane`, whose `node-role.kubernetes.io/control-plane:NoSchedule` taint the template doesn't tolerate (it tolerates only `nvidia.com/gpu`) — so *"should propagate … and schedule it"* fails for a reason that has nothing to do with the operator.

Both close with a guard that fails loudly instead of overwriting:

```go
nodeName, err := kubectlGet("nodes", "", "",
    "{.items[?(@.spec.taints==null)].metadata.name}")
gomega.Expect(err).NotTo(gomega.HaveOccurred())
nodeName = strings.Fields(nodeName)[0] // first schedulable node
// …after decoding the node:
gomega.Expect(node.Status.Capacity).NotTo(gomega.HaveKey(fakeGPUResource),
    "node %s already advertises %s: refusing to overwrite real GPU capacity",
    nodeName, fakeGPUResourceName)
```

test/e2e/static/gpu/gpu-template.yaml
```diff
@@ -0,0 +1,40 @@
...
+    resources:
+      cpu:
+        min: 100m
+        max: 2
+      memory:
+        min: 128Mi
+        max: 4Gi
+      nvidia.com/gpu:
+        min: "0"
+        max: "2"
```
JGuinegagne: `[Security]` Worth recording what this template does **not** bound, because this suite is what makes "GPU requests are bounded" tested behavior and a reader will infer more coverage than exists.

`validateResourceListBounds` (`internal/webhook/v1alpha1/resource_validator.go:76-79`) iterates the **bounds**, not the request, and `continue`s on any resource absent from bounds — and nothing in `validateWorkspaceAgainstTemplate` (`template_validator.go:88-145`) restricts which resource names a workspace may name. So against this exact template, `nvidia.com/gpu: 3` is refused (`gpu-exceed-workspace.yaml`, covered here) while `nvidia.com/mig-1g.5gb: 8`, `amd.com/gpu: 8` or `habana.ai/gaudi: 8` are admitted unbounded.

Not this PR's job to change the semantics, and per-name opt-in may well be intended. But the decision isn't written down anywhere, and accelerator resource names are exactly the family where an admin would assume the bound generalizes. Suggest a follow-up issue alongside #474 to settle whether `resourceBounds` being set should imply denial of *unlisted* extended resources — and a one-line note on this fixture either way.

JGuinegagne (commented):
**One blocking item, and it's about the suite's budget rather than the code.** Measured from CI:

| | specs | ginkgo duration | headroom vs `-ginkgo.timeout 60m` |
|---|---|---|---|
| `main` (last 2 runs) | 206 | 3200s / 3225s (53.4 / 53.8 min) | ~6.5 min (11%) |
| this branch (last 2 runs) | 215 | 3487s / 3524s (58.1 / 58.7 min) | **~1.3 min (2%)** |

+290s for 9 specs (~32s each) against the hard ceiling in `Makefile:187`. At 2% margin one slow `ubuntu-latest` runner times out the whole suite and CI goes red on every open PR until someone bumps the number — and #474 is queued to add more specs to this same `Describe`. Ask: raise `-timeout`/`-ginkgo.timeout` to 90m in this PR, since this is the change that consumes the margin, plus the two reductions in the inline comment on `workspace_gpu_test.go:37`

test/e2e/workspace_drift_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(originalImage).NotTo(BeEmpty())
+
+		By("mutating the primary container image out of band")
+		patch := fmt.Sprintf(
+			`{"spec":{"template":{"spec":{"containers":[{"name":"%s","image":"drifted-image:latest"}]}}}}`,
+			controller.PrimaryContainerName)
+		cmd := exec.Command("kubectl", "patch", "deployment", deploymentName,
+			"-n", workspaceNamespace, "--type=strategic", "-p", patch)
+		_, err = utils.Run(cmd)
```
JGuinegagne: `[Reliability]` Non-blocking. `Eventually(image == originalImage)` is also satisfied by the state *before* the patch, so this spec goes green in two cases where it proves nothing:

1. The patch is a silent no-op. Strategic merge on `containers` is keyed by `name`; a name that doesn't match **appends a container** rather than erroring, and the filter `containers[?(@.name=='<primary>')].image` still returns the original image. Correct today because both sides read `controller.PrimaryContainerName` — but there's no tripwire if that ever diverges from what the builder emits.
2. The controller reverts inside the first 2s poll, so the spec never observes the drift and can no longer distinguish "operator repaired it" from "mutation never landed."

The sibling Service spec is immune because it asserts the UID *changed*. Same guarantee here for one line:

```go
output, err := utils.Run(cmd)
Expect(err).NotTo(HaveOccurred())
Expect(output).NotTo(ContainSubstring("no change"),
    "the out-of-band patch must actually mutate the deployment, "+
        "else the revert assertion below is vacuous")
```

(`.metadata.generation` advancing by ≥2 works too, and doesn't depend on kubectl's output wording.)

test/e2e/workspace_gpu_test.go
JGuinegagne: Half of this is resolved: `3a2cc6c9` raises the ceiling to 90m, and more thoroughly than I asked — `test-e2e`, `test-e2e-focus` and `test-e2e-staging` all moved, not just line 187. Headroom goes from ~2% to ~35% (58.5 min of 90), and the GH job has no `timeout-minutes` so the 360m default does not bind. The blocking item in my summary is cleared.

Leaving this thread open for the other half, which is untouched: hoisting `createTemplateForTest` into the existing `BeforeAll` (identical fixture, currently created and finalizer-deleted six times) and scoping `deleteResourcesForGPUTest` to the one workspace each spec creates instead of waiting on all five. Now purely a runtime nicety rather than a timeout risk, so fine to defer or close as wontfix — with 90m there is room, and #474 will land in the same `Describe`.

JGuinegagne:
Status after `3a2cc6c9`/`8cde5006`/`e6fbebb9` — **the blocking item in my review above is cleared.** The e2e ceiling is now 90m across all three targets, so the 2% headroom concern is gone.

Two notes on the rest of that push, neither blocking:

- `VerifyWorkspaceConditions` now polling is the root-cause fix for #350, and it matches that issue's diagnosis exactly. Worth noting #350 was closed `COMPLETED` in March by *working around* it — `workspace_status_test.go:43-47` still carries a NOTE explaining that the `Progressing=True` assertion was dropped rather than fixed. That workaround and its NOTE are now unnecessary and could go.
- Since 19 call sites across 6 suites inherit the helper, retrying exact-set equality turns "wrong condition set right now" into a tolerated transient: a workspace that flaps `Degraded=True` and then settles will pass where it previously failed. Right trade for #350, and `VerifyConsistentWorkspaceConditions` still covers the strict case — just not the matcher to reach for when the point is that a bad state never appears.

Still open from my review, all non-blocking: the unguarded `nodes[0]` status patch (`helpers_gpu.go:48`), the `resourceBounds` per-name opt-in gap (`gpu-template.yaml:31`), the vacuous-pass window in the drift-revert spec (`workspace_drift_test.go:97`), and the GPU-suite churn reduction (`workspace_gpu_test.go:37`). Those posted three minutes after your last push, so they have not been looked at yet.

JGuinegagne (dismissed):

michaelnchin (dismissed):

test/e2e/helpers_gpu.go
andrii-i: Adopted, with two changes: the node is picked in Go because the `[?(@.spec.taints==null)]` filter matches nothing for an untainted node on kubectl 1.33, and the refusal is skipped when the node carries this suite's label from an earlier run.

test/e2e/workspace_drift_test.go
andrii-i: Adopted the generation check. The mutation itself is now `kubectl set image`, so an unknown container name errors instead of appending a container.

test/e2e/workspace_gpu_test.go
andrii-i: Adopted.

test/e2e/static/gpu/gpu-template.yaml
andrii-i: Note added. Follow-up: #480.

michaelnchin (dismissed):
LGTM, thanks @andrii-i !

JGuinegagne (dismissed):

test/e2e/workspace_status_test.go
```diff
@@ -41,10 +41,6 @@ var _ = Describe("Workspace Status", Ordered, func() {
 			By("creating workspace with desiredStatus=Running")
 			createWorkspaceForTest(runningWorkspace, statusGroupDir, statusSubgroupDir)
 
-			// NOTE: We intentionally do NOT assert Progressing=True immediately after creation.
-			// The controller sets .status.conditions asynchronously, so on slow/loaded CI runners
-			// the conditions may still be empty when checked, causing flaky failures.
-			// See https://github.com/jupyter-infra/jupyter-k8s/issues/350
 			By("waiting for Available condition to become True")
```
JGuinegagne: `[Non-blocking]` Dropping the #350 NOTE here leaves the workaround it explained in place — `Progressing=True` is asserted nowhere in the suite — with nothing recording why, and #350 is closed `COMPLETED`.

The polling fix does not make it assertable either: `Progressing=True` is a transient, and `VerifyWorkspaceConditions` now waits for a settled exact set, so worth restoring one line saying the assertion is deliberately absent for that reason rather than for runner slowness.

andrii-i: Adopted.

JGuinegagne (approved):

michaelnchin (approved):

PR #478, opened 2026-09-08
docs: document WorkspaceIntegrationTemplate and fix stale field ref (#461)

### LINKED ISSUE #461: Docs: update WorkspaceIntegrationTemplate documentation
The docs for `WorkspaceIntegrationTemplate` need updating:

- The generated reference (`docs/source/reference/custom-resources/workspaceintegrationtemplate.md`) says a Workspace attaches these via `spec.integrationRefs`, but the actual field is `spec.integrationTemplateRefs`. Fix the godoc in `api/v1alpha1/workspaceintegrationtemplate_types.go` and regenerate.
- There is no conceptual/authoring guide — only the auto-generated CRD reference. Add a page covering when to use a WIT, the template expressions, the parameter contract, `resourceRefs`, `statusProbe`, and a worked example.

HariniNarayanan:
## Summary

Resolves #461.

Closes two documentation gaps for `WorkspaceIntegrationTemplate` (WIT):

1. **Stale field reference in the generated CRD reference.** The reference stated that a Workspace attaches integrations via `spec.integrationRefs`; the actual field is `spec.integrationTemplateRefs`. Fixed the godoc on `WorkspaceIntegrationTemplate` and regenerated, so the correction propagates to both the CRD manifest and the reference doc. `spec.integrationRefs` no longer appears anywhere in the repo.

2. **No conceptual/authoring guide.** Only the auto-generated CRD reference existed. Added a new `concepts/integration-templates/` section, structured like `concepts/access-strategies/`: an index page for usage, plus one page per authoring concern.

## Changes

**Field fix (source + regenerated artifacts):**
- `api/v1alpha1/workspaceintegrationtemplate_types.go` — godoc `spec.integrationRefs` → `spec.integrationTemplateRefs`
- `config/crd/bases/workspace.jupyter.org_workspaceintegrationtemplates.yaml` — regenerated via `make manifests`
- `docs/source/reference/custom-resources/workspaceintegrationtemplate.md` — regenerated via `make docs-crd-ref`

**New authoring guide — `docs/source/concepts/integration-templates/`:**
- `index.md` — what a WIT is, when to use one, and how a workspace attaches it (`spec.integrationTemplateRefs`)
- `parameters.md` — the parameter contract (all declared parameters required; undeclared/empty rejected at admission)
- `template-expressions.md` — `{{ resource }}` / `{{ .Parameters }}` / `{{ .Workspace }}`, `resourceRefs`, and resolution/drift (values are frozen on unchanged input, so drift never rolls a running pod)
- `deployment-modifications.md` — injected sidecars/volumes/env plus `shareProcessNamespace`; cross-links the access-strategy deployment-modifications reference and calls out that a WIT freezes the resolved result instead of re-resolving live
- `status-probe.md` — the report-only `statusProbe` (never gates readiness or restarts the pod)
- `example.md` — a simplified, faithful version of the `aws-hyperpod` chart's `ray-integration` template

**Wiring:**
- `docs/source/concepts/index.md` — lists WIT as the fourth managed CRD and adds it to the concepts toctree
- `docs/source/integrations/guided-charts/aws-charts.md` — new "AWS-HyperPod Ray Integration Template" subsection

## Testing

- `make manifests` and `make docs-crd-ref` regenerate cleanly.
- Local Sphinx build (`sphinx-build -b html`) succeeds with no warnings; all new pages render and internal cross-links resolve.
- `grep -r 'spec.integrationRefs'` returns nothing across the repo.

The worked example is a simplified form of `charts/aws-hyperpod/templates/ray-integration-template.yaml` in `jupyter-k8s-aws`; the page links to the full chart template for the omitted production details (retry loop, resource sizing, security context, GPU handling).


henrywa2 (approved):

docs/source/concepts/integration-templates/deployment-modifications.md
```diff
@@ -0,0 +1,12 @@
...
+- **The result is frozen, not re-applied live.** This is the key distinction: an access strategy re-resolves its `mergeEnv` on every reconcile, so a change to its inputs flows straight through. A WIT does not work this way. It resolves once, records the rendered sidecars, volumes, and environment variables in `status.resolvedIntegrations`, and replays that frozen result on later reconciles. Editing the referenced resource does not re-render the workspace pod; the injection changes only when the parameters or the template itself change. See [Resolution and drift](template-expressions.md#resolution-and-drift).
+
+## shareProcessNamespace
+
+`spec.shareProcessNamespace` is an admin-only, pod-level toggle that shares the PID namespace across all containers in the workspace pod, so the workspace container can see and signal processes running in an injected sidecar (for example, to attach to a sidecar's local Ray session). Containers sharing a PID namespace can read each other's `/proc` (filesystem and environment), so injected containers should run as non-root.
```
henrywa2, line 12: Why is this field specifically marked with `admin-only`? Are non-admins expected to change the other fields?

api/v1alpha1/workspaceintegrationtemplate_types.go
```diff
@@ -143,7 +143,7 @@ type IntegrationStatusProbe struct {
 // It defines a declarative, template-driven integration for adding runtime capabilities
 // (sidecars, volumes, env vars) to workspace pods with dynamic resource lookup and
 // template expression resolution. A Workspace may attach several of these via
```
henrywa2, line 145: aren't these capped at 1, not several?

psnilesh (approved):

docs/source/concepts/integration-templates/deployment-modifications.md
```diff
@@ -0,0 +1,12 @@
+# Deployment modifications
+
+`spec.deploymentModifications.podModifications` is where a WIT actually changes the workspace pod. It uses the same shape as an access strategy's deployment modifications: `additionalContainers`, `initContainers`, `volumes`, `primaryContainerModifications.volumeMounts`, and `primaryContainerModifications.mergeEnv`. See [Access Strategies: Deployment Modifications](../access-strategies/deployment-modifications) for the field-by-field reference; the fields behave the same here.
```
JGuinegagne, line 3: avoid jargon please, spell out `WorkspaceIntegrationTemplate`

docs/source/concepts/integration-templates/example.md
```diff
@@ -0,0 +1,81 @@
+# Example setup
+
+This is a simplified version of the production `ray-integration` template shipped in the [`aws-hyperpod` chart](https://github.com/jupyter-infra/jupyter-k8s-aws/blob/main/charts/aws-hyperpod/templates/ray-integration-template.yaml). An admin installs it to attach a workspace to an existing `RayCluster`: it fetches the cluster named by the `rayClusterName` parameter, injects a Ray sidecar that joins the cluster as a zero-resource client node, mounts the shared session directories, sets the Ray environment variables in the workspace container, and reports Ray reachability through a `statusProbe`.
+
+```yaml
```
JGuinegagne, line 5: this is highly likely to diverge and become stale, consider a simple redirect.

docs/source/concepts/integration-templates/index.md
```diff
@@ -0,0 +1,57 @@
+# Integration Templates
+
+A **WorkspaceIntegrationTemplate** (WIT) injects runtime capabilities into a workspace pod: sidecar containers, volumes, and environment variables. Unlike a **WorkspaceTemplate**, which sets static defaults and bounds, a WIT resolves its values dynamically at reconcile time by reading a live Kubernetes resource and substituting template expressions.
```
JGuinegagne, line 3: nit: remove "WIT"

JGuinegagne (commented):
Few high-level comments

docs/source/concepts/integration-templates/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+A WIT has a 1:many relationship with workspaces. Multiple workspaces may attach the same template, each supplying its own parameters.
+
+## When to use a WIT
+
+Reach for a WIT when a workspace needs to connect to another resource that exists in the Kubernetes cluster, and the connection details are only known at runtime. The canonical case is wiring a workspace to a `RayCluster` or similar object: the WIT fetches that object, reads fields from it (host, port, name), and renders them into the workspace pod as sidecars or environment variables.
+
+Use a plain **WorkspaceTemplate** instead when the configuration is static (a fixed image, resource bounds, a default access strategy). Use a WIT when the configuration must be computed from another object's live state.
+
+Like access strategies, a WIT is an administrator-owned resource. Admins install a small set of vetted integrations; workspace users attach them by reference and supply parameters. In an enterprise cluster, workspace users should not have permission to create or edit integration templates directly.
```
JGuinegagne, line 13: Worth making the RBAC recommendation concrete here, since this sentence is the only place the trust boundary is stated and no page in `docs/source/` currently spells out verbs for any CRD (`concepts/templates/index.md:34` just says "the right RBAC permission").

Denying write on `WorkspaceIntegrationTemplate` is the control that matters: a user who can author one picks the `kind`, the JSONPath, *and* the object name, which is a full confused-deputy read of anything the operator's ServiceAccount can reach in their namespace. Granting `get`/`list` is fine and necessary (users must be able to discover which parameters to supply). Suggested:

> Workspace users need `get`/`list` on `workspaceintegrationtemplates` to discover available integrations; they must not have `create`, `update`, `patch`, or `delete`. The template author chooses the `kind` and JSONPath a `WorkspaceIntegrationTemplate` reads, so write access on this resource is equivalent to read access on any object of any kind the operator can reach in the namespace.

One thing RBAC on this resource cannot bound, which is the actual vetting criterion: `resourceRefs[].metadata.name` is templated from a user-supplied parameter, and `validateWorkspaceIntegrationParameters` (`integration_template_ref_validator.go:158-171`) only checks present-and-non-empty, never the value. So within the kind you vetted, the *user* chooses which instance. For the Ray example the sidecar then runs `ray start --address=...`, i.e. it joins that cluster rather than just reading it.

That's a non-issue if one namespace is always one team. If that's an invariant of jupyter-k8s deployments, stating *it* here is probably more useful than a caveat. If it isn't, one sentence under `resourceRefs` in `template-expressions.md` would cover it:

> The workspace user chooses which object to fetch, so the `kind` and JSONPath you write define the entire blast radius of the integration.

This also answers @henrywa2's question on `deployment-modifications.md:12` — the whole CRD is admin-owned, which is why singling out `shareProcessNamespace` as "admin-only" reads as if the other fields aren't.

docs/source/concepts/integration-templates/deployment-modifications.md
```diff
@@ -0,0 +1,12 @@
...
+
+Two differences matter:
+
+- **Template context.** A WIT resolves `{{ resource "<handle>" "<jsonpath>" }}`, `{{ .Parameters.<name> }}`, and `{{ .Workspace.* }}`. An access strategy instead exposes `.Workspace` and `.AccessStrategy`, and has no `resource` function. See [Template expressions](template-expressions).
+- **The result is frozen, not re-applied live.** This is the key distinction: an access strategy re-resolves its `mergeEnv` on every reconcile, so a change to its inputs flows straight through. A WIT does not work this way. It resolves once, records the rendered sidecars, volumes, and environment variables in `status.resolvedIntegrations`, and replays that frozen result on later reconciles. Editing the referenced resource does not re-render the workspace pod; the injection changes only when the parameters or the template itself change. See [Resolution and drift](template-expressions.md#resolution-and-drift).
```
JGuinegagne, line 8: This sentence doesn't match the implementation, and the distinction is the kind of thing someone reaches for mid-incident.

`status.resolvedIntegrations[].values` stores **only** the `{{ resource }}` substitution map — keys are `"<resourceRefID>|<jsonPath>"`. The rendered output is explicitly discarded (`captureIntegrationValues`, `resource_manager_integration.go:164-177`), and the field's own godoc says "Storing only these substitutions (not the rendered pod spec) keeps the status payload small." So the rendered sidecars, volumes, and env vars are not recorded.

Nor is "that frozen result" replayed wholesale: `buildIntegrationPodModifications` (`deployment_builder_integration.go:80-105`) reads the pod **shape** from the live template on every reconcile and resolves only the `{{ resource }}` values from the frozen map. Per the file header at `resource_manager_integration.go:29-32`: "the build renders the pod SHAPE from the live template every reconcile, so a shape edit rolls the pod regardless."

The practical consequence the page currently obscures: editing the template's shape (adding a volumeMount, changing a command that contains no `{{ resource }}`) takes effect on the next reconcile and rolls the pod. Only values read out of the referenced resource are frozen. Suggested rewrite:

> **Resource values are frozen; the pod shape is not.** An access strategy re-resolves its `mergeEnv` on every reconcile, so a change to its inputs flows straight through. A `WorkspaceIntegrationTemplate` splits the two. The pod shape (which sidecars, volumes, and env names) is read from the live template every reconcile, so an admin edit to the template takes effect immediately. But the values read out of the referenced resource via `{{ resource }}` are captured once into `status.resolvedIntegrations[].values` and replayed from there, so editing the referenced resource does not re-render the pod. Re-capture happens only when the parameters or the template version change.

docs/source/concepts/integration-templates/status-probe.md
```diff
@@ -0,0 +1,12 @@
+# Status probe
+
+`spec.statusProbe` is an optional probe the operator runs on its poll cadence to check that the integration is reachable from the workspace pod's real runtime context. It is report-only: the verdict is written to `workspace.status.integrationStatuses[]`. Unlike a `readinessProbe`, it never changes the pod's readiness, so it never adds or removes the pod from its `Service` endpoints (a failing probe cannot cut the workspace off from traffic); and unlike a `livenessProbe`, it never restarts the pod. It is named `statusProbe` for that reason.
```
JGuinegagne, line 3: "on its poll cadence" is the only mention of timing, and it doesn't name the knob. It's operator-level, not per-template: `--integration-probe-period`, default 5 minutes (`DefaultIntegrationProbePeriod`), clamped up to a 1s floor (`MinIntegrationProbePeriod`). Worth stating, since `timeoutSeconds` *is* documented here and readers will otherwise assume the period is per-template too.

Also worth a line that the cadence is flat with no per-verdict backoff (`integration_prober.go:30-33`) — a persistently failing probe execs into the pod every period indefinitely. Events are edge-triggered so there's no event spam, but the exec load is constant.

docs/source/concepts/integration-templates/parameters.md
```diff
@@ -0,0 +1,18 @@
...
+
+```yaml
+spec:
+  parameters:
+    - name: rayClusterName
+```
+
+All declared parameters are required. There are no optional parameters or defaults:
+
+- A template expression that references an undeclared parameter is rejected when the admin writes the template.
```
JGuinegagne, line 13: This is true for the field-access form only. `validateTemplateExpression` documents two escapes at `integration_template_validation.go:192-204`: `{{ index .Parameters "X" }}` bypasses `missingkey=error` entirely (text/template's `index` returns the map's zero value for an unknown key rather than erroring), and expressions inside an untaken `{{ if }}`/`{{ range }}` branch are never exercised, because validation seeds every declared parameter to `""`. Both surface at resolve time as a degraded workspace instead.

Suggest softening to "rejected when the admin writes the template, as long as it uses the `{{ .Parameters.X }}` form outside a conditional", plus a pointer that `{{ index .Parameters "X" }}` is not statically checked.

docs/source/concepts/integration-templates/template-expressions.md
```diff
@@ -0,0 +1,30 @@
...
+      name: "{{ .Parameters.rayClusterName }}"   # object name; supports template expressions
+```
+
+`apiVersion` and `kind` identify the resource type; `metadata.name` identifies the specific object and may itself be templated (for example `{{ .Workspace.Name }}-ray`). The object is always looked up in the referencing workspace's own namespace; a `resourceRef` cannot target another namespace.
+
+A WIT requires exactly one `resourceRef` today (minimum one, capped at one). A template exists to resolve values from a referenced resource, so at least one ref is required.
+
+## Resolution and drift
+
+The controller resolves an integration only when its input changes: a hash of the template ref plus the supplied parameters, or the template's own generation (after an admin edits the WIT). On any other reconcile, including external drift in the referenced resource or an idle reconcile, the controller replays the frozen values recorded in `status.resolvedIntegrations` rather than re-reading the resource. Drift in a referenced object therefore never rolls a running pod. Re-resolution happens only on an intentional change to the parameters or the template.
```
JGuinegagne, line 30: "the template's own generation" — it's `"<UID>.<Generation>"` (`getIntegrationTemplateVersion`, `resource_manager_integration.go:246-248`), so deleting and recreating a template with an identical spec also forces re-resolution. Maybe "the template's version (its UID and generation, so a delete-and-recreate also re-resolves)".

docs/source/concepts/integration-templates/deployment-modifications.md
HariniNarayanan: Dropped the label. No field is privileged over another -- write access is what decides, now stated once under Ownership and RBAC.

api/v1alpha1/workspaceintegrationtemplate_types.go
HariniNarayanan: Correct, `MaxItems=1`. Fixed the godoc and regenerated the CRD, `install.yaml`, the chart CRD, and the reference doc.

docs/source/concepts/integration-templates/deployment-modifications.md
HariniNarayanan: Done, and removed "WIT" throughout -- no occurrences left in `docs/source/`.

docs/source/concepts/integration-templates/index.md
HariniNarayanan: Done.

docs/source/concepts/integration-templates/example.md
HariniNarayanan: Done -- the page is a redirect to the chart now, inlined YAML removed.

docs/source/concepts/integration-templates/status-probe.md
HariniNarayanan: Done. New Cadence section: `--integration-probe-period`, operator-level, default 5m, clamped to a 1s floor, and the flat-cadence/no-backoff point.

docs/source/concepts/integration-templates/template-expressions.md
HariniNarayanan: Done -- now `"<UID>.<Generation>"`, with the delete-and-recreate case spelled out.

docs/source/concepts/integration-templates/deployment-modifications.md
HariniNarayanan: Corrected, using your framing. I did restructure the page while I was in it -- the "two differences matter" comparison against access strategies is gone and this is its own section now, so the wording isn't verbatim yours. Also added the `"<resourceRefID>|<jsonPath>"` key format and that only substitutions are stored.

docs/source/concepts/integration-templates/parameters.md
HariniNarayanan: Done.

docs/source/concepts/integration-templates/index.md
HariniNarayanan: Used your wording, under a new Ownership and RBAC section, plus the `resourceRefs` sentence in `template-expressions.md`. Also documented the operator's read grant on the referenced kind as an install step.

docs/source/concepts/integration-templates/index.md
```diff
@@ -0,0 +1,100 @@
...
+  displayName: Alice's Workspace
+  image: my-repository/my-image:my-tag
+  integrationTemplateRefs:
+    - name: ray-integration
+      parameters:
+        - name: rayClusterName
+          value: team-ray
+```
+
+`spec.integrationTemplateRefs` is capped at one entry for now. The reference resolves within the workspace's own namespace. As with templates and access strategies, a workspace may also reference an integration template in the [shared namespace](../templates/shared-namespace), a special namespace identified at the **Jupyter K8s** operator level.
```
JGuinegagne: `[Reliability]` This states the shared-namespace path without naming the field that selects it, and the Usage example above is the form that won't work against the guided chart.

`IntegrationTemplateRef` has an optional `Namespace` (`api/v1alpha1/workspace_types.go:277-279`), and `validateNamespaceScope` (`integration_template_ref_validator.go:133-149`) requires it set **explicitly** — an omitted namespace always means the workspace's own namespace, never a fallback to shared. The accepted set is exactly: unset, the workspace's own namespace, or the configured shared namespace. Anything else is rejected *before* the template read, so the guard is worth stating positively here.

This matters because the one integration template that ships today installs into the chart's namespace (`ray-integration-template.yaml`: `namespace: {{ .Values.namespace }}`), not a per-user workspace namespace. A reader copying the block above gets `integrationTemplateRefs["ray-integration"]: WorkspaceIntegrationTemplate "ray-integration" not found in namespace "alice-team"`.

Two lines fix it — in the example:

```yaml
  integrationTemplateRefs:
    - name: ray-integration
      namespace: jupyter-k8s   # the shared namespace; omit when the template is in the workspace's own namespace
      parameters:
        - name: rayClusterName
          value: team-ray
```

and in the prose: an omitted `namespace` means the workspace's own namespace, and the only other accepted value is the configured shared namespace.

docs/source/concepts/integration-templates/deployment-modifications.md
```diff
@@ -0,0 +1,31 @@
+# Deployment modifications
+
+`spec.deploymentModifications.podModifications` declares what the integration adds to the workspace pod: `additionalContainers`, `initContainers`, `volumes`, `primaryContainerModifications.volumeMounts`, and `primaryContainerModifications.mergeEnv`. These fields take the same shape as the block of the same name on an access strategy; see [Access Strategies: Deployment Modifications](../access-strategies/deployment-modifications) for the field-by-field reference.
```
JGuinegagne: `[Hardening]` `PodModifications` also carries `exposedPorts`, which the cross-linked access-strategy reference documents, but it is a no-op here — `sidecarServicePorts` (`internal/controller/service_builder.go:106`) takes a `*WorkspaceAccessStrategy`, `deployment_builder_integration.go` never reads it, and the integration-template schema still accepts it. Worth a clause noting that sidecar ports are published through an access strategy, so an admin following the link doesn't set it and get a clean admission with an unreachable port.

docs/source/integrations/guided-charts/aws-charts.md
```diff
@@ -38,6 +38,7 @@ Users connect either a/ from their web browser using bearer token URL or b/ from
 **What it deploys:**
 - **Jupyter K8s** controller with the [AWS Plugin](../plugins/aws-plugin) sidecar
 - Preconfigured access strategies for SSM-based remote access (VS Code, Cursor)
+- A Ray integration template that attaches workspaces to an existing `RayCluster` (see [AWS-HyperPod Ray Integration Template](#aws-hyperpod-ray-integration-template) below)
```
JGuinegagne: `[Reliability]` `rayIntegration.enabled` defaults to `false` (`charts/aws-hyperpod/values.yaml:165`) and the template is wrapped in `{{- if .Values.rayIntegration.enabled }}`, so this belongs in the `Optionally, …` group below rather than the unconditional **What it deploys** list.

docs/source/integrations/guided-charts/aws-charts.md
```diff
@@ -53,6 +54,14 @@ Optionally, components for Web UI with:
 | Package | `oci://ghcr.io/jupyter-infra/charts/jupyter-k8s-aws-hyperpod` |
 | Source | [charts/aws-hyperpod](https://github.com/jupyter-infra/jupyter-k8s-aws/tree/main/charts/aws-hyperpod) |
 
+### AWS-HyperPod Ray Integration Template
+
+The chart ships a [`WorkspaceIntegrationTemplate`](../../concepts/integration-templates/index) named `ray-integration` that attaches a workspace to an existing `RayCluster`. When a workspace references it and supplies the `rayClusterName` parameter, the controller injects a Ray sidecar (matched to the cluster's head image) that joins as a zero-resource client node, mounts the shared session directories, and sets `RAY_ADDRESS=auto` plus `RAY_CLUSTER_NAME` in the workspace container, so `ray.init()` connects with no manual address.
+
+- **Enable:** `--set rayIntegration.enabled=true`
```
JGuinegagne: `[Reliability]` The flag alone isn't sufficient. The operator also needs `get` on `ray.io/rayclusters`, which the new [Permissions for the operator](../../concepts/integration-templates/index) section calls a required install step — and I can't find that grant anywhere in `jupyter-k8s-aws`: no ClusterRole in the chart mentions `ray.io`, and `RAY_INTEGRATION.md` doesn't mention RBAC. So `--set rayIntegration.enabled=true` on its own installs a template whose first resolve is `forbidden` → fail-closed → degraded integration.

Here: add "requires the operator RBAC grant (see [Permissions for the operator](...))" next to the enable flag. Separately, worth a `jupyter-k8s-aws` follow-up so `RAY_INTEGRATION.md` states the prerequisite, and ideally the chart ships the grant behind the same flag.

